PR #1999, opened 2021-11-24
Issue/1866 update relative times

### LINKED ISSUE #1866: Created and updated dates in service inventory do not refresh
The created  and updated dates in the service inventory do not refresh. When the page stays open, it will keep saying the same amount of time. Only then reload the page, it will update to the correct x minutes ago.

->relative time is not correctly refreshed. 

-> create auto-updating component
-> update everywhere relative time is used

Stijnkool:
# Description

closes #1866 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/UI/Utils/useTicker.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+  render(<Component />);
+
+  act(() => {
+    jest.advanceTimersByTime(1000);
+  });
+
+  expect(callback).toBeCalledTimes(2);
+});
+
+test("GIVEN useTickerWithTimestamp WHEN provided with timestamp longer than 2 hours ago THEN never executes again", () => {
```
andraskvr, line 23: I'm not sure about this use case.
If we set a limit like this, and someone leaves a tab open overnight, it will still show `2 hours ago` (and suddenly updates when the user navigates) which is not great, and this can also happen if we increase the limit.
If we don't have a limit, then it could potentially run forever, which feels a bit like a waste of resources.

We can go ahead with this, and still change it later based on feedback.

andraskvr (approved):

PR #2005, opened 2021-11-25
Add deploy and repair buttons

### LINKED ISSUE #1807: Add deploy and repair buttons next to the progress bar
Next to the progressbar two buttons need to be added:

- Deploy
- Repair

These buttons should be both secondary styles button. This might fit the https://www.patternfly.org/v4/components/toolbar component.

The entire deploy / repair model needs some changes on the backend, so for now we will use the v1 api instead of a v2 api. The api call is quite simple.

- Deploy: post call to /v1/deploy with agent_trigger_method set to push_incremental_deploy
- Repair: post call to /v1/deploy with agent_trigger_method set to push_full_deploy

- Not allowed when the environment is halted. 
- Silently ignore all error responses.

andraskvr:
# Description

closes #1807 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/UI/Pages/Resources/Summary.tsx
+++ b/src/UI/Pages/Resources/Summary.tsx
@@ -1,6 +1,9 @@
 import React from "react";
+import { Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core";
 import { Query, RemoteData } from "@/Core";
+import { DeployButton } from "./DeployButton";
 import { DeployStateChart } from "./DeployStateChart";
+import { RepairButton } from "./RepairButton";
 
 interface Props {
   data: RemoteData.Type<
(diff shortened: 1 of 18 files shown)


src/UI/Pages/Resources/Summary.tsx
```diff
@@ -16,7 +19,19 @@ export const Summary: React.FC<Props> = ({ data }) =>
       loading: () => null,
       failed: () => null,
       success: (result) => (
-        <DeployStateChart summary={result.metadata.deploy_summary} />
+        <Toolbar>
```
Stijnkool, line 22: In the UI there is a lot of padding. This is because we have nested toolbars.
If you replace the `Toolbar` and `ToolbarContent` here with fragments, it is fixed.

Stijnkool (approved):
Nice 👍

src/UI/Pages/Resources/Summary.tsx
andraskvr: Ok, that's a good remark

PR #2015, opened 2021-11-26
Add facts tab

### LINKED ISSUE #978: Add a facts tab to the resource details
For each resource, add a tab to the resource details page: 

Table with fields

- name
- last updated 
- value

Make table sortable in the front-end

andraskvr:
# Description

closes #978 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/UI/Pages/ResourceDetails/Tabs/FactsTab/FactsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const columns = factsColumnHeads.map((head) => head.apiName);
+  return columns[index];
+}
+
+function columnNameToIndex(columnName: string): number {
+  const columns = factsColumnHeads.map((head) => head.apiName);
+  return columns.indexOf(columnName);
+}
+
+export function sortFactRows(
```
Stijnkool, line 91: I think I would make the compare function a named function and just put the `.sort` inline.
But it doesn't really matter. Just preference.

src/UI/Pages/ResourceDetails/Tabs/FactsTab/FactsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+
+const factsColumnHeads: ColumnHead[] = [
+  { apiName: "name", displayName: words("resources.facts.columns.name") },
+  { apiName: "updated", displayName: words("resources.facts.columns.updated") },
+  { apiName: "value", displayName: words("resources.facts.columns.value") },
+];
+
+export const FactsTable: React.FC<Props> = ({ facts }) => {
+  const [activeSortColumn, setActiveSortColumn] = useState("name");
```
Stijnkool, line 28: We no longer have this sorting as 2 separate properties. We have a `Sort` type that covers this.
I think it is better to use that. Not sure if this sorting needs to be saved in the url.

Stijnkool (changes requested):
Looks good. 👍 
Just 1 remark about the sorting

src/UI/Pages/ResourceDetails/Tabs/FactsTab/FactsTable.tsx
andraskvr: To me, having it like this is a bit more readable

src/UI/Pages/ResourceDetails/Tabs/FactsTab/FactsTable.tsx
andraskvr: The `Sort` type is a good idea.
Saving it in the url is not necessary I think since the endpoint is not paged

Stijnkool (approved):

PR #2109, opened 2021-12-06
Add description and icon to environment cards

### LINKED ISSUE #2016: Add icon and description to the environment cards
https://www.patternfly.org/v4/demos/card-view/html-demos/card-view/

andraskvr:
# Description

closes #2016 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/UI/Pages/Home/CardView.tsx
+++ b/src/UI/Pages/Home/CardView.tsx
@@ -76,16 +76,16 @@ const EnvironmentCard: React.FC<EnvironmentCardProps> = ({
   pathname,
 }) => (
   <Card isHoverable isCompact aria-label={"Environment card"}>
-    <CardHeader>
+    <StyledHeader>
       {environment.icon && (
-        <img
-          style={{ height: "2em" }}
+        <StyledIcon
           src={`data:${environment.icon}`}
           alt={words("home.environment.icon")(environment.name)}
-        ></img>
+          aria-label={`${environment.name}-icon`}
+        />
       )}
       <Actions environment={environment} />
-    </CardHeader>
+    </StyledHeader>
     <CardTitle>{environment.name}</CardTitle>
     <CardBody>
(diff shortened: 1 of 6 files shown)


src/UI/Pages/Home/CardView.tsx
```diff
@@ -78,28 +77,32 @@ const EnvironmentCard: React.FC<EnvironmentCardProps> = ({
...
   <Card isHoverable isCompact aria-label={"Environment card"}>
     <CardHeader>
-      <CardTitle>{environment.name}</CardTitle>
+      {environment.icon && (
+        <img
+          style={{ height: "2em" }}
+          src={`data:${environment.icon}`}
+          alt={words("home.environment.icon")(environment.name)}
+        ></img>
+      )}
```
Stijnkool, line 86: `img` can be a self closing element I think.
And we already have some styled components in this file.
Might as well make the image a styled component as well.

src/UI/Pages/Home/EnvironmentsOverview.test.tsx
```diff
@@ -106,3 +106,10 @@ test("Given environments overview When filtering by name and project Then only t
     })
   ).toBeVisible();
 });
+
+test("Given environments overview When rendering environment with icon Then the icon is shown", async () => {
+  const { component } = setup();
+  render(component);
+  const cardsWithIcons = await screen.findByRole("img");
```
Stijnkool, line 113: This seems like a very broad query. Not sure if adding a test for the alt value is useful here.
Maybe test both the presence and absence of the image for 2 separate environments?

Stijnkool (changes requested):
Screenshot 2021-12-06 at 11 50 27
When there is no image for the environment, the env name is higher up.
Or the card header is smaller. I think it would be better if the header always has the same height. The env names should line up.

Stijnkool (approved):

PR #2153, opened 2021-12-10
Fix broken styling

### LINKED ISSUE #1892: Visual regressions due to patternfly updates
We have noticed some visual regressions.
We think it's due to some patternfly packages being updated by dependabot.

## Input fields have borders again and their text jumps up on focus
no interaction:

focus:

## Input value sticks to the sides (no padding)
text value:

number controls:

andraskvr:
# Description

The problem was not related to dependency updates, but to the reordering of the imports. The custom styling of the components were overwritten by the `base` styling from Patternfly. Importing the `base` from further up the tree seems to have fixed the issue.

closes #1892 

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/1892-fix-broken-styling.yml
+++ b/changelogs/unreleased/1892-fix-broken-styling.yml
@@ -1,0 +1,4 @@
+description: Fix broken styling
+issue-nr: 1892
+change-type: patch
+destination-branches: [master, iso4]
--- a/src/UI/Root/app.tsx
+++ b/src/UI/Root/app.tsx
@@ -1,5 +1,4 @@
 import React, { useContext } from "react";
-import "@patternfly/react-core/dist/styles/base.css";
 import { KeycloakProvider } from "react-keycloak";
 import { Route, Routes } from "react-router-dom";
 import { Spinner, Bullseye } from "@patternfly/react-core";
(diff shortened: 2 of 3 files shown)


Stijnkool:
The fact that the ordering of the imports has an influence is not really good news.
Can we fix this?
What custom styling was being overwritten? Do we need this custom styling?
Ideally we want to fix this in a way where we control the order of the css so we can make sure to have the base always first. and everything after that.

This is why I don't like traditional css and its global scope. Too much risk of stuff breaking.
But there is no way around it for now.

Stijnkool (approved):

andraskvr:
> The fact that the ordering of the imports has an influence is not really good news. Can we fix this? What custom styling was being overwritten? Do we need this custom styling? 

By custom I meant more like the component-specific css. So for example the patternfly `TextInput` style was overridden by the patternfly `base` styles.

> Ideally we want to fix this in a way where we control the order of the css so we can make sure to have the base always first. and everything after that.

This is now basically the first import of the application so I think it's it will be the first in the merged `app.css` as well after the build.

> This is why I don't like traditional css and its global scope. Too much risk of stuff breaking. But there is no way around it for now.

Indeed. There's component specific styling with Patternfly that's bundled with the components themselves, but they all depend on the base styling, so it's kind of a strange mix

PR #2173, opened 2021-12-14
Add desired state table

### LINKED ISSUE #2007: Create an overview table for the desired state
The desired state page should show a table with an overview of the services. See below a mockup of such a table:

image

- This page does not have sorting, but it will have filtering and paging.
- By default the status only includes active, candidate and retired. So only when you explicitly include skipped candidate, it is shown.
- Filters are:
  - version range 
  - date range
  - status (multi select)
 - The source field is actually a list of labels (so labels instead of source) (no header)
 - Each label has a tooltip message
 - The last column has actions:
    - a primary button to show **details**
    - a button to "promote" a version. This button only works/is shown when state is candidate

andraskvr:
# Description

closes #2007 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/UI/Pages/DesiredState/Actions.tsx
+++ b/src/UI/Pages/DesiredState/Actions.tsx
@@ -3,18 +3,16 @@ import { Button, Flex, FlexItem } from "@patternfly/react-core";
 import { words } from "@/UI/words";
 
 export const Actions: React.FC = () => (
-  <>
-    <Flex direction={{ default: "column", "2xl": "row" }}>
-      <FlexItem>
-        <Button isDisabled variant="primary">
-          {words("desiredState.actions.details")}
-        </Button>
-      </FlexItem>
-      <FlexItem>
-        <Button isDisabled variant="secondary">
-          {words("desiredState.actions.promote")}
-        </Button>
-      </FlexItem>
-    </Flex>
-  </>
+  <Flex direction={{ default: "column", "2xl": "row" }}>
+    <FlexItem>
+      <Button isDisabled variant="primary">
+        {words("desiredState.actions.details")}
+      </Button>
+    </FlexItem>
+    <FlexItem>
+      <Button isDisabled variant="secondary">
+        {words("desiredState.actions.promote")}
+      </Button>
+    </FlexItem>
+  </Flex>
 );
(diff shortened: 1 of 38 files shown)


src/UI/Pages/DesiredState/Actions.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { Button, Flex, FlexItem } from "@patternfly/react-core";
+import { words } from "@/UI/words";
+
+export const Actions: React.FC = () => (
+  <>
```
Stijnkool, line 6: React Fragment is not needed here.

src/UI/Pages/DesiredState/DesiredStates.test.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+  userEvent.click(
+    within(screen.getByRole("generic", { name: "FilterBar" })).getByRole(
+      "button",
+      { name: "FilterPicker" }
+    )
+  );
+  userEvent.click(screen.getByRole("option", { name: "Date" }));
+
+  const fromDatePicker = await screen.findByLabelText("From Date Picker");
```
Stijnkool, line 175: Why do we need .find here? Is the date picker not immediately present?
Or does the test have to wait once you have selected the Data filter?

src/UI/Pages/DesiredState/DesiredStates.test.tsx
```diff
@@ -0,0 +1,266 @@
...
+  userEvent.click(fromDatePicker);
+  userEvent.type(fromDatePicker, `2021-12-06`);
+  const toDatePicker = await screen.findByLabelText("To Date Picker");
+  userEvent.click(toDatePicker);
+  userEvent.type(toDatePicker, `2021-12-07`);
+
+  userEvent.click(await screen.findByLabelText("Apply date filter"));
+
+  expect(apiHelper.pendingRequests[0].url).toMatch(
+    `/api/v2/desiredstate?limit=20&sort=version.desc&filter.status=active&filter.status=candidate&filter.status=retired&filter.date=ge%3A2021-12-`
```
Stijnkool, line 185: This url is missing part of the date filter. Is this on purpose?

Stijnkool (approved):
Small bits. 👍

PR #2351, opened 2022-01-12
Issue/1960 initial loading flow

### LINKED ISSUE #1960: Analyse initial loading flow using redux-devtools
Using the redux devtools, we can check every state transition and see how the application is presented.
During actual usage, some of the api calls are so quick, you barely see a transition.
Sometimes you see a flash and ignore it.
With the devtools we can properly analyse these problems.

The main goal of this ticket is to find bugs or inconsistencies that we can improve.
Small things can be fixed immediately. For more substantial changes we should create extra tickets.

This is an example of a view that can be improved:

- The error message for the sidebar is wrong. The environment is nog missing, it is loading.
The error message is visually unreadable because of the black background.
- The environment selector loading visual is too big for the header, pushing the header down.
- The entire page is black which is inconsistent with the rest off the application.
When loading the application, you see a black flash. I suspect this background is the cause of it.

Stijnkool:
# Description

- Added RemoteDataView
- Replace a lot of usage of RemoteData.fold with RemoteDataView
- Improve Spinner & LoadingView
- Improve loading state of EnvironmentControls

closes #1960 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/1960-improve-loading-flow.yml
+++ b/changelogs/unreleased/1960-improve-loading-flow.yml
@@ -1,0 +1,4 @@
+description: Improve loading flow
+issue-nr: 1960
+change-type: patch
+destination-branches: [master, iso4]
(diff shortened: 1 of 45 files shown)


andraskvr (approved):
Looks good :+1:

Issue #2350, opened 2022-01-12
Environment selector is out of sync


andraskvr:
When creating, modifying or deleting environments, the data in the `EnvironmentSelector` is not updated.
This is likely fallout from the navigation refactor. 
(Since there are multiple redux slices holding similar data, some updates are probably missing. Alternatively the two slices for the environment list - with and without icons - can also be merged).

- [ ] make one redux slice for all environment related stuff.

Stijnkool:
The "Select this environment" button also does not work for a newly created environment. Only after a refresh it works.

arnaudsjs:
> The "Select this environment" button also does not work for a newly created environment. Only after a refresh it works.

FYI: I had this issue on Firefox. I didn't test it with Chrome.

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/web-console/2350?planning-poker) with ZenHub @bartv @andraskvr @Stijnkool

PR #2353, opened 2022-01-12
Environment data docs

### LINKED ISSUE #2172: Environment Data is handled in a messy way
We are fetching and storing environments in a lot of different ways.
We have some classes handling related stuff. It is quite difficult to follow.
We should try to simplify this.
Especially the order of things is hard to follow.
What happens first? Maybe make a diagram to make this clear.

- [ ] document the loading mechanisms for environment
- [ ] if possible, simplify 

Timebox to 1 day

andraskvr:
# Description

I found that the latest status of this data handling is generally ok (there have been changes since the ticket was created though, so it might be thanks to that).
I've included the analysis nevertheless. If we decide to do one of the suggested improvements, I'll create a separate ticket for that and remove it from the PR, and if we don't I'll keep it there with some explanation why it wasn't implemented.

closes #2172 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/2172-document-environments.yml
+++ b/changelogs/unreleased/2172-document-environments.yml
@@ -1,0 +1,4 @@
+description: Document environment data usage
+issue-nr: 2172
+change-type: patch
+destination-branches: [master, iso4]
--- a/docs/environment-data.md
+++ b/docs/environment-data.md
@@ -1,0 +1,36 @@
+[Index](./index.md)
+
+# Environment data
+
+## Introduction
+
+The main areas where the frontend uses the environment data are:
+
+- the data layer (communcation with the backend)
+- the UI itself (showing the user which environment are they working in, and the status of it)
+- the url, for shareability
+
+## Analysis
+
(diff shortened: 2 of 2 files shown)


Stijnkool (approved):
Nice overview. 👍 
Yes I remember changing some things related to the environment slices in another ticket.
I introduced the `MERGE` option.
This is why the `GetEnvironments` query for the `Initializer` and `EnvSelector` can use the same slice and query.
I think what improvement 2 describes is already happening.
Because of the `MERGE strategy`, the slice is not reset to `loading`.
And if the call from the `Initializer` already returned data, that will be used.
The call will just be executed again.

We could rename some queries to make it a bit more obvious.
But the way it is now is fine for me. Nothing is incorrectly named.

andraskvr:
Indeed the `MERGE` option is already there, and the current naming is ok for me as well. 
I added my suggestion about the slices to #2350, and we'll see when we get to that whether we implement it or not.

PR #2358, opened 2022-01-12
Issue/2244 diff view docs only

### LINKED ISSUE #2244: Design a diff functionality for desired state
The goal of this ticket is to design/mock/investigate a diff functionality that can compare two desired state versions. 

## Data
Attached to this issue is a set of scripts and data based on our 5g demo. 

The dl.py script downloads the desired state of all versions and resources and places this in the out directory (also in the zip). The diff.py script generates a json based diff between two versions of desired state.

The current proposed structure for the desired state is:
- A mapping of added resources
- A mapping of removed resources
- A mapping of modified resources

The mapping is resource id to attributes. For each attribute the r1 (source version of the diff) and r2 (target version of the diff) is provided + a text based diff.

The r1 and r2 data is always provided. If the resource is removed or added the r1 or r2 has null as value.

For the text based diff, the diff.py script tries to guess some of the types and then convert it to a string representation. For example a list is sorted and then rendered as string with an element per line. Json data is rendered to json, ...

File data is not yet fetched.

[diff.zip](https://github.com/inmanta/web-console/files/7769963/diff.zip)

The diff.json file is generated between version 4110 and 4122

## Visualisation

A good example of a well known UI for visualizing diffs is how for example gitlab (or github) does this. A view where on the side you have a list of all the changed resources using green and red to indicate add, delete or modify and then a diff overview per attribute per resource.

image

This might be a continuous list or per resource.

## Results

- Notes data format, libraries, ...
- Mocks
- Small MVP?

Stijnkool:
# Description

This PR only contains the analysis that was added to the docs.
This way it can be merged in and the ticket closed.

closes #2244 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


andraskvr (approved):
Looks good. We shouldn't forget to update it after we make the final decisions

PR #2370, opened 2022-01-13
Issue/2265 diagnose links

### LINKED ISSUE #2265: Diagnose links in menu still link to old frontend
When there is an error in the diagnose page, the links provided in the kebab menu still point to the old frontend.

These should be updated to the equivalent pages in the web console.

Stijnkool:
# Description

- Cleanup of the dropdown links in Rejection & Failure cards
- Fixed our custom Link component. It was not passing down className and received refs.
This is needed for it to work when passing it as the DropdownItem component prop.
- Wrote doc entry. We should never use `useNavigateTo` for general links.
- Also fixed a bug in the navigateTo logic.
The new search param should always start with a question mark.
If it does not start with a question mark, it throws an exception.

closes #2265 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/UI/Routing/Utils.ts
```diff
@@ -12,18 +12,26 @@ type NavigateTo = (
...
+    if (newSearch !== undefined) validateSearch(newSearch);
     const pathname = routeManager.getUrl(routeKind, params);
-    navigate(`${pathname}?${newSearch || search}`);
+    navigate(`${pathname}${newSearch || search}`);
   };
 };
 
+const validateSearch = (search: string): void => {
+  if (search.startsWith("?")) return;
+  throw new Error("A search string should start with a question mark (?).");
```
andraskvr, line 32: We could also add the question mark if it's not there

andraskvr (approved):

src/UI/Routing/Utils.ts
Stijnkool: I opted not to do that as the usage of `useNavigateTo` is completely in our hands.
We should use our own methods correctly. And when you forget it, an error is thrown.

PR #2372, opened 2022-01-13
Fix image importing

### LINKED ISSUE #2198: Import images directly so they also work for local dev
The background image urls are hardcoded into the frontend component.
They should be imported and webpack should resolve the urls.
Locally they currently only work when fetched from root /.
Once deployed, they always work.
Maybe the server handles these requests differently.

-> Import the image like any other asset or file, and let webpack handle the rest.

andraskvr:
# Description

By using the webpack 5 asset module loading mechanism.

closes #2198 

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/2198-image-import.yml
+++ b/changelogs/unreleased/2198-image-import.yml
@@ -1,0 +1,4 @@
+description: Fix image importing
+issue-nr: 2198
+change-type: patch
+destination-branches: [master, iso4]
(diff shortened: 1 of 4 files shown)


Stijnkool (changes requested):
Looks good.
The webpack config still contains a lot of special rules for images and fonts.
Can't these be removed?
I don't even see where we import fonts.

andraskvr:
>  I don't even see where we import fonts.

They are imported by the `base.css` file from Patternfly, so I would keep the font rule as it is.
We really didn't use some of the svg rules, so I removed those.

Stijnkool (approved):

PR #2385, opened 2022-01-14
Fix background image

### LINKED ISSUE #2201: Background image is no longer shown in the header
The background image is no longer shown after the navigation refactor

andraskvr:
# Description

closes #2201 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


Stijnkool (changes requested):
The background image component is on the same level as the Page.
But it should only be visible behind the header.
I dont think it should be visible on the page body as that should always be white.
Maybe move the background image to the header?

Check this screenshot. The homepage loading state is now "broken".
Maybe the Homepage just needs the PageContainer.
But even than, I think if the background image is only practically visible on the header, it should be location there.

Screenshot 2022-01-14 at 09 46 31

Stijnkool (approved):

PR #2494, opened 2022-01-25
Add missing action to pause and unpause an agent

### LINKED ISSUE #2085: Pause/unpause an individual agent
#1811 mentioned that this would come in a later ticket, but I haven't found it.
It should be possible to pause and unpause one specific agent via the table on the `Agents` page.
The placeholder button is already there.
The v2 `agent_action` endpoint (`/agent/<name>/<action>`) should be used for this. 
This button should be disabled if the environment is halted.

andraskvr:
# Description

closes #2085 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/Core/Command/AgentAction.ts
```diff
@@ -0,0 +1,16 @@
+import { Maybe } from "@/Core/Language";
+import { Query } from "@/Core/Query";
+
+export interface AgentAction {
```
Stijnkool, line 4: The naming of this command is not really in line with the other commands.
It should be a "command", a statement, using a verb.
I know AgentAction is the one used in the backend.
But I don't think we have followed this terminology for the other commands.

Stijnkool (approved):
Looks good. I just think the naming should be improved.

Issue #2542, opened 2022-01-27
Handle large integers correctly in forms


Stijnkool:
Currently the forms handle numbers as numbers.
Form validation for these fields is also done as numbers.
If we need to support large numbers here, we need to change this.

Options:
- We could use a string representation.
  - Validation needs to be changed
  - When sending data to the API, it needs to be a number again.
This can touch quite some code as the api layer is kept dumb in regards to the data that is transmitted.
It receives data, and sends it along. Turning data into a json string is done implicitly by the Fetch API.
We cant just send a big number to the api layer because this number would lose precision.
If we send a string representing a number then it needs to be turned back into an actual number again.
This might introduce quite some complexity.
- We could use the BigInt native type
  - Validation needs to be changed
  - actual form input might still be a string
  - We can just send a bigInt value to the api layer and it should work.
  - JSON stringifying a bigint is not supported by default.
We can implement is easily as shown in this thread: https://github.com/GoogleChromeLabs/jsbi/issues/30
  - To support BigInt, we need to change our compile target to ES2020.
Currently we compile down to ES2015. There might be other solutions possible.

Both options introduce some complexity and quite some changes.
In general I think supporting BigInt is the way to go.
I think it will introduce the least amount of complexity and will also be the most "correct".

Should all this special handling of large integers be configurable?

1. Support bigint in the number input field
2. Add rationale to design documentation

bartv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/web-console/2542?planning-poker) with ZenHub @andraskvr @Stijnkool

PR #2543, opened 2022-01-27
Issue/203 big numbers

### LINKED ISSUE #203: The web console looses precision on large integers.
When a service entity has an integer attribute it is possible that the API returns values that are to large for javascript integers. An example is an order_id attribute which has value 90161132727657997171

There is no simple fix for this because this data loses precision when the json is loaded from the API. There are a number of solutions:

1. Try to detect it and show it to the user
2. Use a json library that supports large integers: https://www.npmjs.com/package/json-bigint However this needs to be investigated carefully because a bad json library is bad for performance and can create security issues.
3. Let the API also return "string" versions of the value.

For 2 it is also important to think about this when creating or updating instances through the API.

also possible: https://www.npmjs.com/package/lossless-json

We need to evaluate the reliability of the libraries we use.

Stijnkool:
# Description

This does not touch the handling of big numbers in forms.
So creating and editing service instances still needs work.
I have created a separate ticket for this as it seemed out of scope. (https://github.com/inmanta/web-console/issues/2542)
So this ticket focuses simply adding support for large integers coming from the api into the application.

closes #203 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/Core/Contracts/JsonParser.ts
```diff
@@ -0,0 +1,4 @@
+export interface JsonParser {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  parse(text: string): any;
```
andraskvr, line 3: Maybe `Record<string, unknown>` instead of `any`?

andraskvr (approved):
Looks good

src/Core/Contracts/JsonParser.ts
Stijnkool: My first attempt was `unknown`. But the native json.parse method returns the any type.
So this way it is more inline with what was already there.

Also, the api layer does not do any data structure validation.
So currently the `any` type does not complain when we actually want to return a domain type.
If we introduce `unknown` now, we would be typecasting anyway from unknown to an unvalidated assumed domain type.
I agree we should stay away from `any` as much as possible.
But I'd rather introduce a change like this explicitly with a ticket that has a goal of improving the type safety of the api layer.

PR #2670, opened 2022-02-04
Fix file fetcher

### LINKED ISSUE #2622: FileFetcher doesn't work correctly
The `FileFetcher` expects the response from the `/v1/file/<id>` endpoint to have the content in an envelope called `data`.
Since this is a v1 endpoint, the `data` envelope is not used, the access to the content should be just `response.content` instead of `response.data.content` in the `unpack` method:
 https://github.com/inmanta/web-console/blob/f64ad30fc9055d9b40d0cb653283213b1e7803ee/src/Data/API/FileFetcherImpl.ts#L44

andraskvr:
# Description

closes #2622 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/Data/API/FileFetcherImpl.test.ts
+++ b/src/Data/API/FileFetcherImpl.test.ts
@@ -1,12 +1,21 @@
-import { BaseApiHelper } from "./BaseApiHelper";
+import { DeferredApiHelper } from "@/Test";
 import { FileFetcherImpl } from "./FileFetcherImpl";
 
 test("GIVEN FileFetcher WHEN executing get THEN the result is correctly returned", async () => {
-  const fileFetcher = new FileFetcherImpl(new BaseApiHelper(), "env");
-  fetchMock.mockResponse(`{"content": "abcdefgh"}`);
-  const response = await fileFetcher.get("id1");
-  if (response.kind === "Left") {
-    throw Error("Fetching file resulted in an error");
-  }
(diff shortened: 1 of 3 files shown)


src/Data/API/FileFetcherImpl.test.ts
```diff
@@ -0,0 +1,12 @@
+import { BaseApiHelper } from "./BaseApiHelper";
+import { FileFetcherImpl } from "./FileFetcherImpl";
+
+test("GIVEN FileFetcher WHEN executing get THEN the result is correctly returned", async () => {
+  const fileFetcher = new FileFetcherImpl(new BaseApiHelper(), "env");
+  fetchMock.mockResponse(`{"content": "abcdefgh"}`);
```
Stijnkool, line 6: I do not really see a reason for using the fetchMock api here.
Th only reason for using that would be to test the BaseApiHelper.
Here I would just use our DeferredApiHelper.
Then you can also assert on the request being done, increasing the test coverage.

Stijnkool (changes requested):

src/Data/API/FileFetcherImpl.test.ts
andraskvr: I don't think it improves the coverage that much, but it is more consistent with the other test cases. I applied the changes

PR #2734, opened 2022-02-14
Speed up for forms slow on firefox

### LINKED ISSUE #2157: Instance Form slow on Firefox
Followup ticket of #1875
The Create/Edit service instance form is slow on Firefox, when having nested embedded entities.
This is because Patternfly uses `grid` layout for the `FormFieldGroupExpandable` component that we use for showing the inputs for the attributes, and Firefox currently doesn't handle the deeply nested grid layout.
Issue on Firefox bug tracker:
https://bugzilla.mozilla.org/show_bug.cgi?id=1682686

There are a few alternatives for us in this case:

1. Wait for this underlying issue in Firefox to be fixed
2. Recreate the Patternfly component using another layout that doesn't have this problem and looks similar, for example using flexbox
3. Change the styling of this form altogether

andraskvr:
# Description

related to #2157

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/2157-form-speedup-firefox.yml
+++ b/changelogs/unreleased/2157-form-speedup-firefox.yml
@@ -1,0 +1,4 @@
+description: Speed up for forms slow on firefox
+issue-nr: 2157
+change-type: patch
+destination-branches: [master, iso5, iso4]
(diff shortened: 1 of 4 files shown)


andraskvr:
The workaround suggested in https://bugzilla.mozilla.org/show_bug.cgi?id=1682686 does speed up the form significantly, but it still gets a bit slower with each new `grid`. So for the majority of cases it will work fine, but if it's a very complicated and nested model, it can still slow down (only later). @bartv Should we create a ticket for that / keep this open?

Stijnkool (approved):

bartv:
lets keep this one open en move it back to icebox once this is merged

PR #2769, opened 2022-02-15
Issue/2720 remove processing events

### LINKED ISSUE #2720: Processing events is not allowed as a filter but is reported as a state

Stijnkool:
# Description

closes #2720 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/Test/Data/Resource.ts
+++ b/src/Test/Data/Resource.ts
@@ -39,7 +39,7 @@ export const response = {
         resource_id_value: "/etc/file2",
       },
       requires: [],
-      status: "orphaned",
+      status: "processing_events",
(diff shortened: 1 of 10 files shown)


src/Test/Data/Resource.ts
```diff
@@ -96,7 +96,6 @@ export const response = {
         undefined: 2,
         deploying: 2,
         available: 1,
-        processing_events: 0,
```
andraskvr, line 99: Maybe add a test case that if it is actually in the response, the application can still handle it?

andraskvr (approved):

PR #2803, opened 2022-02-18
Fix security alert related to node-fetch

andraskvr:
# Description

Resolve node-fetch to a version that doesn't have the vulnerability mentioned in https://github.com/inmanta/web-console/security/dependabot/7

# Self Check:

- [x] ~Attached issue to pull request~
- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,7 +165,9 @@
     "nth-check": "^2.0.1",
     "browserslist": "^4.16.5",
     "json-schema": "^0.4.0",
-    "node-forge": "^1.0.0"
+    "node-forge": "^1.0.0",
+    "cross-fetch": "^3.1.5",
+    "node-fetch": "^2.6.7"
   },
   "madge": {
     "fileExtensions": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -165,7 +165,9 @@
     "nth-check": "^2.0.1",
     "browserslist": "^4.16.5",
     "json-schema": "^0.4.0",
-    "node-forge": "^1.0.0"
+    "node-forge": "^1.0.0",
+    "cross-fetch": "^3.1.5",
+    "node-fetch": "^2.6.7"
```
Stijnkool, line 170: This list keep getting longer and longer. I suppose at some point when the libraries using these deps are updated, these deps no longer have to be mentioned here. But that is a manual revision I gues.

Stijnkool (approved):

package.json
andraskvr: Yes, you are correct, we also have to keep an eye on that.

PR #2813, opened 2022-02-18
Improve error message format

### LINKED ISSUE #2719: Pydantic error messages are unreadable in the web-console
When the backend provides an error this is often entirely unreadable. The goal of this ticket is twofold:

1. When error messages contain newlines, render these in the browser so that the content is readable. For example:

2. Check if pydantic provides more structured ways of reporting errors and how they can be represented in the web-console. The result is a ticket in the backend to improve error messages and a frontend ticket for explaining errors.

andraskvr:
# Description

spin-off backend ticket: inmanta/inmanta-core#3850

closes #2719 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/Data/API/BaseApiHelper.test.ts
+++ b/src/Data/API/BaseApiHelper.test.ts
@@ -15,6 +15,22 @@ test("BaseApiHelper.get executes a GET request with correct url & env", async ()
   expect(requestInit?.headers).toEqual({ "X-Inmanta-Tid": env });
 });
 
+test("BaseApiHelper.get handles a failed a GET request", async () => {
+  const apiHelper = new BaseApiHelper();
+  const url = "/test-url";
+  const env = "environment_a";
+
+  fetchMock.mockResponse(JSON.stringify({ message: "Something happened" }), {
+    status: 400,
+  });
+  const response = await apiHelper.get(url, env);
+  expect(response).toEqual(
+    Either.left(
+      `The following error occured while communicating with the server: 400 Bad Request \nSomething happened`
+    )
+  );
+});
+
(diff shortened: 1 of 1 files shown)


Stijnkool (changes requested):
I think it would be good if we had some storybook entries for these components with some actual complex error messages. So we have an example of what this looks like.
Ideally this code uses the format error function which has changed.

This is unrelated to this PR, but I guess the error handling should not be done in the BaseApiHelper. It should be done in a separate class so our test implementations can actually use it. Increasing test coverage again.

Stijnkool (approved):

andraskvr:
> I think it would be good if we had some storybook entries for these components with some actual complex error messages. So we have an example of what this looks like. Ideally this code uses the format error function which has changed.

I added some stories, for the service inventory and the alert itself.

> This is unrelated to this PR, but I guess the error handling should not be done in the BaseApiHelper. It should be done in a separate class so our test implementations can actually use it. Increasing test coverage again.

Well sending requests and handing the results back to the caller is what we use the `BaseApiHelper` for, and deciding whether it's successful or not is really closely related to that (but it would be possible to extract it). 
Formatting the errors though, can be considered a different responsibility more clearly, especially if we add some more logic to it

Stijnkool:
>  Well sending requests and handing the results back to the caller is what we use the BaseApiHelper for, and deciding whether it's successful or not is really closely related to that (but it would be possible to extract it).
> Formatting the errors though, can be considered a different responsibility more clearly, especially if we add some more logic to it
Yea, for me the purpose is mainly that the code doing fetch calls needs to be in 1 class so we can replace it with a mocked implementation during testing. All the other code needs to be detached from that so it can be tested.

PR #2851, opened 2022-02-24
Refresh environment details on settings change

### LINKED ISSUE #2792: When changing an environment setting, environment details in the rest of the app should be updated.
When changing an environment setting, we should force a refetch of the environment details used in the rest of the app.
Example:
Switching the `server_compile` setting from `true` to `false`.
The setting is saved. The data in the environmentDetails slice is still outdated.
When you then go to the DesiredState page, you see the compile button enabled.
And after some seconds (when the next scheduler loop fires), the button becomes disabled.

This is a case of stale data. This is fixable.
- We have the moment when the data is updated (after the setting save call is done)
- We should trigger a refetch of the environment details data.
- Store slice entry should go to loading because otherwise we are still showing stale data.
- Maybe we should move the environmentSetting data into the environments slice. It's all related data.

Maybe the staleness of this data is not really a problem and there is no fix required.

andraskvr:
# Description

closes #2792 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/Data/Store/EnvironmentSlice.ts
```diff
@@ -65,4 +83,36 @@ export const environmentSlice: EnvironmentSlice = {
...
+  setSettingsData: action((state, { environment, value, merge }) => {
+    state.settingsByEnv[environment] = merge
+      ? mergeData(state.settingsByEnv[environment], value)
+      : value;
+  }),
 };
+
+function mergeData(
+  prev: RemoteData.Type<
+    Query.Error<"GetEnvironmentSettings">,
```
Stijnkool, line 96: Maybe use a type variable for this as its the same gigantic type 3 times.

Stijnkool (approved):
Just 1 remark. Nice 👍

PR #2855, opened 2022-02-24
Issue/2495 api helper resolve

### LINKED ISSUE #2495: Add a matchable resolve method to DeferredApiHelper
Currently the resolve method just resolves the next pending request.
It would be good if we have another resolve method that is capable of matching a specific request.
It might also be useful to have an ignore method which is capable of simply removing a pending request if we don't care about it.
This is useful in test cases where some components on a page also fire of requests.
But these components are not the focus of this specific test.

- [ ] Add ignore method for specific requests
- [ ] Add matchable resolve method to resolve specific requests

Stijnkool:
# Description

closes #2495 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/Test/Mock/DeferredApiHelper.test.ts
+++ b/src/Test/Mock/DeferredApiHelper.test.ts
@@ -84,3 +84,20 @@ test("GIVEN DeferredApiHelper WHEN resolveRequest THEN resolves first matching r
     data: { test: "def" },
   });
 });
+
+test("GIVEN DeferredApiHelper WHEN resolveRequest THEN resolves first matching request", async () => {
+  const helper = new DeferredApiHelper();
+  expect(() =>
+    helper.resolveRequest(
+      { method: "GET", url: "url2", environment: "env2" },
+      { test: 123 }
+    )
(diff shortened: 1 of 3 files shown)


src/Test/Mock/DeferredApiHelper.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  expect(helper.pendingRequests).toEqual([
+    { method: "GET", url: "url2", environment: "env2" },
+  ]);
+  expect(helper.resolvedRequests).toHaveLength(1);
+  expect(helper.resolvedRequests).toEqual([
+    { method: "GET", url: "url1", environment: "env1", data: { test: 123 } },
+  ]);
+});
+
+test("GIVEN DeferredApiHelper WHEN resolveRequest THEN resolves first matching request", async () => {
```
andraskvr, line 54: Can you add a test case when there is no matching request?

andraskvr (changes requested):
One comment otherwise looks good

andraskvr (approved):

PR #2899, opened 2022-02-28
Render json strings in popover of attributes table

### LINKED ISSUE #2801: Render json strings in the service instance attributes
Because of limitations in the API we store JSON in some attributes. These are shown as strings, however this is not very readable for users.

The goal of this ticket is to render a value as json when it is valid json. Do a stringify

andraskvr:
# Description

closes #2801 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/UI/Components/TreeTable/TreeRow/CellWithCopy.tsx
+++ b/src/UI/Components/TreeTable/TreeRow/CellWithCopy.tsx
@@ -32,12 +32,15 @@ export const CellWithCopy: React.FC<Props> = ({ label, value, className }) => {
       {value}
     </Td>
   );
+  const formattedValue = isJson(value)
+    ? JSON.stringify(JSON.parse(value), null, 2)
+    : value;
 
   return wrapWithPopover ? (
     <Popover
       bodyContent={
         <>
-          <StyledPopoverBody>{value}</StyledPopoverBody>
+          <StyledPopoverBody>{formattedValue}</StyledPopoverBody>
           <StyledButton
             value={value}
             tooltipContent={words("attribute.value.copy")}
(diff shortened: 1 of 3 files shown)


src/UI/Components/TreeTable/TreeRow/CellWithCopy.tsx
```diff
@@ -32,12 +32,15 @@ export const CellWithCopy: React.FC<Props> = ({ label, value, className }) => {
       {value}
     </Td>
   );
+  const formattedValue = isJson(value)
```
Stijnkool, line 35: Would it not be simpler if you had a function `formatValue` that just returns this string? The `isJson` function can be a part of that. Just an idea. Not required.

Stijnkool (approved):
👍

src/UI/Components/TreeTable/TreeRow/CellWithCopy.tsx
andraskvr: It was a good idea to extract that as well :+1:

src/UI/Components/TreeTable/TreeRow/CellWithCopy.tsx
```diff
@@ -32,9 +32,7 @@ export const CellWithCopy: React.FC<Props> = ({ label, value, className }) => {
       {value}
     </Td>
   );
-  const formattedValue = isJson(value)
-    ? JSON.stringify(JSON.parse(value), null, 2)
-    : value;
+  const formattedValue = formatValue(value);
```
Stijnkool, line 35: You could even inline this function call. No need for a variable.

PR #2907, opened 2022-02-28
Improve loading visual for Resources Page

### LINKED ISSUE #2881: Fix strange loading during paging on the resource page
The resources table loads very strange when doing paging. Cfr discussions of a few weeks ago when talking about the performance.

Only the table should reload. The progress bar and the paging components should stay visible, like the filter.

Stijnkool:
# Description

closes #2881 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/UI/Pages/Resources/Page.tsx
+++ b/src/UI/Pages/Resources/Page.tsx
@@ -1,5 +1,5 @@
-import React, { useContext } from "react";
-import { Resource } from "@/Core";
+import React, { useContext, useEffect, useState } from "react";
+import { RemoteData, Resource } from "@/Core";
 import {
   useUrlStateWithFilter,
   useUrlStateWithPageSize,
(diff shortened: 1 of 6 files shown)


src/UI/Pages/Resources/Page.tsx
```diff
@@ -43,28 +43,32 @@ export const Page: React.FC = () => {
     pageSize,
   });
 
+  const [staleData, setStaleData] = useState(data);
+
+  useEffect(() => {
+    if (RemoteData.isLoading(data)) return;
+    setStaleData(data);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [data.kind]);
```
andraskvr, line 52: Could you add a test case, that verifies that if the summary data is updated, the summary widget is still correctly updated? 
(So the data is not stale all the time)

andraskvr (changes requested):
Looks good, one suggestion

src/UI/Pages/Resources/Page.tsx
Stijnkool: 👍

andraskvr (approved):

PR #2929, opened 2022-03-01
Make the columns of the attributes table collapsible

### LINKED ISSUE #699: Add collapsable columns to nested attribute table
This might be useful. Might be a UI experiment. 

When a set is empty it should be collapsed by default.

This should be "paired" with a zoom function as well that only shows the zoomed column and collapses the others by default

Should also investigate the pf component for this.

andraskvr:
# Description

closes #699 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/699-collapsible-columns.yml
+++ b/changelogs/unreleased/699-collapsible-columns.yml
@@ -1,0 +1,4 @@
+description: Make the columns of the attributes table collapsible
+issue-nr: 699
+change-type: minor
+destination-branches: [master, iso5]
--- a/src/UI/Components/TreeTable/ColumnHeaders.tsx
+++ b/src/UI/Components/TreeTable/ColumnHeaders.tsx
@@ -1,0 +1,96 @@
+import React, { useState } from "react";
+import { Button } from "@patternfly/react-core";
+import { CaretLeftIcon, CaretRightIcon } from "@patternfly/react-icons";
+import { Th, ThProps } from "@patternfly/react-table";
+import styled from "styled-components";
+import { ColumnExpansionHelper } from "./Helpers";
+
+interface Props {
+  columns: string[];
+}
+
+export const ColumnHeaders: React.FC<Props> = ({ columns }) => {
+  const columnExpansionHelper = new ColumnExpansionHelper(
+    60,
+    columns.length,
+    10
+  );
+  const [columnWidths, setColumnWidths] = useState<Record<string, number>>(
+    columnExpansionHelper.getDefaultState(columns)
+  );
+  return (
+    <>
+      {columns.map((column) => (
+        <ColumnHeader
+          key={column}
+          column={column}
+          width={columnWidths[column]}
+          minWidth={columnExpansionHelper.getMinColumnWidth()}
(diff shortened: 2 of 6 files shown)


Stijnkool (commented):
Some small code remarks. None of the remarks are really required though.

Some thoughts on the design:
Screenshot 2022-03-01 at 15 15 42

The ticket says empty columns should be collapsed by default.
I do not seem to see this behaviour. Maybe you missed it.

Currently there are 2 icons per column. These give u 2 scenarios:
- zoom in on 1 column and auto collapse the others 👍 
- collapse 1 column and split the remaining width between the remaining columns 🤷‍♂️ 
I dont think this last option is super useful.
What do you think? @bartv

The zoom functionality is there. But it was not immediately obvious to me.
I understand the usage of the current icons. But I think a zoom icon or collapse icon is more useful.

My suggestion is to make the column header text clickable.
- clicking it zooms in on that column
- clicking it again goes back to default view (empty columns collapsed)
- maybe provide an expand/collapse or zoom-in/zoom-out icon based on the status of the column. But make the whole thing clickable.
- This goal is for it to work more like the current "sorting on column" behaviour

A small bug:
- When the rollback column is collapsed, the icon is gone, but I can still click it.
- This is quite confusing
Screenshot 2022-03-01 at 15 14 18
 
@andraskvr @bartv overall thoughts?

andraskvr:
> The ticket says empty columns should be collapsed by default. I do not seem to see this behaviour. Maybe you missed it.

I forgot that indeed, thanks for the reminder :+1: 
 
>     * collapse 1 column and split the remaining width between the remaining columns 
>       I dont think this last option is super useful.

This way a user could collapse one column if they are only interested in comparing the values of the other two, but this may not be required

> The zoom functionality is there. But it was not immediately obvious to me. I understand the usage of the current icons. But I think a zoom icon or collapse icon is more useful.

Our icon set doesn't really have good collapse icons, zoom could work.

> My suggestion is to make the column header text clickable.
> 
>     * clicking it zooms in on that column
> 
>     * clicking it again goes back to default view (empty columns collapsed)
> 
>     * maybe provide an expand/collapse or zoom-in/zoom-out icon based on the status of the column. But make the whole thing clickable.
> 
>     * This goal is for it to work more like the current "sorting on column" behaviour

I like the idea in, but I think in it's current form it's only viable if we don't need the functionality to explicitly collapse a column, only to expand. I'll give it some more thought though


> A small bug:
> 
>     * When the rollback column is collapsed, the icon is gone, but I can still click it.
> 
>     * This is quite confusing

Nice catch

Stijnkool (commented):
Looks great 👍 
Screenshot 2022-03-02 at 12 33 44
Some small remarks maybe:
- The Name header is styled differently and vertically aligned differently then the others. This looks a bit odd.
- When click a column, the text becomes black. But it's only because of focus. When you click away, the black becomes grey again. It is strange that this focus color is there as it doesn't really mean anything. Not sure what the patternfly guidelines are for signifying that a column header is clickable.... The sorting has the arrows. Maybe a zoom icon...?

Stijnkool (approved):
Looks great! 👍

PR #2969, opened 2022-03-02
Issue/2880 resource deps loading

### LINKED ISSUE #2880: Show a loading state when opening the dependencies on the resources table
When the number of resources is opened it will show a white block and only after the data is fetched show the actual data.

Most other pages use a loading state.

# Todo
- [x] There already is a loading mechanism. Lower the delay from 500ms to 200ms.
- [x] The loading view should be equally high as the resulting table. (We know the number of dependencies so we know the number of rows)
- [x] Try to add the Skeleton loading component to the requires table. This can replace the default Loading spinner. Which is too big anyway when there are only 1 or 2 rows.

Stijnkool:
# Description

closes #2880 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/UI/Pages/Resources/Components/RequiresTableWithData.tsx
+++ b/src/UI/Pages/Resources/Components/RequiresTableWithData.tsx
@@ -1,15 +1,20 @@
 import React, { useContext } from "react";
+import { RemoteData } from "@/Core";
+import {
+  RequiresTable,
+  LoadingRequiresTable,
+  ErrorView,
+} from "@/UI/Components";
 import { EmptyView } from "@/UI/Components/EmptyView";
-import { RemoteDataView } from "@/UI/Components/RemoteDataView";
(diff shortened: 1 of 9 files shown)


src/UI/Pages/Resources/Components/RequiresTableWithData.tsx
```diff
@@ -1,27 +1,43 @@
...
+      loading: () => (
+        <LoadingRequiresTable
+          numberOfRows={deps}
+          aria-label="ResourceRequires-Loading"
+        />
+      ),
+      failed: (error) => (
+        <ErrorView
+          message={words("error.general")(error)}
+          aria-label="ResourceRequires--Failed"
```
andraskvr, line 37: The double dash is a typo I guess

andraskvr (approved):
Cool :sunglasses:

PR #2972, opened 2022-03-02
Improve action on resource table

### LINKED ISSUE #2879: Improve actions on resources table
The resource table has a button at the end of the row which does not in line with the patternfly design guidelines.

The goal of this ticket is to:

1. Remove the action header from the column
2. Replace the button with an action link
3. Use an action "Show details" instead of "Details"

Stijnkool:
# Description

closes #2879 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/UI/Pages/Resources/ResourceTableRow.tsx
+++ b/src/UI/Pages/Resources/ResourceTableRow.tsx
@@ -1,15 +1,14 @@
-import React, { useContext } from "react";
+import React from "react";
 import { Button } from "@patternfly/react-core";
 import { AngleDownIcon, AngleRightIcon } from "@patternfly/react-icons";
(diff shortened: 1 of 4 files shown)


src/UI/Pages/Resources/ResourceTableRow.tsx
```diff
@@ -46,20 +46,15 @@ export const ResourceTableRow: React.FC<Props> = ({
...
-        <Td>
+        <StyledCell>
           <Link
-            to={{
-              pathname: routeManager.getUrl("ResourceDetails", {
-                resourceId: row.id,
-              }),
-              search: location.search,
-            }}
+            pathname={routeManager.getUrl("ResourceDetails", {
```
andraskvr, line 51: I think there is a `ResourceLink` component that also leads to the `ResourceDetails` page, we could reuse that here

PR #2989, opened 2022-03-02
Show warning when no attributes for EditInstance page

### LINKED ISSUE #2865: Disable the edit button when there are not attribute to edit
Currently the edit button is enabled, even if there are not attribute to edit.

The options are:

1. Keep it but at least mention it and enable the confirm button. This allows to move the service into an update state.
2. Disable the button entirely.

Stijnkool:
# Description

I am not really sure about enabling the confirm button when the environment is halted. Is this correct?
I am missing some context as to what this actually means.

closes #2865 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/2865-edit-instance-no-attributes.yml
+++ b/changelogs/unreleased/2865-edit-instance-no-attributes.yml
@@ -1,0 +1,4 @@
+description: Show warning when no attributes for edit instance page
+issue-nr: 2865
+change-type: patch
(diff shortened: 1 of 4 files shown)


andraskvr:
iirc it should still be disabled when the environment is halted, and the goal is that when it isn't halted, and there aren't any attributes to edit, we show a message instead of the middle part (between the header and the buttons) being empty

andraskvr (approved):

PR #2992, opened 2022-03-03
Show resource id without version in resources table

### LINKED ISSUE #2870: Only show the resource id str (not version str) in the resources tab
In the resources tab of the service inventory the full resource version id is shown. This should become only the resource id.

The name of the column should also just be "Resource" and not "Resource id"

andraskvr:
# Description

closes #2870 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/2870-resource-id-in-table.yml
+++ b/changelogs/unreleased/2870-resource-id-in-table.yml
@@ -1,0 +1,4 @@
+description: Show resource id without version in resources table
+issue-nr: 2870
+change-type: patch
+destination-branches: [master, iso5]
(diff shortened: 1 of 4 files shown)


changelogs/unreleased/2870-resource-id-in-table.yml
```diff
@@ -0,0 +1,4 @@
+description: Show resource id without version in resources table
+issue-nr: 2870
+change-type: patch
+destination-branches: [master, iso5]
```
Stijnkool, line 4: Can't this easily go into iso4? iirc, this sprint was mostly about fixing small stuff that could go into iso4 as well.

Stijnkool (approved):

PR #3054, opened 2022-03-09
Make agents table styling consistent with the other tables

### LINKED ISSUE #2998: Improve agent table consistency
1. Remove actions header
2. Align actions + kebab to the right of the table
3. Use links for the processes instead of buttons
3. Rename the unpause header to "On resume"
4. The size of the columns is strange (see screenshot 2). Non of the items in the first column (even if you scroll down) fill the first columns. But still the Unpause header is ellipsed.

andraskvr:
# Description

closes #2998

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


Stijnkool (approved):

Stijnkool (changes requested):
I checked the patternfly components.
We don't have to style these cells ourselves.

```
<Td modifier="fitContent" />
<Td isActionCell />
```
See https://www.patternfly.org/v4/components/table/#composable-actions

src/UI/Pages/Agents/Actions.tsx
```diff
@@ -28,6 +28,7 @@ export const Actions: React.FC<Props> = ({ name, paused }) => {
         isOpen={isOpen}
         isPlain
         onSelect={() => setIsOpen(false)}
+        position="right"
```
Stijnkool, line 31: This should not be necessary.

src/UI/Pages/Agents/AgentsTableRow.tsx
```diff
@@ -26,26 +25,24 @@ export const AgentsTableRow: React.FC<Props> = ({ row }) => {
...
         </Td>
-        <Td dataLabel={words("agents.columns.unpause")}>
+        <Td width={10} dataLabel={words("agents.columns.unpause")}>
           {row.unpause_on_resume !== null && row.unpause_on_resume !== undefined
             ? JSON.stringify(row.unpause_on_resume)
             : null}
         </Td>
-        <Td dataLabel={words("agents.columns.actions")}>
+        <Td isActionCell modifier="fitContent">
           <Actions name={row.name} paused={row.paused} />
```
Stijnkool, line 46: The Actions component should probably be split up. I know this is not really described in this ticket.
But the patternfly tables use a different `Td` for the action link and the kebap dropdown.
I would split up the actions component and put them in separate `Td`s. The link should be `fitContent`.
The kebap should be `isActionCell`.... if we want to do it like patternfly of course.

src/UI/Pages/Agents/AgentsTablePresenter.ts
```diff
@@ -20,9 +20,8 @@ export class AgentsTablePresenter implements TablePresenter<Agent, AgentRow> {
         displayName: words("agents.columns.unpause"),
         apiName: "unpause_on_resume",
       },
-      { displayName: words("agents.columns.actions"), apiName: "actions" },
     ];
-    this.numberOfColumns = this.columnHeads.length + 1;
+    this.numberOfColumns = this.columnHeads.length + 2;
```
Stijnkool, line 24: If you were to add another `Td` this would become +3

Stijnkool (approved):
If we want to do it exactly like patternfly, some changes are required.
Up to you.

src/UI/Pages/Agents/Actions.tsx
andraskvr: Since it's on the right edge, it is, otherwise it can be partially hidden (maybe it's different on mac?):
Screenshot from 2022-03-10 11-53-54

PR #3070, opened 2022-03-10
Issue/2861 resources error inventory

### LINKED ISSUE #2861: Resources list in the service inventory shows errors on transition
When the resources (maybe other data as well) is open en the service transitions it is possible that the refresh shows an error. This is probably caused by first fetching the details before updating the instance state. This results in the wrong version being sent to the server.

I have only seen it here but probably other views suffer from it as well.

## Technical
It is quite complex to make the resources call be dependent on the latest executed instance call as they are both registered within the same scheduler and executed at the same time. Registered tasks within the scheduler do not have any relation to each other. They are just executed in the order the are registered.

## Solution
We cannot prevent the resources request containing on outdated version. Therefor we cannot prevent the request from
returning a 409 Conflict error response. So when we get this specific error, we know we sent the request with an outdated version. So we need to refetch the latest instance, use that version to refetch the resources.

The InstanceResources QueryManager has to be changed to handle this behaviour. When everything is successful, we just need the current default behaviour of the `QueryManager.ContinuousWithEnv` class. But when the resources request fails with a 409, we need to trigger the new behaviour.

New behaviour:
- unregister query from scheduler
- fetch latest instance
- fetch resources based on response
- if successful, show data and register query with scheduler
- if unsuccessful, redo the loop until we are successful.
We can limit the amount of retries we do to 20 for example.
When we reach the limit, we switch to the default behaviour.
Default behaviour:
Where there is an error, just show it.
Query is registered with scheduler so it updated every 5s.

The ApiHelper currently only returns an error string. When the request was not successful, we need to return more information so a QueryManager can use that data for more complex logic.

Nothing has to change on the component level.

## Todo
- [x] Adapt the ApiHelper to also return the HTTP StatusCode upon error
- [x] Create a custom InstanceResources QueryManager
- [x] QueryManager: Add loop with limit
- [x] QueryManager: Add test for loop limit
- [x] QueryManager: register with scheduler on success
- [ ] Redux Store: Update the instance in the inventory list on success (keep data in sync)

butterybread:
# Description

closes #2861 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -7,3 +7,4 @@
 - [Notes](./notes.md)
 - [Desired State Diff](./desired-state-diff/index.md)
 - [BigInt](./big-int.md)
+- [Instance Resources](./instance-resources.md)
--- a/docs/instance-resources.md
+++ b/docs/instance-resources.md
@@ -1,0 +1,29 @@
+[Index](./index.md)
+
+# Instance Resources
+
+In the service inventory, you can check the resources of an instance.  
+The request for the resources required the latest instance version.  
+When the instance is transitioning to newer versions, it is hard to pinpoint the latest version.  
+If we send the request for the resources with an outdated version, we get a 409 Conflict error response.
+
+This is not helpful for the user. The user just wants to see the data.  
+This API endpoint might change in the future. But for now we have implemented a fix in the frontend.  
+Most of this logic is implemented in the `QueryManager` for the `GetInstanceResources` query.
+
+When we get a conflict error for the resources request, we fetch the instance again to get the latest version.  
+We fetch the resources again with the new instance version. If it fails again, we repeat until it is successful.  
+We have a `retryLimit` of 20 so we don't keep retrying indefinitely.
+
+The scheduler mechanism is still active for this query.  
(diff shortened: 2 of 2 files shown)


src/Data/Managers/GetInstanceResources/QueryManager.test.tsx
```diff
@@ -0,0 +1,275 @@
...
+  // 3rd resources
+  expect(apiHelper.pendingRequests).toEqual([resourcesRequest(3)]);
+  await act(resolveAs.resourcesConflict);
+  expect(
+    await screen.findByRole("generic", { name: "Dummy-Failed" })
+  ).toBeVisible();
+  expect(apiHelper.pendingRequests).toHaveLength(0);
+});
+
+test("Given the InstanceResourcesQueryManager Then a task is registered on the scheduler", async () => {
```
andraskvr, line 174: The task is now always registered. The test is about handling a conflict in the registered task

andraskvr (approved):
I think it's ok, but adding some documentation would be a good idea I think

PR #3090, opened 2022-03-14
Add status field to compile report page

### LINKED ISSUE #2996: Add a status field to the compile report table
The compile report table now makes the status implicit by using colored rows. This is not consistent with other view and make the filter not clear.

The goal of the ticket is to:

- Split out the Finished status (no longer a finished state) in to Success and Failed in status and remove the result
- Use one filter box that translates this to the correct api filters
- Add a field that shows the status of the report (In Progress, Success, Failed and Queued)
- Use labels for the status (green, red and blue for in progress and white/grey for queued)
- Show the waiting indicator next to the In Progress status
- Remove the first column

andraskvr:
# Description

closes #2996 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/UI/Pages/CompileReports/Components/StatusLabel.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { Label } from "@patternfly/react-core";
+
+import { CompileStatus } from "@/Core";
+
+export const StatusLabel: React.FC<{ status: CompileStatus }> = ({
```
Stijnkool, line 6: I think we discussed during the planning meeting that the label should be like the label in the dryrun dropdown.
So the loading spinner should be inside the label.
Maybe that component can be turned into something reusable.
I think all our current status labels are lowercase, so I think this one should be lowercase as well.

src/UI/Pages/CompileReports/Components/StatusLabel.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { CompileStatus } from "@/Core";
+import { Spacer } from "@/UI/Components";
+import { StatusLabel } from "./StatusLabel";
+
+export default {
+  title: "CompileReport/StatusLabel",
+  component: StatusLabel,
+};
+
+export const Default = () => (
```
Stijnkool, line 11: The progress spinner should be visible in storybook as well.
But that will automatically be fixed.

src/UI/Pages/CompileReports/FilterWidget/StatusFilter.tsx
```diff
@@ -0,0 +1,38 @@
...
+  selected: string | null;
+  setSelected: (selected: string | null) => void;
+}
+
+export const StatusFilter: React.FC<Props> = ({
+  isVisible,
+  selected,
+  setSelected,
+}) => {
+  const compileStatuses = Object.keys(CompileStatus).map(
```
Stijnkool, line 18: You could also have a namespace `CompileReports`.
And then You can have `CompileReports.Status` & `CompileReports.StatusList`.
But not really required.

src/UI/Pages/CompileReports/FilterWidget/CompileReportsFilterWidget.tsx
```diff
@@ -17,27 +16,15 @@ export const CompileReportsFilterWidget: React.FC<Props> = ({
...
-        selectedCompileStatuses.length > 0
-          ? selectedCompileStatuses.map(
-              (compileStatus) =>
-                CompileReportParams.CompileStatus[
-                  compileStatus.replace(/\s+/g, "")
-                ]
-            )
-          : undefined,
+      status: selectedCompileStatus
+        ? CompileStatus[selectedCompileStatus.replace(/\s+/g, "")]
```
Stijnkool, line 26: This does not do anything. 😅 
If it's a type problem, we should explicitly fix this.

Stijnkool (changes requested):
Overall looks good.
Some small remarks about the loading spinner and some code here and there.

src/UI/Pages/CompileReports/FilterWidget/CompileReportsFilterWidget.tsx
andraskvr: It removes the space from `in progress` so the enum member is used

Stijnkool: Yea I know what it does 😅 . I mean, the only possible options are the ones from the enum... So why go from enum value to enum key to enum value... Thats what I mean with it doesn't do anything.
If this is in there because of a type issue, the type cast should be done explicitly.

Stijnkool (changes requested):
The table column of the status grows when the progress status is in there.
This makes the table jump around a bit. Maybe have the minimum width of the column be the size of the longest status label. This is not really required though.

And then the bit about the `CompileReportsFilterWidget`.

src/UI/Pages/CompileReports/FilterWidget/CompileReportsFilterWidget.tsx
andraskvr: The enum values are used as display names, and the keys are used internally, and they are not exactly the same, so it's an actual lookup and not only a cast. If you prefer we can use a string literal type instead of an enum and not have this distinction

Stijnkool: Yea thats probably the reason why I never create enums. String literal types are more forgiving and simpler to use and they are equally safe. But we have plenty of enums in the codebase. So no need to change this now.

Stijnkool (approved):

PR #3157, opened 2022-03-22
Navigate to new environment when creating one

### LINKED ISSUE #3137: Go to new environment when creating a new one
When a new environment is created, it should go directly to the environment instead of back to the overview

andraskvr:
# Description

closes #3137 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3137-navigate-to-new-environment.yml
+++ b/changelogs/unreleased/3137-navigate-to-new-environment.yml
@@ -1,0 +1,4 @@
+description: Navigate to new environment when creating one
+issue-nr: 3137
+change-type: minor
+destination-branches: [master, iso5]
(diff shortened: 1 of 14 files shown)


Stijnkool (commented):
Why do the triggers have to change to now return data?
Before they only returned the error because the response data was not being used.
I still do not see the response data being used...?
We only seem to be interested if the data is `Right` not the actual data, which is the same as the previous `Maybe.none`.

andraskvr:
> I still do not see the response data being used...? 
The id of the new environment (returned from the `PUT` call) is used to navigate to the next page. (Line 69 of `CreateEnvironmentForm`)

Stijnkool (approved):

PR #3231, opened 2022-03-31
Take the default value in account when using the server_compile setting

### LINKED ISSUE #3188: Within a new environment, the recompile button is disabled while the server_compile setting shows enabled
### Problem
Within a new environment, the recompile button is disabled while the server_compile setting shows enabled.

### API
A new environment does not have the `server_compile` setting set.
The backend response does not return this property and is therefor undefined.

### web-console
On the configuration tab of the settings page, the server_compile setting is shown as enabled.
This is the default value. The default value is used for undefined settings.

The web-console never uses any of the settings. These settings have no effect on the web-console.
It is just data shown to the user.
Except for the `server_compile` property. It is used to disable the Recompile button.

### Fix
Does the frontend need to use the default value for any settings it uses that are undefined?
Or does the backend need to return default values?

andraskvr:
# Description

Load the environment settings when an environment specific page is loaded, and update the `EnvironmentModifier` to take the default setting to account

closes #3188 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/UI/Root/Components/PageFrame/PageInitializer.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useContext } from "react";
+import { RemoteData } from "@/Core";
+import { DependencyContext } from "@/UI/Dependency";
+
+interface Props {
+  environment: string;
+}
+
+export const PageInitializer: React.FC<Props> = ({ children, environment }) => {
```
FloLey, line 9: Why is children needed? Is this to keep things generic?

FloLey: I feel like this does everything needed:

```
export const PageInitializer: React.FC<Props> = ({ environment }) => {
  const { queryResolver } = useContext(DependencyContext);

  queryResolver.useOneTime<"GetEnvironmentSettings">({
    kind: "GetEnvironmentSettings",
    environment: environment,
  });
  return null;
};

```

andraskvr: It was because at first I intended to use it as a wrapper, but then I realized that it wasn't needed, so you're right, it can be even simpler. :+1:

FloLey (approved):

PR #3262, opened 2022-04-04
Fix issue when the resource progress bar doesn't update automatically

### LINKED ISSUE #3242: Resource progress bar does not refresh anymore
The progress bar on the resource page does not refresh anymore. When you see the resource  states changing, the progress bar stays the same.

See recording of a full demo flow:

https://images.zenhubusercontent.com/209321501/12ed3e92-be6e-44e0-a570-3a3b3035afca/connect_demo_inmanta_com_novalocal_1__rocky_.mp4

andraskvr:
# Description

The data for the progress bar is cached so it provides a more consistent look when paging. 
In this case the table shows the loading view (as the state of the data is set to loading), but when the regular updates happen after 5 seconds, we don't show a loading view, only update the table with the new results. 
The `kind` of the data is not updated in this case, so the whole update of the progress bar was skipped. With this PR, the contents of the response are taken into account as well for the update.

closes #3242 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/Slices/Resource/UI/ResourcesPage/Page.tsx
+++ b/src/Slices/Resource/UI/ResourcesPage/Page.tsx
@@ -56,7 +56,7 @@ export const Page: React.FC = () => {
     if (RemoteData.isLoading(data)) return;
     setStaleData(data);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [data.kind, JSON.stringify(data)]);
+  }, [JSON.stringify(data)]);
 
   const updateFilter = (
     updater: (filter: Resource.Filter) => Resource.Filter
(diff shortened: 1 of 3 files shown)


src/Slices/Resource/UI/ResourcesPage/Page.tsx
```diff
@@ -56,7 +56,7 @@ export const Page: React.FC = () => {
     if (RemoteData.isLoading(data)) return;
     setStaleData(data);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [data.kind]);
+  }, [data.kind, JSON.stringify(data)]);
```
FloLey, line 59: isn't `data.kind` included in `JSON.stringify(data)`? so it could be removed? 
Also I don't know react enough to know if it is good practice to compare stringified data but I don't see why it would be bad :p

andraskvr: Good point about `kind`
About stringify, it's one of the options for this problem, the data we can have here is relatively small so I think it should be ok. (An alternative could be a custom hook with deep comparison https://github.com/kentcdodds/use-deep-compare-effect)

FloLey (approved):

andraskvr:
merged in separate PR

PR #3298, opened 2022-04-08
Show inter service relations on the catalog view

### LINKED ISSUE #3212: Add suport for interservice relations in service catalog attribute table
The service report in the catalog needs support for inter service relations. This is a new type of relation that points to other services.

When the server [reports this type](https://github.com/inmanta/inmanta-lsm/blob/ca6c75dcc43e5562fc0fe2cd93580a370d24bd24/src/inmanta_lsm/model.py#L324) These relations should be added to the table.

The relation should be presented as a normal attribute. However, the type should link to the correct catalog entry.

andraskvr:
# Description

closes #3212 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/UI/Components/TreeTable/RowReferenceContext.tsx
+++ b/src/UI/Components/TreeTable/RowReferenceContext.tsx
@@ -1,0 +1,12 @@
+import { createContext } from "react";
+
+interface TreeTableCell {
+  onClick(cellValue: string): void;
+}
+
+export const TreeTableCellContext = createContext<TreeTableCell>({
+  onClick: () => {
+    // Default to no-op
+    return;
+  },
+});
--- a/changelogs/unreleased/3212-catalog-inter-service-relations.yml
+++ b/changelogs/unreleased/3212-catalog-inter-service-relations.yml
@@ -1,0 +1,4 @@
+description: Show inter service relations on the catalog view
+issue-nr: 3212
+change-type: minor
+destination-branches: [master, iso5]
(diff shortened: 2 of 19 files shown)


src/UI/Components/TreeTable/RowReferenceContext.tsx
```diff
@@ -0,0 +1,12 @@
+import { createContext } from "react";
+
+interface TreeTableCell {
+  onClick(cellValue: string): void;
+}
+
+export const TreeTableCellContext = createContext<TreeTableCell>({
```
FloLey, line 7: I'm not sure I understand this part, why is it needed?

andraskvr: It's a way to pass this data down the component tree, without having to pass the props explicitly at every level in between (https://en.reactjs.org/docs/context.html).
In this specific case this is useful, because we have the refs to the service entity rows in the parent component (the `DataList`), and we need to do an action with it (expand the row and scroll to it) deep in the component tree, at the cell level in the `TreeTable`. (We'll be able to use it for the inventory view as well).
The naming is not the best, I'm open to suggestions :slightly_smiling_face:

FloLey: the name is fine by me, I don't think I will find a better one.

FloLey (approved):

PR #3476, opened 2022-05-16
Learning inmanta js

FloLey:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/Slices/ResourceV2/UI/resourceDetailsTab.tsx
+++ b/src/Slices/ResourceV2/UI/resourceDetailsTab.tsx
@@ -1,0 +1,57 @@
+import React, { useContext } from "react";
+import { Flex, FlexItem } from "@patternfly/react-core";
+import styled from "styled-components";
+import { Query, RemoteData } from "@/Core";
+import { useUrlStateWithString } from "@/Data";
+import { TabKey, Tabs } from "@/Slices/ResourceDetails/UI/Tabs";
+import { DependencyContext, words } from "@/UI";
+import {
+  Description,
+  PageContainer,
+  ResourceStatusLabel,
+} from "@/UI/Components";
+
+interface ResourceTabsProps {
+  resourceId: string;
+}
+
+export const ResourceDetailsTab: React.FunctionComponent<ResourceTabsProps> = ({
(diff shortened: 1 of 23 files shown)


FloLey:
@andraskvr I've added some test, it would clearly not be enough would this not be an exercise but I think it is enough for now. 
Do you agree or would you want me to add some more? Maybe you could do a little review of my code to?

src/Slices/ResourceV2/UI/resourceDetailsTab.tsx
```diff
@@ -0,0 +1,57 @@
...
+  resourceId: string;
+}
+
+export const ResourceDetailsTab: React.FunctionComponent<ResourceTabsProps> = ({
+  resourceId,
+}) => {
+  const { queryResolver } = useContext(DependencyContext);
+  const [activeTab, setActiveTab] = useUrlStateWithString<TabKey>({
+    default: TabKey.Attributes,
+    key: `tab`,
```
andraskvr, line 24: It's ok for now, but the key is not unique on the page, so if you change the active tab at one place you will change it in every row.

src/Slices/ResourceV2/UI/resourceDetailsTab.tsx
```diff
@@ -0,0 +1,57 @@
...
+  Description,
+  PageContainer,
+  ResourceStatusLabel,
+} from "@/UI/Components";
+
+interface ResourceTabsProps {
+  resourceId: string;
+}
+
+export const ResourceDetailsTab: React.FunctionComponent<ResourceTabsProps> = ({
```
andraskvr, line 18: There is also the `React.FC` shorthand instead of `React.FunctionComponent`, both are fine

andraskvr (commented):
Some suggestions for the future, but in general it's ok :+1:

src/Slices/ResourceV2/UI/resourceDetailsTab.tsx
FloLey: I see, this I need to remember to change in the future :p

PR #3550, opened 2022-05-30
issue/3519: Creating an environment fails when repository field is filled and cleared again

### LINKED ISSUE #3519: Creating an environment fails when repository field is filled and cleared again
When creating an environment, if a user writes something in the repository field, and then deletes it, the request fails with:
Screenshot from 2022-05-20 10-41-13

This can be quite confusing, since the field is empty, but the request is sent with the repository set to an empty string (instead of not setting it at all), causing the request to fail.

Creating the environment should still be possible after cancelled edits.

FloLey:
# Description

When creating an environment, if a user writes something in the repository field, and then deletes it, the request fails. This commit fixes this and allow to set fields back to empty.

closes #3519 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3519-fix-clear-field-env-creation.yml
+++ b/changelogs/unreleased/3519-fix-clear-field-env-creation.yml
@@ -1,0 +1,3 @@
+description: Fix env creation by allowing to clear optional fields
+change-type: patch
+destination-branches: [master, iso5]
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/3519-fix-clear-field-env-creation.yml
```diff
@@ -0,0 +1,3 @@
+description: Fix env creation by allowing to clear optional fields
+change-type: patch
+destination-branches: [master, iso5]
```
andraskvr, line 3: I think the issue also exists on iso4

andraskvr (changes requested):

andraskvr (approved):

PR #4055, opened 2022-09-16
replace deprecated class

matborowczyk:
# Description

* Short description here *

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/4002-replace-deprecated-class.yml
+++ b/changelogs/unreleased/4002-replace-deprecated-class.yml
@@ -1,0 +1,9 @@
+---
+description: replace KeyCloakInstance as it is depraceted
+issue-nr: 4002
+change-type: patch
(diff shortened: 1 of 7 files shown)


LauraStordeur (approved):
We can't completely test it, but we have no choice in updating it so... I think it's allright to go with it. It doesn't break anything existant, and we will dive deeper into the KeyCloak stuff when it works :D

bartv:
> We can't completely test it, but we have no choice in updating it so... I think it's allright to go with it. It doesn't break anything existant, and we will dive deeper into the KeyCloak stuff when it works :D

Fine for me, maybe wait until the release branch-off is done. This should be in the next days. After that there should be a ticket to track this. I did not have enough time yet to make a working Keycloak setup for testing.

PR #4163, opened 2022-10-06
add compile in progress indication

matborowczyk:
# Description

* Short description here *

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3950-compile-report-attribute.yml
+++ b/changelogs/unreleased/3950-compile-report-attribute.yml
@@ -1,5 +1,5 @@
 ---
-description: add indication for "Compile Reports" navigation link that inform about any ongoing compilations
+description: Add an indication in the navigation sidebar when there are ongoing compilations. The indication will be added to the "Compile Report" menu-item.
 change-type: patch
 destination-branches: 
  - master
--- a/changelogs/unreleased/3950-compile-report-attribute.yml
+++ b/changelogs/unreleased/3950-compile-report-attribute.yml
@@ -1,0 +1,8 @@
+---
+description: add indication for "Compile Reports" navigation link that inform about any ongoing compilations
+change-type: patch
+destination-branches: 
+ - master
+ - iso5
+sections: 
+  minor-improvement: "{{description}}"
--- a/src/Data/Managers/GetCompilationState/QueryManager.ts
+++ b/src/Data/Managers/GetCompilationState/QueryManager.ts
@@ -1,0 +1,67 @@
+import { useContext, useEffect, useState } from "react";
+import {
+  ApiHelper,
+  ContinuousQueryManager,
+  QueryManagerKind,
+  Query,
+  RemoteData,
+  Scheduler,
+} from "@/Core";
+import { DependencyContext } from "@/UI/Dependency";
+
+export class GetCompilationStateQueryManager
+  implements ContinuousQueryManager<"GetCompilationState">
+{
+  constructor(
+    private readonly apiHelper: ApiHelper,
+    private readonly scheduler: Scheduler
+  ) {}
+
+  useContinuous(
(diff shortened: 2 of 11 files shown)


LauraStordeur:
can you share a screenshot/gif  of what it looks like? :D

changelogs/unreleased/3950-compile-report-attribute.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: add indication for "Compile Reports" navigation link that inform about any ongoing compilations
```
LauraStordeur, line 2: Add an indication in the navigation sidebar when there are ongoing compilations. The indication will be added to the "Compile Report" menu-item.

src/Data/Managers/GetCompilationState/QueryManager.ts
```diff
@@ -0,0 +1,67 @@
...
+{
+  constructor(
+    private readonly apiHelper: ApiHelper,
+    private readonly scheduler: Scheduler
+  ) {}
+
+  useContinuous(
+    query: Query.SubQuery<"GetCompilationState">
+  ): [RemoteData.RemoteData<undefined, boolean>, () => void] {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
LauraStordeur, line 23: Why do you need to disable the linting rule specifically?

src/UI/Root/Components/Sidebar/Navigation/NavigationItem.tsx
```diff
@@ -26,8 +27,15 @@ export const NavigationItem: React.FC<Link> = ({
...
 }) => {
-  if (locked) return <LockedItem label={label} key={id} />;
-  if (external) return <ExternalItem label={label} url={url} key={id} />;
+  if (locked) {
+    return <LockedItem label={label} key={id} />;
+  }
+  if (external) {
+    return <ExternalItem label={label} url={url} key={id} />;
+  }
+  if (id === "CompileReports") {
```
LauraStordeur, line 36: The day you change the id, it won't work anymore. I think it would have been better to add an optional property to the Link Interface, just like it's done for the other elements. I think that would be more dynamic and flexible.

Also, maybe something to discuss later, but I don't know if you know the concept of 'magic' strings. This is what I consider a magic string, and usually, that's something you would rather store in a const.

src/UI/Root/Components/Sidebar/Navigation/NavigationItem.tsx
```diff
@@ -68,6 +76,42 @@ const ExternalItem: React.FC<Label & Url> = ({ label, url }) => (
   </NavItem>
 );
 
+const CompileReportItem: React.FC<Label & Url> = ({ label, url }) => {
+  const { queryResolver } = useContext(DependencyContext);
+  const [data, retry] = queryResolver.useContinuous<"GetCompilationState">({
+    kind: "GetCompilationState",
+  });
+
+  useEffect(() => {
```
LauraStordeur, line 85: Am I understanding it right, that you are adding and removing the listener on every API/query call?

src/Data/Managers/GetCompilationState/QueryManager.ts
LauraStordeur: But, there is I suppose a good reason why that shouldn't be used, right? I remember it being a hard rule for a reason, I checked an I saw it was used in multiple places, but I have a hunch that the guys before us did bypass the rule just because they were lazy to understand it.

src/UI/Root/Components/Sidebar/Navigation/NavigationItem.tsx
matborowczyk: that return is useEffect version of componentsDidUnmount

LauraStordeur (approved):

PR #4180, opened 2022-10-11
replace expand/collapse icons with dropdown

matborowczyk:
# Description

* Short description here *

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/UI/Components/TreeTable/Helpers/TreeTableHelper.ts
```diff
@@ -66,16 +66,18 @@ export abstract class BaseTreeTableHelper<A extends AttributeTree>
...
 
     const nodes = this.attributeHelper.getMultiAttributeNodes(this.attributes);
 
-    return Object.entries(nodes)
-      .map(([key, node]) => treeRowCreator.create(key, node))
-      .sort((a, b) => a.id.localeCompare(b.id));
+    return {
+      rows: Object.entries(nodes)
+        .map(([key, node]) => treeRowCreator.create(key, node))
+        .sort((a, b) => a.id.localeCompare(b.id)),
```
LauraStordeur, line 77: if by any hasard "a" doesn't have an id, your code will break. 
I know it's handy to use such one liners, but in this case, I'd recommend to block it with an if case and return statement.

src/UI/Components/TreeTable/Helpers/TreeTableHelper.ts
```diff
@@ -44,13 +44,13 @@ export abstract class BaseTreeTableHelper<A extends AttributeTree>
...
+  ): { rows: TreeRow[]; openAll: () => void; closeAll: () => void } {
     const createOnToggle = (key: string) => () =>
       setState(this.expansionManager.toggle(expansionState, key));
-    const createCloseAll = (key: string) => () =>
-      setState(this.expansionManager.toggleAll(expansionState, key, false));
-    const createOpenAll = (key: string) => () =>
-      setState(this.expansionManager.toggleAll(expansionState, key, true));
+    const createCloseAll = () => () =>
+      setState(this.expansionManager.toggleAll(expansionState, false));
+    const createOpenAll = () => () =>
```
LauraStordeur, line 52: One question for them all, why chaining two anonymous functions?

src/UI/Components/TreeTable/Helpers/TreeTableHelper.ts
```diff
@@ -44,13 +44,13 @@ export abstract class BaseTreeTableHelper<A extends AttributeTree>
   createRows(
     expansionState: ExpansionState,
     setState: (state: ExpansionState) => void
-  ): TreeRow[] {
+  ): { rows: TreeRow[]; openAll: () => void; closeAll: () => void } {
     const createOnToggle = (key: string) => () =>
```
LauraStordeur, line 48: Does this not work?
`const createOnToggle = (key: string) => setState(this.expansionManager.toggle(expansionState, key));`

matborowczyk: As fas as I understand this piece of code, not really, as it is used in creator when that key is used and returns a function that doesn't require any key/path to toggle visibility, I mean it could work, but then we would just move that anonymous function somewhere else

LauraStordeur: Hmm, it's a pattern they are reusing a lot... returning anonymous functions. Fine, I guess it can stay like this since it would require too much logic or adjustment to avoid having this kind of situations.

LauraStordeur (approved):

PR #4242, opened 2022-10-20
add preselected only option for select inputs in form

### LINKED ISSUE #4127: When only one option available this one should be preselected
For required values that are proposed in Select Components, it would be nicer to have the only available value preselected. This would avoid the user to have to click twice for something that he can't make a choice on.

matborowczyk:
# Description

from now one value in select inputs will be preselected if only one option is available - Create Instance Form

closes #4127 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx
+++ b/src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useCallback } from "react";
 import {
   Button,
   FormFieldGroupExpandable,
@@ -28,8 +28,7 @@ interface Props {
   path: string | null;
 }
 
-type Update = (value: unknown) => void;
-type GetUpdate = (path: string, multi?: boolean) => Update;
+type GetUpdate = (path: string, value: unknown, multi?: boolean) => void;
 
 const makePath = (path: string | null, next: string): string =>
   path === null ? next : `${path}.${next}`;
@@ -40,6 +39,12 @@ export const FieldInput: React.FC<Props> = ({
   getUpdate,
   path,
 }) => {
+  const getEnumUpdate = useCallback(
+    (value) => {
+      getUpdate(makePath(path, field.name), value);
+    },
+    [getUpdate, path, field.name]
+  );
   switch (field.kind) {
     case "Boolean":
       return (
@@ -49,7 +54,7 @@ export const FieldInput: React.FC<Props> = ({
           isOptional={field.isOptional}
           isChecked={get(formState, makePath(path, field.name)) as boolean}
           handleInputChange={(value) =>
-            getUpdate(makePath(path, field.name))(toOptionalBoolean(value))
+            getUpdate(makePath(path, field.name), toOptionalBoolean(value))
           }
           description={field.description}
           key={field.name}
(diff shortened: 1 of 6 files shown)


src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
```diff
@@ -29,9 +29,8 @@ export const ServiceInstanceForm: React.FC<Props> = ({
       : createFormState(fields)
   );
 
-  const getUpdate =
-    (path: string, multi = false) =>
-    (value: unknown): void => {
+  const getUpdate = useCallback(
```
LauraStordeur, line 32: Soo... based on what I read on React about this.. is that you technically need to pass the arguments in that array at the end of your hook. 
https://reactjs.org/docs/hooks-reference.html#usecallback

so in our case it would be something like this: 
```

const getUpdate = useCallback(
> () => {
>> //logic
>}, [path, value, multi]
);
```
I'm not so familiar either with this, so I can be wrong and the documentation is kinda abstract.

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.test.tsx
```diff
@@ -77,6 +77,65 @@ test("GIVEN ServiceInstanceForm WHEN passed an EnumField THEN shows that field",
...
+      onCancel={jest.fn()}
+      onSubmit={jest.fn()}
+    />
+  );
+  const component = screen.getByRole("generic", {
+    name: `EnumFieldInput-${Test.Field.enumFieldSingleOption.name}`,
+  });
+  expect(component).toBeVisible();
+
+  const preselectedOption = screen.getByText("local");
```
LauraStordeur, line 124: I'd get the value from your EnumObject :) "local" is hardcoded and if you change the test data you're stuck with changing it :)

src/UI/Components/ServiceInstanceForm/Components/SelectFormInput.tsx
```diff
@@ -29,7 +29,12 @@ export const SelectFormInput: React.FC<Props> = ({
   const selectOptions = Object.keys(options)
     .sort()
     .map((key) => <SelectOption key={key} value={key} />);
-
+  useEffect(() => {
```
LauraStordeur, line 32: can you add a newline between 31-32 ? makes it more readable ^^

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
matborowczyk: I think that wouldn't work as value is coming from one of different type of children, and we would have to add some state management which is kind of pointless in my opinion. I am curious why I had to use useCallback twice to make error with re-renders disappear

src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx
```diff
@@ -40,6 +39,12 @@ export const FieldInput: React.FC<Props> = ({
   getUpdate,
   path,
 }) => {
+  const getEnumUpdate = useCallback(
```
LauraStordeur, line 42: Also, maybe make sure to add some docstrings to explain why you had to use memoization :D

LauraStordeur:
Can you check the checkboxes in the description aswell (good habit to not forget stuff) as add the ticket to the description? 
A gif of the behaviour would also be nice to see in action. I can also checkout the branch if you don't have time for it.

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
LauraStordeur: Then I fear this is not the right solution somehow 🤔 But then again, I know React can be a pain about rerenders... So, I'm not confident on this matter

matborowczyk:
> Can you check the checkboxes in the description aswell (good habit to not forget stuff) as add the ticket to the description? A gif of the behaviour would also be nice to see in action. I can also checkout the branch if you don't have time for it.

I don't have in current db dump select value with only one option, so gif won't cover task :(

LauraStordeur (approved):

PR #4293, opened 2022-11-07
Issue/4115 env config rwd

### LINKED ISSUE #4115: Environment Config buttons falling off screen on smaller widths.

matborowczyk:
# Description

* Short description here *

closes #4115 

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/Slices/Settings/UI/Tabs/Configuration/Container.tsx
+++ b/src/Slices/Settings/UI/Tabs/Configuration/Container.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { Alert, AlertActionCloseButton } from "@patternfly/react-core";
 import { TableComposable, Tbody } from "@patternfly/react-table";
 import styled from "styled-components";
@@ -18,6 +18,20 @@ export const Container: React.FC<Props> = ({
   onErrorClose,
   className,
 }) => {
+  const [showUpdateBanner, setShowUpdateBanner] = useState(false);
+
+  useEffect(() => {
+    const updateSuccessBanner = () => {
+      setShowUpdateBanner(true);
+      setTimeout(() => {
+        setShowUpdateBanner(false);
+      }, 2000);
+    };
+    document.addEventListener("Settings update", updateSuccessBanner);
+    return () => {
+      document.removeEventListener("Settings update", updateSuccessBanner);
+    };
+  }, [setShowUpdateBanner]);
   return (
     <Wrapper className={className}>
       {errorMessage && (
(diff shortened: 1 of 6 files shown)


src/Slices/Settings/UI/Tabs/Configuration/Container.tsx
```diff
@@ -18,6 +18,20 @@ export const Container: React.FC<Props> = ({
...
+  const [showUpdateBanner, setShowUpdateBanner] = useState(false);
+
+  useEffect(() => {
+    const updateSuccessBanner = () => {
+      setShowUpdateBanner(true);
+      setTimeout(() => {
+        setShowUpdateBanner(false);
+      }, 2000);
+    };
+    document.addEventListener("Settings update", updateSuccessBanner);
```
LauraStordeur, line 30: I think we should have a standard for this or agree on a casing or something. Events in html are always lowercased. To keep it readable, we can dash them. so this one would be 
`settings-update`

If you agree of course, we could make a decision log for it in the notes (that's the place I've added decisions/logs about choices so far, until we have a dedicated place for that)

src/Slices/Settings/UI/Tabs/Configuration/Container.tsx
```diff
@@ -29,6 +43,19 @@ export const Container: React.FC<Props> = ({
           isInline
         />
       )}
+      {showUpdateBanner && (
+        <StyledAlert
+          variant="success"
+          title={"Setting Changed"}
```
LauraStordeur, line 49: extract to the words list, I think that's also a standard we need to start doing. Atm it's not aligned and we have a mix. But everything that's printed to the screen, should be in that `words` list. That way, we have a central place for all hardcoded strings.

src/Slices/Settings/UI/Tabs/Configuration/Container.tsx
matborowczyk: Fine for me, I'll change and add info about this to notes

src/UI/Components/DictEditor/DictEditor.tsx
```diff
@@ -128,7 +128,9 @@ const Row: React.FC<RowProps> = ({
 const StyledTextInput = styled(TextInput)`
   &&& {
```
LauraStordeur, line 129: what is &&& ? Never seen that before. I know how to use a single ampersand... but three in a row?

matborowczyk: It was also something new for me, generally its solution from styled-components lib to group all styles that need !important value and give them required priority

LauraStordeur: !important... whoo getting chills from it 😅

LauraStordeur (approved):

src/UI/Components/DictEditor/DictEditor.tsx
matborowczyk: I know, but patternfly doesn't like to get some of its styles overwritten without it :D

Issue #4125, opened 2022-09-28
Inform user when he leaves unsaved form


LauraStordeur:
When you are filling a form it is currently possible to leave the form. The user should be informed that he is losing unsaved changes. This can be done by having a "isDirty" state in the form and showing an alert/modal that would prevent the user from leaving the page unknowingly.

bartv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/frontend-600a82c1f666040010200918/issues/inmanta/web-console/4125?planning-poker) with Zenhub @LauraStordeur @matborowczyk

PR #4311, opened 2022-11-10
Create hooks that check status of edit/create instance form

matborowczyk:
# Description

* Short description here *

# Self Check: #4125

- [ ] Changelog entry

# Reviewer Checklist:

https://user-images.githubusercontent.com/113331659/201123710-898f1f8c-30c4-4d40-89d2-e3fbe2043ec9.mov

### DIFF
--- a/src/Slices/CreateInstance/UI/CreateInstance.tsx
+++ b/src/Slices/CreateInstance/UI/CreateInstance.tsx
@@ -36,9 +36,14 @@ export const CreateInstance: React.FC<Props> = ({ serviceEntity }) => {
     service_entity: serviceEntity.name,
   });
 
-  const onSubmit = async (attributes: InstanceAttributeModel) => {
+  const onSubmit = async (
+    attributes: InstanceAttributeModel,
+    setIsDirty: (value: boolean) => void
+  ) => {
+    setIsDirty(false);
     const result = await trigger(fields, attributes);
     if (result.kind === "Left") {
+      setIsDirty(true);
       setErrorMessage(result.value);
     } else {
       handleRedirect();
--- a/src/Slices/EditInstance/UI/EditForm.tsx
+++ b/src/Slices/EditInstance/UI/EditForm.tsx
@@ -46,9 +46,14 @@ export const EditForm: React.FC<Props> = ({ serviceEntity, instance }) => {
     id: instance.id,
     version: instance.version,
   });
-  const onSubmit = async (attributes: InstanceAttributeModel) => {
+  const onSubmit = async (
+    attributes: InstanceAttributeModel,
+    setIsDirty: (value: boolean) => void
+  ) => {
+    setIsDirty(false);
     const result = await trigger(fields, currentAttributes, attributes);
     if (Maybe.isSome(result)) {
+      setIsDirty(true);
       setErrorMessage(result.value);
     } else {
       handleRedirect();
--- a/src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.test.tsx
+++ b/src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.test.tsx
@@ -1,17 +1,47 @@
 import React from "react";
+import { Route, Routes, BrowserRouter as Router } from "react-router-dom";
 import { render, screen, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
+import {
+  BooleanField,
+  DictListField,
+  EnumField,
+  NestedField,
+  TextField,
+} from "@/Core";
 import * as Test from "@/Test";
 import { ServiceInstanceForm } from "./ServiceInstanceForm";
 
-test("GIVEN ServiceInstanceForm WHEN passed a TextField THEN shows that field", async () => {
-  render(
-    <ServiceInstanceForm
-      fields={[Test.Field.text]}
-      onCancel={jest.fn()}
-      onSubmit={jest.fn()}
-    />
+const setup = (
+  fields: (
+    | TextField
+    | BooleanField
+    | NestedField
+    | DictListField
+    | EnumField
+  )[],
+  func: undefined | jest.Mock = undefined
+) => {
+  return (
+    <Router>
+      <Routes>
+        <Route
+          path="/"
+          element={
+            <ServiceInstanceForm
+              fields={fields}
+              onCancel={jest.fn()}
+              onSubmit={func ? func : jest.fn()}
+            />
+          }
+        />
+      </Routes>
+    </Router>
   );
+};
+test("GIVEN ServiceInstanceForm WHEN passed a TextField THEN shows that field", async () => {
+  render(setup([Test.Field.text]));
+
   expect(
     screen.getByRole("generic", {
       name: `TextFieldInput-${Test.Field.text.name}`,
@@ -26,13 +56,7 @@ test("GIVEN ServiceInstanceForm WHEN passed a TextField THEN shows that field", 
 });
 
 test("GIVEN ServiceInstanceForm WHEN passed a BooleanField THEN shows that field", async () => {
-  render(
-    <ServiceInstanceForm
-      fields={[Test.Field.bool]}
-      onCancel={jest.fn()}
-      onSubmit={jest.fn()}
-    />
-  );
+  render(setup([Test.Field.bool]));
   expect(
     screen.getByRole("generic", {
       name: `BooleanFieldInput-${Test.Field.bool.name}`,
@@ -47,13 +71,8 @@ test("GIVEN ServiceInstanceForm WHEN passed a BooleanField THEN shows that field
 });
 
 test("GIVEN ServiceInstanceForm WHEN passed an EnumField THEN shows that field", async () => {
-  render(
-    <ServiceInstanceForm
-      fields={[Test.Field.enumField]}
-      onCancel={jest.fn()}
-      onSubmit={jest.fn()}
-    />
-  );
+  render(setup([Test.Field.enumField]));
+
   expect(
     screen.getByRole("generic", {
       name: `EnumFieldInput-${Test.Field.enumField.name}`,
@@ -78,13 +97,8 @@ test("GIVEN ServiceInstanceForm WHEN passed an EnumField THEN shows that field",
 });
 
 test("GIVEN ServiceInstanceForm WHEN passed an EnumField with more than one value THEN shows that field with default prompt", async () => {
-  render(
-    <ServiceInstanceForm
-      fields={[Test.Field.enumFieldTwoOptions]}
-      onCancel={jest.fn()}
-      onSubmit={jest.fn()}
-    />
-  );
+  render(setup([Test.Field.enumFieldTwoOptions]));
+
   const component = screen.getByRole("generic", {
     name: `EnumFieldInput-${Test.Field.enumFieldTwoOptions.name}`,
   });
@@ -109,13 +123,7 @@ test("GIVEN ServiceInstanceForm WHEN passed an EnumField with more than one valu
 });
 
 test("GIVEN ServiceInstanceForm WHEN passed an EnumField with only one value THEN shows that field with preselected option", async () => {
-  render(
-    <ServiceInstanceForm
-      fields={[Test.Field.enumFieldSingleOption]}
-      onCancel={jest.fn()}
-      onSubmit={jest.fn()}
-    />
-  );
+  render(setup([Test.Field.enumFieldSingleOption]));
   const component = screen.getByRole("generic", {
     name: `EnumFieldInput-${Test.Field.enumFieldSingleOption.name}`,
   });
@@ -139,13 +147,7 @@ test("GIVEN ServiceInstanceForm WHEN passed an EnumField with only one value THE
 });
 
 test("GIVEN ServiceInstanceForm and a NestedField WHEN clicking the toggle THEN the nested FlatField is shown", async () => {
-  render(
-    <ServiceInstanceForm
-      fields={[Test.Field.nested([Test.Field.text])]}
-      onCancel={jest.fn()}
-      onSubmit={jest.fn()}
-    />
-  );
+  render(setup([Test.Field.nested([Test.Field.text])]));
 
   const group = screen.getByRole("group", {
     name: "nested_field",
@@ -165,13 +167,7 @@ test("GIVEN ServiceInstanceForm and a NestedField WHEN clicking the toggle THEN 
 });
 
 test("GIVEN ServiceInstanceForm and a DictListField WHEN clicking all toggles open THEN the nested FlatField is shown", async () => {
-  render(
-    <ServiceInstanceForm
-      fields={[Test.Field.dictList([Test.Field.text])]}
-      onCancel={jest.fn()}
-      onSubmit={jest.fn()}
-    />
-  );
+  render(setup([Test.Field.dictList([Test.Field.text])]));
 
   const group = screen.getByRole("group", {
     name: "dict_list_field",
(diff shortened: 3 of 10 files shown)


LauraStordeur:
If you click on cancel, you shouldn't have the alert imo. It's willingly canceling the form, so the user is aware he's abandoning the current form.

src/UI/Utils/useIsDirty.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+    },
+    {
+      property: "baseValue",
+    },
+    {
+      property: "baseValue",
+    }
+  );
+  render(component);
+  expect(screen.getByText("true")).toBeVisible();
```
LauraStordeur, line 88: please use the words method

src/UI/Utils/useIsDirty.tsx
```diff
@@ -0,0 +1,99 @@
...
+ */
+export default function useIsDirty(
+  object,
+  baseToCompare: InstanceAttributeModel | undefined = undefined
+) {
+  const [isDirty, setIsDirty] = useState(false);
+  const overrideState = (value: boolean) => {
+    setIsDirty(value);
+  };
+  useEffect(() => {
```
LauraStordeur, line 20: I'm gonna ask more explanation on the reasoning of this complicated sequence, please

src/UI/Utils/usePrompt.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  });
+  expect(prompt).toBeCalledTimes(0);
+  expect(window.location.pathname).toMatch("/page");
+});
+
+test("GIVEN usePrompt WHEN hook's parameter is equal true and user confirm alert window THEN page is changed", async () => {
+  const prompt = jest.spyOn(window, "confirm").mockImplementation(() => true);
+  render(setup());
+
+  //cleanup doesn't reset page url so I had to manually go back to "/"
```
LauraStordeur, line 78: you can do that in the afterEach maybe?

src/UI/Utils/usePrompt.tsx
```diff
@@ -0,0 +1,57 @@
+import { useCallback, useContext, useEffect } from "react";
+import { UNSAFE_NavigationContext as NavigationContext } from "react-router-dom";
```
LauraStordeur, line 2: UNSAFE ??

src/UI/words.tsx
```diff
@@ -540,6 +540,8 @@ const dict = {
   "notification.drawer.clearAll": "Clear all",
   "notification.drawer.details": "Details",
   "notification.drawer.clear": "Clear",
+  "notification.instanceForm.prompt":
+    "There are some modified properites in Form. Do you want exit?",
```
LauraStordeur, line 544: Are you sure you want to leave this page? You have unsaved changes.

src/UI/Utils/useIsDirty.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+  render(component);
+  expect(screen.queryByText("true")).not.toBeInTheDocument();
+
+  const button = screen.getByText("Change Data");
+
+  await act(async () => {
+    await userEvent.click(button);
+  });
+
+  expect(screen.getByText("true")).toBeVisible();
```
LauraStordeur, line 145: Aside from the last test case, it might be nice to have a loop, cause all other tests are a repetition, the input of the setup method is the only thing that changes

LauraStordeur (commented):
Okay... so I've slightly been shocked at the complexity required to achieve this. 

In the explanation on your hook, you say it is continuous, does it mean like the other continuous hook, every 5 seconds?

I had in mind this wouldn't be that extreme to implement. Every input field has a change event, that is being triggered that can inform you that the state of one input field has been changed somewhere. So that would set a variable that is available in your State object. 

If that one is set to true, when the user tries to leave the page, you would get a prompt. When the user confirms, the state variable is being reset. When the user submit a form the variable would also be reset. 

I would have imagined the only complex thing would be to catch the route change attempt... I didn't really expect the full blown compare of the data in your form.

LauraStordeur:
Also, ticket is not linked in zenhub ;)

src/UI/Utils/useIsDirty.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+};
+
+afterEach(() => {
+  cleanup;
+});
+test("GIVEN useIsDirty WHEN hook's parameter that is being checked isn't changed from it's default THEN isDirty value is falsy and span isn't visible in any moment of lifecycle", async () => {
+  Field.nestedEditable, ServiceInstance.nestedEditable.candidate_attributes;
+  const { component } = setup(
+    {
+      property: "",
```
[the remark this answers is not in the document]
matborowczyk: I extracted objects to variables, to avoid repetition, I don't know why I didn't do it in the first place, but if it comes what they mean it is told when you hover over setup()

src/UI/Utils/useIsDirty.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  await act(async () => {
+    await userEvent.click(button);
+  });
+
+  expect(screen.getByText("true")).toBeVisible();
+});
+
+test("GIVEN useIsDirty WHEN hook's parameter that is being checked is changed from empty value to the same that hook's baseObject has THEN isDirty value is falsy and span isn't visible", async () => {
+  Field.nestedEditable, ServiceInstance.nestedEditable.candidate_attributes;
```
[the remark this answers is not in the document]
matborowczyk: No purpose at all, I have no idea how I didn't saw it

src/UI/Utils/useIsDirty.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+      property: "baseValue",
+    },
+    {
+      property: "baseValue",
+    }
+  );
+  render(component);
+  expect(screen.getByText("true")).toBeVisible();
+
+  const button = screen.getByText("Change Data");
```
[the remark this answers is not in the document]
matborowczyk: It's not, it's just made for the sake of test case, I changed it to variable tho

src/UI/Utils/usePrompt.test.tsx
matborowczyk: It doesn't work

src/UI/Utils/usePrompt.tsx
matborowczyk: I know, but I didn't find other way to halt and resume react-router navigation in v6, and there is no other solution online that doesn't use this

src/UI/Utils/useIsDirty.test.tsx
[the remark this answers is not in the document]
LauraStordeur: I can't hover when reviewing :) And when functions get complicated, hovering isn't always clear as to what is being passed :)

matborowczyk:
> Okay... so I've slightly been shocked at the complexity required to achieve this.
> 
> In the explanation on your hook, you say it is continuous, does it mean like the other continuous hook, every 5 seconds?
> 
> I had in mind this wouldn't be that extreme to implement. Every input field has a change event, that is being triggered that can inform you that the state of one input field has been changed somewhere. So that would set a variable that is available in your State object.
> 
> If that one is set to true, when the user tries to leave the page, you would get a prompt. When the user confirms, the state variable is being reset. When the user submit a form the variable would also be reset.
> 
> I would have imagined the only complex thing would be to catch the route change attempt... I didn't really expect the full blown compare of the data in your form.

Continuous was meant as on every change of an object form.
To be fair I somewhat disliked approach to adding some kind of checker to update function itself as I couldn't find elegant solution to handle different input types at once as we get few ways how properties can be updated especially when editing, but If you find my implementation overkill I can try to change it

LauraStordeur:
@matborowczyk If I'm not mistaken, we have the 'getUpdate' callback that we can use to detect changes. The FieldInput component is encapsulating all possible inputs/select/toggles we are using in forms. 
You could set the isDirty state there :) 
image

src/UI/Utils/usePrompt.tsx
LauraStordeur: normally window.alert does that natively but you're using a different implementation

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.test.tsx
```diff
@@ -234,10 +223,13 @@ test("GIVEN ServiceInstanceForm WHEN clicking the submit button THEN callback is
...
-    [dictListField.name]: [{ [dictListField.fields[0].name]: "test text 3" }],
-  });
+  expect(submitCb).toHaveBeenCalledWith(
+    {
+      [Test.Field.text.name]: "test text",
+      [Test.Field.bool.name]: true,
+      [nestedField.name]: { [nestedField.fields[0].name]: "test text 2" },
+      [dictListField.name]: [{ [dictListField.fields[0].name]: "test text 3" }],
+    },
+    expect.any(Function)
```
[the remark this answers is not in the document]
matborowczyk: That whole was extended due to the fact that I needed additional parameter which is anonymous function declared inside of a ServiceInstanceForm and I don't think there is a way to extract it and comparing it to another function won't work.
I mean I found [this](https://medium.com/fredwong-it/react-jest-testing-mock-function-inside-functional-component-1e0d0bfb431a) but isn't it just tweaking component to fit tests?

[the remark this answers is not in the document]
LauraStordeur: I didn't mean to extract it; but expect.any() seems to be used as the second part of your assertion. In your example, I believe that no matter what happens in the test, the assertion will always pass.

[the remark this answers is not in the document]
LauraStordeur: Mainly because a react Component will already be considered as a Function.

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
```diff
@@ -28,7 +33,9 @@ export const ServiceInstanceForm: React.FC<Props> = ({
       ? createEditFormState(fields, originalAttributes)
       : createFormState(fields)
   );
-
+  const [dirtyInputs, setDirtyInputs] = useState<string[]>([]);
```
LauraStordeur, line 36: Why the array? 
The getUpdate is being called on any change in any input field, so you can by default assume the isDirty state can be set to True.

matborowczyk: Yeah, but we should be able to tell if user clear all data manually, Uncommon but not impossible

matborowczyk: if we would base dirty state then I don't see easy way to tell when to change it back to false

LauraStordeur: yeah, but performance wise this is like a huge difference... imagine having an instance with a really complex form with a lot of fields and nested fields... You'll be comparing continuously the difference on each change... Each letter that you write is triggering the getUpdate.

If you keep it simple and set isDirty on the first change you catch, you only need to set it once, and can make a check If IsDirty just continue executing rest of the code, otherwise set it to true.

The case where the user will be clearing the whole form manually is 1/XXXXXX They'd rather click the cancel button in that case.

src/UI/Routing/CustomRouter.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import { HistoryRouterProps, Router } from "react-router-dom";
+
+function CustomRouter({ basename, children, history }: HistoryRouterProps) {
```
LauraStordeur, line 4: Add comment as to why we needed it ;)

src/Slices/CreateInstance/UI/CreateInstance.tsx
```diff
@@ -36,9 +36,15 @@ export const CreateInstance: React.FC<Props> = ({ serviceEntity }) => {
     service_entity: serviceEntity.name,
   });
 
-  const onSubmit = async (attributes: InstanceAttributeModel) => {
+  const onSubmit = async (
+    attributes: InstanceAttributeModel,
+    dirtyInputs: string[],
+    setIsDirty: (values: string[]) => void
+  ) => {
+    setIsDirty([]);
```
LauraStordeur, line 44: what if the submit fails? You should clear the isDirty only if the submit is okay

src/Slices/EditInstance/UI/EditForm.tsx
```diff
@@ -46,9 +46,15 @@ export const EditForm: React.FC<Props> = ({ serviceEntity, instance }) => {
     id: instance.id,
     version: instance.version,
   });
-  const onSubmit = async (attributes: InstanceAttributeModel) => {
+  const onSubmit = async (
+    attributes: InstanceAttributeModel,
+    dirtyInputs: string[],
+    setIsDirty: (values: string[]) => void
+  ) => {
+    setIsDirty([]);
```
LauraStordeur, line 54: same remark here

package.json
```diff
@@ -131,7 +131,7 @@
     "copy-to-clipboard": "^3.3.2",
     "easy-peasy": "^5.1.0",
     "file-saver": "^2.0.5",
-    "history": "5.3.0",
+    "history": "5",
```
LauraStordeur, line 134: I checked the lock file, it still installs the 5.3.0

LauraStordeur:
There are other forms too, a more generic or low-level solution might be worth the effort. Currently the solution only covers the instance form, but we have the one for the environment, we have the settings too, the callback form, and I might be missing some others.

src/Slices/CreateInstance/UI/CreateInstance.tsx
matborowczyk: if it fails we put these values again,, changing it to just clearing would cause the same issues as in canceling, as setState is async so redirect would take place first and that would trigger prompt

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
LauraStordeur: It will, cause currently your logic is being run on each input field that is being touched. You set a state which causes rerenders on each input change. 

While on the other hand if you only have 1 boolean attribute that is being set once, and causes 1 rerender.

src/Slices/CreateInstance/UI/CreateInstance.tsx
LauraStordeur: So placing it in the else section before the redirect with an await doesn't work?

src/UI/Utils/usePrompt.test.tsx
```diff
@@ -0,0 +1,106 @@
+import React, { useEffect, useState } from "react";
+import { Link, Route, Routes } from "react-router-dom";
+import { cleanup, render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { act } from "react-dom/test-utils";
+import CustomRouter from "../Routing/CustomRouter";
+import history from "../Routing/history";
+import { usePrompt } from "./usePrompt";
+jest.mock("react-router-dom", () => ({
```
LauraStordeur, line 9: can you add new lines between methods? :D

src/Slices/CreateInstance/UI/CreateInstance.tsx
matborowczyk: No as it doesn't return Promise, but change to the state doesn't happen immediate, or can happen later, which in that case happens and redirect trigger prompt

LauraStordeur: Okay, can you add it as a comment in the code ;)

package.json
matborowczyk: As it's most actual one, "5" doesn't mean to stick to 5.0.0, it's just most current one in range of major version 5

LauraStordeur (approved):

PR #4357, opened 2022-11-21
Issue/4332 service details bugs

### LINKED ISSUE #4332: [BUG]Service details callback page issues
# Task checklist
- [x] Solve the UUID being kept in the form
- [x] Change delete icon to 'Delete' text
- [x] Move collapsible arrow to the front of the row
- [x] Change the Minimal Log Level numerical code to text
- [x] Add a tooltip over the UUID to display full value instead of 4 first chars.
- [x] Truncate UUID after 8 char instead of 4 in table

# Issue 1

When you add a callback to your service details, and submit, the form isn't cleared correctly and the last used uuid remains in the id field. 

# Issue 2
The icon from the delete button is not centered => According to PF guidelines, icons only is not a good practice. And because we are already using text in the add-button, we decided to change it to a text button too. 

# Issue 3
The arrow to expand the row needs to be moved in front of it, to be aligned with all other tables used so far. 

# Issue 4
After saving a callback, the Minimal Log Level is being displayed as a numerical code. We need to change that to the text label matching the code. This is already available in the API response.

# Issue 5
The UUID displayed in the ID collumn is being truncated, 4 characters is not enough. It would be better to show 8, add a tooltip with the full value.

matborowczyk:
# Description

* Short description here *

closes #4332 

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

Screenshot 2022-11-22 at 11 59 01

### DIFF
--- a/changelogs/unreleased/4332-service-details-bugs.yml
+++ b/changelogs/unreleased/4332-service-details-bugs.yml
@@ -1,5 +1,5 @@
 change-type: patch
-description: 'Fix service callbacks issues in display, in readabilty and form behaviour after adding component'
+description: 'Fix service callbacks component issues in display, in readabilty and form behaviour after adding component'
 issue-nr: 4332
 destination-branches:
 - master
--- a/changelogs/unreleased/4332-service-details-bugs.yml
+++ b/changelogs/unreleased/4332-service-details-bugs.yml
@@ -1,5 +1,5 @@
 change-type: patch
-description: 'Fix service callbacks component issues in display, in readabilty and form behaviour after adding component'
+description: 'Fix service callbacks issues in display, in readabilty and form behaviour after adding callback'
 issue-nr: 4332
 destination-branches:
 - master
--- a/changelogs/unreleased/4332-service-details-bugs.yml
+++ b/changelogs/unreleased/4332-service-details-bugs.yml
@@ -1,6 +1,8 @@
 change-type: patch
-description: 'Fix service details issues'
+description: 'Fix service callbacks issues in display, in readabilty and form behaviour after adding component'
 issue-nr: 4332
 destination-branches:
 - master
 - iso5
+sections: 
+  minor-improvement: "{{description}}"
--- a/changelogs/unreleased/4332-service-details-bugs.yml
+++ b/changelogs/unreleased/4332-service-details-bugs.yml
@@ -1,0 +1,6 @@
+change-type: patch
+description: 'Fix service details issues'
+issue-nr: 4332
+destination-branches:
+- master
+- iso5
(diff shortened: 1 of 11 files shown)


changelogs/unreleased/4332-service-details-bugs.yml
```diff
@@ -0,0 +1,6 @@
+change-type: patch
+description: 'Fix service details issues'
```
LauraStordeur, line 2: I would add a section for this, since there are multiple bugs being fixed

LauraStordeur:
Can you add a screenshot/recording of the fixes :D

changelogs/unreleased/4332-service-details-bugs.yml
```diff
@@ -0,0 +1,8 @@
+change-type: patch
+description: 'Fix service callbacks issues in display, in readabilty and form behaviour after adding component'
```
LauraStordeur, line 2: after adding a Callback
(Components are a concept that we understand, but for end users reading the changelogs, it doesn't make a lot of sense)

src/Core/Domain/Uuid.ts
```diff
@@ -8,4 +8,4 @@ export const getUuidFromRaw = (raw: string): Uuid => ({
   short: raw.substring(0, 4),
 });
 
-export const getShortUuidFromRaw = (raw: string): string => raw.substring(0, 4);
+export const getShortUuidFromRaw = (raw: string): string => raw.substring(0, 8);
```
LauraStordeur, line 11: add docstrings ;)

matborowczyk: I will, but I thought this is self-explanatory :D

src/Slices/ServiceDetails/UI/Tabs/Callbacks/Row.tsx
```diff
@@ -29,25 +29,35 @@ export const Row: React.FC<Props> = ({
...
+              value={callback.callback_id}
+              tooltipContent={words("catalog.callbacks.uuid.copy")}
+            >
+              {shortUuid}
+            </TextWithCopy>
+          </Tooltip>
+        </Td>
+        <Td>{callback.minimal_log_level_text}</Td>
+        <Td>
+          {callback.event_types &&
```
LauraStordeur, line 55: Are you ellypsing ? Or what's the idea of this?

matborowczyk: That was previous approach to truncate chosen event_type, it was in ListWithTeaser, but that component was redundant after changing position of the toggle

LauraStordeur: I don't know for you, but I don't like the way it's displayed. Since they are shown in the expansion, what about displaying in the table : "X Event Types" and the tooltip/expansion will show the full selection? 
@bartv what is your thought on this? The ellypsing of the list means bitterly nothing to the user, the first name isn't even fully readable and the user is forced anyway to click open the row.

LauraStordeur (approved):

Issue #4305, opened 2022-11-09
Research E2E automation


LauraStordeur:
Investigate how we can automate e2e tests with a docker container.

PR #4374, opened 2022-11-25
4305 investigate e2e automation

LauraStordeur:
# Description

This PR closes the first steps in the automation for the e2e tests. Two scripts have been added separately, and are being sequentially called in the jenkinsfile to run the tests. 

Both commands can be ran independently from the pipeline. 

## setup-server.sh
This will pull in the local-setup repository from Gitlab. Builds the docker containers for the orchestrator and install the lsm-test modules. It will always copy the latest dist available on your current branch (The jenkins pipeline is building a new version in the e2e stage). But it means you can have any version locally you wish to run the tests. 

## clear-server.sh
This script is meant to be run AFTER you have setup a server. It will destroy the docker containers and remove the temp folder containing the local-setup repo.

closes *#4305*

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


shell-scripts/setup-server.sh
```diff
@@ -0,0 +1,47 @@
...
+
+mkdir temp
+cd temp
+
+git clone $local_setup_repo || exit 1
+
+cd local-setup
+
+# This can be adjusted to pull in iso4 by replacing the command with 'yarn run pull:4dev'
+yarn run pull:5dev
```
LauraStordeur, line 28: could actually maybe become a variable in the env.sh I guess?

matborowczyk: Right now there is no need for that, I guess, as there is only one script that use it, and in the future we can modify it.
But Also I don't mind if you would prefer to extract it :D

LauraStordeur: Great ! I'll let it be for now, and keep it on a side note in my head for later :D thanks for your opinion :D

PR #4518, opened 2023-01-06
Fix configuration update issue 

### LINKED ISSUE #4481: Settings can only be updated one by one
When a user adjust the value of more than one setting on the settings tab and clicks the save button behind one of the updated settings, the values of all other settings reset to their original value. As such, all settings have to be updates one by one at the moment.

[Screencast from 12-23-2022 08:34:09 AM.webm](https://user-images.githubusercontent.com/2684622/209293425-5db54f35-f466-4db4-af78-7abe8118abfe.webm)

matborowczyk:
# Description

* Short description here *

closes #4481 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/Slices/Settings/UI/Tabs/Configuration/Provider.tsx
+++ b/src/Slices/Settings/UI/Tabs/Configuration/Provider.tsx
@@ -1,4 +1,5 @@
-import React, { useContext, useEffect, useState } from "react";
+import React, { useContext, useEffect, useReducer, useState } from "react";
+import _ from "lodash";
 import { EnvironmentSettings } from "@/Core";
 import { DependencyContext } from "@/UI/Dependency";
 import { Container } from "./Container";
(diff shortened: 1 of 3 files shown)


matborowczyk:
https://user-images.githubusercontent.com/113331659/210999034-5ccd0309-12f3-4254-b7c4-1c8f320a6ca3.mov

matborowczyk:
There is still one small issue, reseting inputs still appear in nested objects, as I couldn't find solution that doesn't include rearranging whole flow, and I want to discuss it on Monday on sprint planning as Laura wanted to discuss UX of that Tab anyway

src/Slices/Settings/UI/Tabs/Configuration/Provider.tsx
```diff
@@ -8,10 +9,42 @@ interface Props {
   settings: EnvironmentSettings.EnvironmentSettings;
 }
 
+function reducer(
```
LauraStordeur, line 12: Can you add a comment as to Why the reducer was needed ? For later reference, cause I already forgot parts of the explanation from last week x)

matborowczyk: Done :)

LauraStordeur (approved):

PR #4622, opened 2023-02-09
refresh

### LINKED ISSUE #4608: Page does not refresh went deleting a catalog entry
When you delete a catalog entry and click yes in the modal, nothing happens until the next refresh.

matborowczyk:
# Description

* Short description here *

closes #4608 

# Self Check:

- [x] Changelog entry


changelogs/unreleased/4613-dependabot.yml
```diff
@@ -0,0 +1,6 @@
+change-type: patch
```
LauraStordeur, line 1: why is this one in there?

LauraStordeur: the merge tool will block the merge if there are multiple yml files :)

matborowczyk: I made pull request from master, and something had to happen in between, as it's treated as something new, I made pull on master and there is that file on master

LauraStordeur: yes, but if you pulled in master, this file should already be present and thus, not be in the diff 🤔

matborowczyk: I am aware, but this file exist on master already https://github.com/inmanta/web-console/blob/master/changelogs/unreleased/4613-dependabot.yml

LauraStordeur (approved):

PR #4774, opened 2023-04-06
fix issue that missing setting and it's definition crash the app

### LINKED ISSUE #4772: fix-env-settings-error

matborowczyk:
# Description

useSetting implementation is based with approach that envSetting definition will be fallback source of information which wasn't the case for the expertMode issue reported by Sander

closes #4772 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/UI/Dependency/EnvironmentModifier.test.tsx
+++ b/src/UI/Dependency/EnvironmentModifier.test.tsx
@@ -3,8 +3,10 @@ import { render, screen } from "@testing-library/react";
 import { StoreProvider } from "easy-peasy";
 import { act } from "react-dom/test-utils";
 import { EnvironmentModifier, RemoteData } from "@/Core";
+import { DefinitionMap } from "@/Core/Domain/EnvironmentSettings";
 import { getStoreInstance } from "@/Data";
 import { EnvironmentDetails, EnvironmentSettings } from "@/Test";
+import ErrorBoundary from "../Utils/ErrorBoundary";
 import { EnvironmentModifierImpl } from "./EnvironmentModifier";
 
 const DummyComponent: React.FC<{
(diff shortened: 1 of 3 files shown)


src/UI/Dependency/EnvironmentModifier.test.tsx
```diff
@@ -79,3 +85,21 @@ test("Given the environmentModifier When the server compile setting is requested
...
+    EnvironmentSettings.definition
+  );
+  // No setting is specified, and the dafault is missing
+  store.dispatch.environment.setEnvironmentDetailsById({
+    id: environmentId,
+    value: RemoteData.success({ ...EnvironmentDetails.a, settings: {} }),
+  });
+
+  render(component);
+  //expect to see div as the value would be false insteand of error boundary page
```
LauraStordeur, line 101: instead*
and see a div is really abstract, is it an empty div?

The test description also doesn't match what you are testing. You're not asserting a return value being false, but testing if a server-compile-disabled element is visible in the DOM. 

Also, being completely accurate, if you write in the comment that one should be displayed instead of another, you should also test that the other element is not present. Theoretically, both could be present in the DOM and this test could result in a fake passing test.

src/UI/Dependency/EnvironmentModifier.test.tsx
```diff
@@ -17,28 +19,32 @@ const DummyComponent: React.FC<{
...
+      definition,
     }),
   });
   const environmentModifier = EnvironmentModifierImpl();
   environmentModifier.setEnvironment(environmentId);
   const component = (
-    <StoreProvider store={store}>
-      <DummyComponent environmentModifier={environmentModifier} />
-    </StoreProvider>
+    <ErrorBoundary>
```
LauraStordeur, line 35: You added the errorBoundary element, but you're not making any assertion against it

src/UI/Dependency/EnvironmentModifier.test.tsx
matborowczyk: the setup is made in a way that setting value is represented by the div, either with the name "disabled" or "enabled" in it and It couldn't be possible to have both in them but I'll add assesrtion as it doesn't cost a thing. I'll add also assertion for consol.error to have errrorBoundary component covered

LauraStordeur: Yes, but setup can change :) 
About the div, it's more the comment not being descriptive enough, a div could be anything :)

src/UI/Dependency/EnvironmentModifier.test.tsx
```diff
@@ -79,3 +82,27 @@ test("Given the environmentModifier When the server compile setting is requested
     await screen.findByRole("generic", { name: "server-compile-enabled" })
   ).toBeVisible();
 });
+
+test("Given the environmentModifier When the missing setting is requested Then render component as the value would be false without throwing an error", async () => {
```
LauraStordeur, line 86: The description of your tests still says that you are testing the value being returned would be false. You have no assertions explicitly testing that.

matborowczyk: by the setup if there is server_compile false then div with label "server-compile-disabled" is visible so checking the visibility of that div is sufficient, right?

matborowczyk: The title of the test says that when setting is missing then render component as the value would be fale

LauraStordeur (approved):

PR #4787, opened 2023-04-13
fix notification issue

### LINKED ISSUE #4767: [BUG] Some notifications don't clear on hitting clear button.
Some notifications still go unread instead of being cleared. This is an edge case I suppose, as this doesn't happen all the time.

https://user-images.githubusercontent.com/44098050/229540658-bdeda61d-df92-4fe7-902f-a7c1820b6691.mp4

matborowczyk:
# Description

* Short description here *

closes #4767 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/Slices/Notification/UI/Drawer/Item.tsx
+++ b/src/Slices/Notification/UI/Drawer/Item.tsx
@@ -84,7 +84,7 @@ const ActionList: React.FC<Props> = ({ notification, onUpdate }) => {
         <DropdownItem
           key="cleared"
           component="button"
-          onClick={() => onUpdate({ read: true, cleared: true })}
+          onClick={() => onUpdate({ cleared: true })}
         >
           {words("notification.drawer.clear")}
         </DropdownItem>,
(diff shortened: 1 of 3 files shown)


src/Slices/Notification/UI/Drawer/Item.tsx
```diff
@@ -84,7 +84,7 @@ const ActionList: React.FC<Props> = ({ notification, onUpdate }) => {
         <DropdownItem
           key="cleared"
           component="button"
-          onClick={() => onUpdate({ read: true, cleared: true })}
+          onClick={() => onUpdate({ cleared: true })}
```
LauraStordeur, line 87: Question, there's a reason why they are set on read, isn't this supposed to trigger an API call ? When you go to the notification page, you can filter the notifications based on their (un)read status. I rather think the issue lies in the updateMethod instead of the passed arguments.

matborowczyk: I checked once again and notification Item has a onclick and event was propagated to it which resulted of doubling request where latter one has body to only mark notification as "read". So my attempt resolved issue but in sub-optimal way, I'll change that :)

LauraStordeur (approved):

PR #4817, opened 2023-04-28
Fix for optional Enums in InstanceForm

### LINKED ISSUE #4790: Value select does not work anymore in instance forms
When a field is an enum we show a select with the options. In the latest release 6.2.0 this is broken. It shows it for a required field but the optional field at the bottom does not have it

Image

In a version from 20230303 you can see that it still worked:

Image

matborowczyk:
# Description

* Short description here *

closes #4790 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/4790-select-input-fix.yml
+++ b/changelogs/unreleased/4790-select-input-fix.yml
@@ -3,5 +3,4 @@ description: 'Fix issue that cause optional enums to display wrong input in inst
 destination-branches:
 - master
 - iso6
-- iso5
 sections: {}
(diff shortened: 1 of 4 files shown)


changelogs/unreleased/4790-select-input-fix.yml
```diff
@@ -0,0 +1,7 @@
+change-type: patch
+description: 'Fix issue that cause optional enums to display wrong input in instance form'
+destination-branches:
+- master
+- iso6
+- iso5
```
LauraStordeur, line 6: You can remove the iso-5 target ;)

LauraStordeur (approved):

PR #4838, opened 2023-05-08
Display embedded entity - smart composer

### LINKED ISSUE #4756: Add support to the composer for displaying an embedded entity
Extend #4756 to also display embedded entities as an instance with a relation to the parent entity.

matborowczyk:
# Description

* Short description here *

closes #4756 

# Self Check:

- [x] Changelog entry


src/UI/Components/Diagram/actions.ts
```diff
@@ -40,45 +48,328 @@ export function showLinkTools(linkView: dia.LinkView) {
...
+
+  //auto-layout provided by JointJS
+  layout.DirectedGraph.layout(graph, {
+    dagre: dagre,
+    graphlib: graphlib,
+    nodeSep: 80,
+    edgeSep: 80,
+    rankDir: "TB",
+  });
+  return instanceAsTable.getBBox();
```
LauraStordeur, line 247: new line before return statement

src/UI/Components/Diagram/helpers.ts
```diff
@@ -0,0 +1,38 @@
+import { ServiceInstanceModel, ServiceModel } from "@/Core";
+
+export const extractRelationsIds = (
```
LauraStordeur, line 3: Same observation as for the other method regarding the different attribute state types, would be nice to later on to remove the repetition.

src/UI/Components/Diagram/actions.ts
```diff
@@ -40,45 +48,328 @@ export function showLinkTools(linkView: dia.LinkView) {
...
+  service: ServiceModel
+): g.Rect {
+  const flatAttributes = service.attributes.map((attribute) => attribute.name);
+
+  const instanceAsTable = new ServiceEntityBlock().setName(
+    serviceInstance.service_entity
+  );
+
+  //check if for any presentable attributes, if there is a set, then append them to  JointJS shape and try to display and connect embedded entities
+  if (serviceInstance.active_attributes !== null) {
```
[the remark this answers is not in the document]
matborowczyk: That's valid question that I missed, I will ask bart, but probably we would like to have only active attributes, with no mix

src/Core/Domain/AttributeValidation.ts
```diff
@@ -36,6 +36,11 @@ interface IntValidation {
 }
 
 interface IpValidation {
-  validation_type: "ipaddress.IPv4Address" | "ipaddress.IPv4Address?";
-  validation_parameters: null;
+  validation_type:
```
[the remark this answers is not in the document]
matborowczyk: I'll add that as soon as I will get the list from the core team

package.json
```diff
@@ -136,6 +136,7 @@
     "@patternfly/react-tokens": "^4.94.6",
     "@react-keycloak/web": "^3.4.0",
     "copy-to-clipboard": "^3.3.3",
+    "dagre": "^0.8.5",
```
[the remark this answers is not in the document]
matborowczyk: Yes, it was needed to enable that auto-layout thing, which it was said that was out of the box but i ran into issues that it wasn't working and I had to install it anyway.

matborowczyk:
I removed logic for active and candidates attributes, and I will introduce toggle with toolbar as that it's needed for that anyway. :)

src/Core/Domain/ServiceModel.ts
```diff
@@ -7,13 +7,13 @@ export type AttributeModel = AttributeValidation & {
...
   modifier: string;
-  default_value: string | null;
+  default_value: string | string[] | null;
   default_value_set: boolean;
 };
 
 export interface StateModel {
   deleted: boolean;
-  label?: "info" | "success" | "warning" | "danger";
+  label?: "info" | "success" | "warning" | "danger" | null;
```
LauraStordeur, line 16: is there a reasoning behind adding the null possibility ?

src/UI/Components/Diagram/helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+  service: ServiceModel,
+  instance: ServiceInstanceModel
+): string[] => {
+  const relationKeys = service.inter_service_relations?.map(
+    (relation) => relation.name
+  );
+  if (!relationKeys) {
+    return [];
+  }
+  if (instance.candidate_attributes !== null) {
```
LauraStordeur, line 13: I'd have the "state" as an argument for the toggling, I think it might be easier.

something like this :
`instance[attribute_state]`

src/Core/Domain/ServiceModel.ts
matborowczyk: at the current mock taken from real world example few states were missing label property so I had to add it to cover whole range of possible "states"

src/UI/Components/Diagram/helpers.ts
matborowczyk: I thought that would require additional cast of type on it, but just checked and that would work. I will change it in then on current branch that I'm working on if that's okay with you

src/Core/Domain/ServiceModel.ts
LauraStordeur: I'm a bit confused, don't we need to add the missing states then, instead of covering it with null?

matborowczyk: If there is a possibility that backend returned label as null then I think it should be covered by corresponding interface. At least that's how I understood that case, especially that there is "no label" possibility in the service Inventory/Details

LauraStordeur (approved):

PR #4848, opened 2023-05-16
update Edit Instance Form to fit api v2 flow

### LINKED ISSUE #4766: Refactor Form data manipulation
> Blocker for Composer forms, needs to be discussed before starting on forms in the new feature !

**The way the data is handled in the form currently is like so:** 
- It gets the data from API
- It formats the data, storing into a configuration object the fields that will be displayed to the view. It discards any field that is hidden to the user. **This only causes issues in the edit flow.**
- When the form is being submitted, the output is the updated config object. 
- A merge is being executed with the original data, when it is an edit action.
- The data is sent to the BE.

**The issue with this flow is described as follow:** 
The merge action on the FE is an ordinary merge, it will fuse two object/arrays together. Changed values will be updated, but deleted values will not be seen as deleted, the old value will remain. Which means the user isn't able to delete nested values for example. 

The way the form discards the original object, makes it a lot of logic to define which value was actually deleted by the user or not. 

I don't know what is the best, to invest in a refactoring of the form, which would keep awareness of hidden values, or look into a way to make the form output the original format with changed values. 

Example : 
**Original data**
```
{
    "name": "embedded-entity",
    "service_id": "0001",
    "vlan_assigment_r1": {
        "address": "1.2.3.5/32",
        "vlan_id": 1,
        "router_ip": "1.2.3.2",
        "interface_name": "eth0"
    },
    "vlan_assigment_r2": {
        "address": "1.2.3.9/32",
        "vlan_id": 2,
        "router_ip": "1.2.3.6",
        "interface_name": "eth0"
    },
    "should_deploy_fail": false
}
```
**Form output with a changed value**
```
{
    "should_deploy_fail": false,
    "vlan_assigment_r1": {
        "router_ip": "1.2.3.1",
        "interface_name": "eth0",
        "address": "1.2.3.5/32",
        "vlan_id": 1
    },
    "vlan_assigment_r2": {
        "router_ip": "1.2.3.6",
        "interface_name": "eth0",
        "address": "1.2.3.9/32",
        "vlan_id": 2
    }
}
```

In the scenario where one of the embedded entities can be deleted, the object would result in something like this: 
```
{
    "should_deploy_fail": false,
    "vlan_assigment_r1": {
        "router_ip": "1.2.3.1",
        "interface_name": "eth0",
        "address": "1.2.3.5/32",
        "vlan_id": 1
    }
}
```

When comparing the original data with this, we can't tell if the attributes  `name` and `service_id` were deleted or if they weren't needed to build the form.

## Proposed solution

Keep the initial data, but only show the key-attribute fields as disabled, the other readonly fields can be hidden.

matborowczyk:
# Description

I changed at the end approach due to the attributes that aren't covered by FieldCreator, so logic practically the same but formState for v2 is just current state of the attributes, and there was no need for what I first tried to implement.

this test attributes were changed to reflect how the attributes really looks on our side, and that change didn't have any impact on other parts of the unit tests

closes #4766 

# Self Check:

- [x] Changelog entry


src/Slices/EditInstance/Data/CommandManager.test.ts
```diff
@@ -42,38 +42,36 @@ test("GIVEN getBodyV1 WHEN changing optional attributes THEN generates correct b
...
-//     operation: "replace",
-//     target: ".",
-//     value: {
-//       attr2: "",
-//       attr3: null,
-//       attr4: 42,
-//     },
-//   });
-// });
+  //to simulate actual behavior after refactor attr1 was added as from now on all of the fields are passed to getBodyV2 function in the flow.
```
LauraStordeur, line 57: I think, if I remember well, the reason for one attribute missing, is that the property itself can actually be removed. And that's different from setting it to undefined/null.

matborowczyk: okay, make sense, I'll adjust

matborowczyk: I had to leave attr1 as field for that attr is required, but I deleted other, attr2.

LauraStordeur (approved):

PR #4913, opened 2023-06-21
4910/add support textarea

### LINKED ISSUE #4910: Add support for Textarea in form
The form currently doesn't support Textarea. In some cases, users need to type in very long texts into the input fields, it was requested to make the input as a textarea in theses cases. 

We agreed on a workaround until a better solution has been implemented on the BE. 

If the constr is 255+ long, the input field should be adjusted to a textarea on the FE. An example will be added to the local-setup.

LauraStordeur:
# Description

Added a workaround based on a constraint to change the input field into a textarea. 

closes #4910 

image

### DIFF
--- a/src/UI/Components/ServiceInstanceForm/Helpers/FieldCreator.ts
+++ b/src/UI/Components/ServiceInstanceForm/Helpers/FieldCreator.ts
@@ -234,7 +234,8 @@ export class FieldCreator {
           (attribute.type === "string" || attribute.type === "string?") &&
           (attribute.validation_type === "pydantic.constr" ||
             attribute.validation_type === "pydantic.constr?") &&
-          attribute.validation_parameters.max_length === 10000
+          attribute.validation_parameters.max_length &&
+          attribute.validation_parameters.max_length > 255
         ) {
           return {
             kind: "Textarea",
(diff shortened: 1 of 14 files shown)


src/UI/Components/ServiceInstanceForm/Helpers/FieldCreator.ts
```diff
@@ -229,6 +229,25 @@ export class FieldCreator {
           };
         }
 
+        // WORKAROUND TO ADD SUPPORT FOR TEXTAREA
+        if (
+          (attribute.type === "string" || attribute.type === "string?") &&
+          (attribute.validation_type === "pydantic.constr" ||
+            attribute.validation_type === "pydantic.constr?") &&
+          attribute.validation_parameters.max_length === 10000
```
matborowczyk, line 237: max_length will be always 10000?

matborowczyk (approved):

src/UI/Components/ServiceInstanceForm/Helpers/FieldCreator.ts
LauraStordeur: it's the workaround we agreed on, it's until there is a better implementation on the BE for these :D

matborowczyk: Oh, that was behind workaround, now I understand :D

bartv: is it 10000 or is it longer than 255?

LauraStordeur: I made the adjustment !

Issue #4858, opened 2023-05-22
[Composer] Create halo functionality


matborowczyk:
The task is to append Halo onto the shapes with buttons with as close design as is possible.
![Image](https://github.com/inmanta/web-console/assets/113331659/a4126737-6129-41ce-9fba-c1dd52c4de12)

Optimally the task should be done before forms are fully implemented as edit form button will be here

Issue #4749, opened 2023-03-30
[Composer] adjust router (links between blocks) to fit design


matborowczyk:
currently connection between shapes/entities are based on the router implemented in the demo provided by jointjs team.

PR #4995, opened 2023-07-27
Toolbar multiple tasks

### LINKED ISSUE #4992: fix error when displaying null/undefined values in the composer

matborowczyk:
# Description

Batch of tasks:
1. added zoom-in/zoom-out buttons in the corner, they are a little bit more out of the corner, as for some reason canvas bar was overlapping them for split second and spamming the button could result with moving the view. 

2. Created Halo, which currently can delete shape/entity with links connected to it, and from now halo hold button to create links with drag&drop. 
I went through docs to check if it's possible to do links by click on source, then on target, but I didn't  find anything ready to implement, If we want to stick to it as it was in the design I'll sent a message to jointJS team if they have some demo for that.

3.Additionally fixed the issue with crashing the view when value is undefined/null, as they don't have to have tooltip,(short enough) so i took it out of the scope that was handling that.

closes #4992 #4993 #4749 #4858 

# Self Check:

- [x] Changelog entry

https://github.com/inmanta/web-console/assets/113331659/f3630f94-3904-430c-853c-3a2565735bab


src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -103,3 +128,43 @@ const Container = styled.div`
   margin: 0 40px;
   height: 100%;
 `;
+
+const ZoomWrapper = styled.div`
+  z-index: 10;
```
LauraStordeur, line 133: From personal experience, I know this can become easily a mess when you have to add more elements in between or over or under that index. 
At my previous projects, we would have a global set of variables for the z-indexes. Which allowed to have an overview of the used index and easily adjust them if needed. I don't know if you like that approach or not.

matborowczyk: forgot to delete that, that was my attempt to "hide" scrollbar, but it didn't help and I ended up moving these buttons a little bit :D But I agree that z-indexes can get messy very fast

LauraStordeur (approved):

PR #5085, opened 2023-08-25
Issue/5083 edit composer

### LINKED ISSUE #5083: Add EditForm to Composer

matborowczyk:
# Description

This is the base logic for data manipulation in created shapes, which will also result in changes to "composer state" that will be translated accordingly for the endpoint (in the next iteration)

closes #5083 

# Self Check:

- [x] Changelog entry
- [] Sufficient test cases (reproduces the bug/tests the requested feature)


src/UI/Components/Diagram/actions.ts
```diff
@@ -177,10 +189,13 @@ export function appendEmbeddedEntity(
     return createdInstances;
   } else {
     const instanceAsTable = new ServiceEntityBlock()
-      .setTabColor("#0066CC")
+      .setTabColor("#009596")
```
LauraStordeur, line 192: it would be a nice idea to extract these into variables, also easier to read what color it is then :) You can make an object named `Colors` maybe?

LauraStordeur (approved):

src/UI/Components/Diagram/actions.ts
matborowczyk: Good point, Done :)

matborowczyk:
https://github.com/inmanta/web-console/assets/113331659/44c1aee7-a7da-45b2-b4d5-a9d97b8db0f4

PR #5112, opened 2023-09-07
Issue/5092 composer api

### LINKED ISSUE #5092: composer api call

matborowczyk:
# Description
At the moment after request is being sent objects are being "stuck" with "in_progress" status, due to to lack of functionality on the core side

To review AFTER 5084 PR merge, as its build on top of it 

Improved displaying of shapes, and connection between them, 
completed state tracking and transformation shapes into proper api call

closes #5092 

# Self Check:

- [x] Changelog entry


src/UI/Components/Diagram/helpers.ts
```diff
@@ -220,3 +221,109 @@ const doesElementsIsEmbeddedWithExhaustedConnections = (
...
+  const matchingInstances = instances.filter(
+    (checkedInstance) => checkedInstance.embeddedTo === instance.instance_id,
+  );
+
+  const notMatchingInstances = instances.filter(
+    (checkedInstance) => checkedInstance.embeddedTo !== instance.instance_id,
+  );
+
+  //iterate through matching (embedded)instances and group them according to property type to be able to put them in the Array if needed at once
+  const groupedEmbedded: { [key: string]: InstanceForApi[] } =
```
LauraStordeur, line 249: I know key is a generic often used, but in our case, I think it would be nicer to use what will actually be used as key. If I'm not mistaken, this would be the instanceId or serviceId, it will make more sense when we come back later on that.

src/UI/Components/Diagram/helpers.ts
```diff
@@ -220,3 +221,109 @@ const doesElementsIsEmbeddedWithExhaustedConnections = (
...
+        const updated: { [key: string]: unknown }[] = [];
+        instancesToEmbed.forEach((instance) => {
+          const updatedInstance = shapesDataTransform(
+            notMatchingInstances,
+            instance,
+            true,
+          );
+
+          areEmbeddedEdited =
+            areEmbeddedEdited !== true &&
```
LauraStordeur, line 268: Any reason why you're not using !areEmbeddedEdited and !instance.action ?

src/UI/Components/Diagram/helpers.ts
```diff
@@ -220,3 +221,109 @@ const doesElementsIsEmbeddedWithExhaustedConnections = (
...
+            instance.action === null &&
+            (updatedInstance.action !== null ||
```
LauraStordeur, line 270: you got three actions, create, update, delete, so if it's not null, it's one of the three. So your first check cancels out the second. to be of type "delete" it's by definition, not null. Since you used an or statement, it will never bother doing the "delete" check if the first statement is valid. In the case the first statement is invalid, it means that the action is null, and thus can never be "delete".

src/UI/Components/Diagram/helpers.ts
```diff
@@ -220,3 +221,109 @@ const doesElementsIsEmbeddedWithExhaustedConnections = (
...
+              updatedInstance.action === "delete");
+
+          if (updatedInstance.action !== "delete") {
+            updated.push(updatedInstance.value as { [key: string]: unknown });
```
LauraStordeur, line 274: I've seen this type quite often, it might be relevant to extract it to a more explicit typing :)

src/UI/Components/Diagram/helpers.ts
```diff
@@ -220,3 +221,109 @@ const doesElementsIsEmbeddedWithExhaustedConnections = (
...
+      } else {
+        const data = shapesDataTransform(
+          notMatchingInstances,
+          instancesToEmbed[0],
+          true,
+        );
+
+        areEmbeddedEdited =
+          instance.action === null &&
+          (data.action !== null || data.action === "delete");
```
LauraStordeur, line 288: same remark here, if the first statement is valid, it won't bother checking for the "delete" part.

src/UI/Components/Diagram/helpers.ts
```diff
@@ -220,3 +221,109 @@ const doesElementsIsEmbeddedWithExhaustedConnections = (
...
+  if (instance.relatedTo) {
+    instance.relatedTo.forEach((attrName, id) => {
+      if (instance.value) {
+        instance.value[attrName] = id;
+      }
+    });
+  }
+
+  //if any of its embedded instances were edited, and its action is indicating no changes to main attributes, change it to "update"
+  if (areEmbeddedEdited && instance.action === null) {
```
LauraStordeur, line 307: any reason why you duplicate the instance.action === null check, while you already check it when you assign areEmbeddedEdited?

src/UI/Components/Diagram/shapes.ts
```diff
@@ -244,24 +244,25 @@ export class ServiceEntityBlock extends shapes.standard.HeaderedRecord {
...
     } else {
       const relationMap = new Map();
-      relationMap.set(id, instanceName);
-      this.set("relatedTo", relationMap.set(id, instanceName));
+      this.set("relatedTo", relationMap.set(id, relationName));
     }
   }
 
-  removeRelation(id: string): void {
+  removeRelation(id: string): boolean {
```
LauraStordeur, line 258: The name of the method is misleading, you want to know if it was sucessful or not, I get it, but the way you're using the method in the halo file, in an ifcase, is confusing.

src/UI/Components/Diagram/actions.ts
```diff
@@ -61,25 +70,39 @@ export function showLinkTools(graph: dia.Graph, linkView: dia.LinkView) {
...
+            didTargetCellChanged = true;
           }
 
           // resolve any possible relation connections between cells
           const sourceRelations = sourceCell.getRelations();
           const targetRelations = targetCell.getRelations();
 
           if (sourceRelations && sourceRelations.has(target.id as string)) {
-            sourceCell.removeRelation(target.id as string);
+            didSourceCellChanged =
```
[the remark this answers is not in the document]
matborowczyk: when deleting a Link, there is no indication on which side of the cell the information about possible relation is being stored, so if there are (inter-service)relations, and the relation with target is one of them, then we are removing that, and setting the outcome of the removal(which should be true, as there was that relation stored and deleting position from Map will return `true`).

Now as I wrote that I realised that `didSourceCellChanged ||` is redundant as there is no possibility for one link to be responsible for embedded connection and relation, so I'll change that

src/UI/Components/Diagram/actions.ts
```diff
@@ -224,6 +287,24 @@ export function appendEmbeddedEntity(
         );
         connectEntities(graph, instanceAsTable, appendedEntity);
       });
+
+      embeddedEntity.inter_service_relations?.map((relation) => {
+        const relationId = entityAttributes[relation.name] as
```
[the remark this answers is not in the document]
matborowczyk: entityAttributes  are type of `InstanceAttributeModel` which is `Record<string, unknown>`, but I'll create type for this purpose here

src/UI/Components/Diagram/actions.ts
```diff
@@ -387,12 +494,27 @@ function handleAttributes(
...
       serviceModel.name,
+      instanceToConnectRelation,
     );
     appendedEntities.map((entity) => {
       handleInfoIcon(entity, presentedAttr);
     });
     connectEntities(graph, instanceAsTable, appendedEntities);
   });
+  serviceModel.inter_service_relations?.forEach((relation) => {
+    const relationId = attributesValues[relation.name];
```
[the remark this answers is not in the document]
matborowczyk: attributesValues are `Record<string, unknown>`, I moved casting to `const relationId = attributesValues[relation.name] as string;`

src/UI/Components/Diagram/shapes.ts
matborowczyk: Hmm, what's your proposition? I based myself on delete() that is in the Map, but added this whole if()as some explanation, because delete returns the same information, but it could be more confusing without these extra few lines

src/UI/Components/Diagram/helpers.ts
matborowczyk: I was chiseling in that function quite a bit and I probably forgot to use it :D

src/UI/Components/Diagram/helpers.ts
matborowczyk: I had to get lost in those conditional statements, thank for pointing it out!

src/UI/Components/Diagram/shapes.ts
LauraStordeur: Why not work with a promises ? It would make more sense to me when using the removeRelation method. 

``` javascript
      if (relations) {
        didElementChange =
          didElementChange === true ||
          elementAsService.removeRelation(cellView.model.id as string).then(handle response)"
      }
```

src/UI/Components/Diagram/helpers.ts
matborowczyk: too many conditionals at once i guess :D

src/UI/Components/Diagram/helpers.ts
matborowczyk: Good point, I propose to use `InstanceAttributeModel` which is essentially it(i mentioned it above) and is used with instance forms

src/UI/Components/Diagram/shapes.ts
matborowczyk: But why make it async?

LauraStordeur: The way you wrote the method is similar to any sort of promise. Execute a task and then return if it was successful or not. That's a promise to me.

matborowczyk: I kind of agree, but in my opinion emphasis for promises should be on the fact of awaiting for that feedback. I wouldn't use actual Promise for that as that action in our case is completely synchronous.

src/UI/Components/Diagram/actions.ts
```diff
@@ -13,9 +14,14 @@ import { Colors, EntityConnection, ServiceEntityBlock } from "./shapes";
  * https://resources.jointjs.com/docs/jointjs/v3.6/joint.html#linkTools
  *
  * @param {dia.LinkView} linkView  - The view for the joint.dia.Link model.
+ * @param {(): void} linkView  - The view for the joint.dia.Link model.
```
LauraStordeur, line 17: I think you forgot to adjust the copy pasted line ;) 
you got two arguments missing in the docs

src/UI/Components/Diagram/actions.ts
```diff
@@ -135,30 +154,76 @@ export function appendInstance(
...
-    (relatedInstance) => {
-      return appendInstance(paper, graph, relatedInstance, services, false, {
-        id: serviceWithReferences.instance.data.id,
-        name: serviceWithReferences.instance.data.service_entity,
-      });
-    },
-  );
-  connectEntities(graph, instanceAsTable, appendedInstances);
+  serviceWithReferences.relatedInstances.forEach((relatedInstance) => {
+    const isMainInstanceBoolean = false;
```
LauraStordeur, line 174: `isMainInstance` already implies it's a boolean :)

src/UI/Components/Diagram/actions.ts
```diff
@@ -135,30 +154,76 @@ export function appendInstance(
...
+        paper,
+        graph,
+        relatedInstance,
+        services,
+        isMainInstanceBoolean,
+        instanceAsTable,
+      );
+    } else {
+      //If cell is already in the graph, we need to check if it got in its inter-service relations the one with id that corresponds with created instanceAsTable
+      let connected = false;
```
LauraStordeur, line 187: it's common practice to have boolean values starting with "is" or "has". So it would be isConnected or hasConnections

src/UI/Components/Diagram/halo.ts
```diff
@@ -43,16 +44,29 @@ const createHalo = (
...
       if (isEmbedded && isEmbeddedToTHisCell) {
         element.set("embeddedTo", null);
+        didElementChange = true;
       }
       const relations = elementAsService.getRelations();
 
       if (relations) {
-        elementAsService.removeRelation(cellView.model.id as string);
+        didElementChange =
+          didElementChange === true ||
```
LauraStordeur, line 57: The or statement still confuses me. Does it mean that if the Element already changed, you don't remove the relation?

the removeRelation method serves two purposes, maybe it would be best to split it?

src/UI/Utils/useFeatures.tsx
```diff
@@ -11,7 +11,7 @@ const extractFeatures = (fileContent) => {
 };
 
 const useFeatures = (): string[] => {
-  const [features, setFeatures] = useState([]);
+  const [features, setFeatures] = useState(["instanceComposer"]);
```
LauraStordeur, line 14: Is this meant to land in production?

src/UI/Components/Diagram/actions.ts
matborowczyk: Yes, I couldn't find better name and `isMainInstance` is used in the function parameters itself :D Do you have any suggestions?

src/UI/Components/Diagram/halo.ts
matborowczyk: Yes, because if element changed, that means that connection that was made to that element is already removed, there is no way that we have both embedded entity connection and inter-service relation between two shapes

matborowczyk: I can definitely do that if you have feeling that it's confusing, I just accepted it, as the default behaviour of function on which we are basing is giving that opportunity not to delete and check whether something changed

src/UI/Utils/useFeatures.tsx
matborowczyk: Nope, sorry about that, I always double checking if I don't include that, and this time forgot when was applying remarks

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -21,6 +24,13 @@ const Canvas = ({
   mainServiceName: string;
   instance?: InstanceWithReferences;
 }) => {
+  const { environmentHandler } = useContext(DependencyContext);
+  const environment = environmentHandler.useId();
+  const baseUrlManager = new PrimaryBaseUrlManager(
+    globalThis.location.origin,
+    globalThis.location.pathname,
+  );
+  const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
```
LauraStordeur, line 33: I believe this is available in the DependencyContext. 
You should find it with ` urlManager.getApiUrl()` if you add the urlManager to line 27.

src/UI/Components/Diagram/actions.ts
LauraStordeur: If you end up needing the same name, maybe it's worth extracting the logic 🤔 Or maybe it means there is some overcomplication somewhere. Maybe the appendInstance needs to accept a type of instance, rather than a boolean value defining if it's a main instance or not. You'd then have an Enum InstanceType = "main" | "whichever"

src/UI/Components/Diagram/halo.ts
LauraStordeur: yeah, but if you read the code now, you assume that you'll delete the relation if the element hasn't changed and it has relations.

LauraStordeur (approved):

PR #5149, opened 2023-09-22
set BigInt serialization as default

### LINKED ISSUE #5147: fix BigInt serialization and enable by default
Currently, a setting controls whether BigInts are expected and should be taken into account when serializing json ([here](https://github.com/inmanta/web-console/blob/45d6a6006bee651a0b8045bb85212fe810812a68/src/Injector.tsx#L60)). This controls the parsing of json into `BigInt`. However, these `BigInt` values are then converted back [as strings](https://github.com/inmanta/web-console/blob/45d6a6006bee651a0b8045bb85212fe810812a68/src/Core/Language/Extensions.ts#L8), which are rejected by the server.

There are 2 points of action for this ticket:
- [ ] fix the bug, e.g. by using [json-bigint](https://github.com/sidorares/json-bigint)'s `stringify` instead of the `toString` approach. I believe this means changing the stringify calls [here](https://github.com/inmanta/web-console/blob/master/src/Data/API/BaseApiHelper.ts#L182).
- [ ] run some benchmarks on serialization and deserialization of json (without large ints in them) using both the normal `JSON` functions and `json-bigint`'s json functions. If there is no significant performance impact, simply always use the big int one and ignore the option. Create a follow-up ticket on the inmanta-ui repo to deprecate the option.

matborowczyk:
# Description

* Short description here *

closes #5147 

# Self Check:

- [x] Changelog entry


src/Test/Mock/DeferredApiHelper.test.ts
```diff
@@ -101,3 +101,82 @@ test("GIVEN DeferredApiHelper WHEN resolveRequest THEN resolves first matching r
...
+  const apiHelper = new DeferredApiHelper();
+
+  //POST
+  apiHelper.post<{ foo: bigint }, { foo: bigint }>("", "", {
+    foo: 9223372036854775807n,
+  });
+  apiHelper.pendingRequests;
+  expect(apiHelper.pendingRequests[0]).toEqual({
+    body: {
+      foo: 9223372036854775807n,
```
LauraStordeur, line 115: maybe a stupid question, but can we send the request with the "n" at the end of the BigInt ? Is that supported ?

LauraStordeur (approved):

PR #5158, opened 2023-10-03
Jest jointjs

matborowczyk:
# Description

It's part one so I won't close ticket yet

closes  --- 

# Self Check:

- [x] Changelog entry


jestSetupAfterEnv.js
```diff
@@ -6,3 +6,7 @@ import "@testing-library/jest-dom";
  * we increase the default timeout here.
  */
 jest.setTimeout(10000);
+
+//JointJS mock to make library work
+window.SVGAngle = jest.fn();
```
LauraStordeur, line 11: why did you need it twice?

src/Test/Utils/custom-queries.ts
```diff
@@ -0,0 +1,38 @@
+import {
+  queryHelpers,
+  buildQueries,
+  Matcher,
+  MatcherOptions,
+} from "@testing-library/react";
+
+const getMultipleError = (attrName: string) => (_c, selectorValue) =>
```
LauraStordeur, line 8: aren't these supposed to be in the "words" file?

LauraStordeur: The composer is part of the app, the words file is a dictionary of any pieces of text (or at least should, there are places where it wasn't used out of laziness I assume) :D So you can add those two in the file :D

matborowczyk: I wouldn’t mix error messages from tests with displayed text, but I don't mind, I'll add those :D

LauraStordeur (approved):

PR #5223, opened 2023-11-02
update endpoint for support archives

### LINKED ISSUE #5218: Replace download archive endpoint with new one.
There is a new endpoint available to download support archives; the current logic should be replaced with it.

`/api/v2/support `

matborowczyk:
# Description

I was trying to check in the docs whether should I put `/full` or not and then realised that v2 link aren't there yet so I think that will await until available

closes #5218 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5218-support-archive.yml
+++ b/changelogs/unreleased/5218-support-archive.yml
@@ -1,0 +1,6 @@
+change-type: patch
+description: "Update support archive link for v2"
+issue-nr: 5218
+destination-branches:
+  - master
+  - iso6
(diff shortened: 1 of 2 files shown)


LukasStordeur:
@matborowczyk  Do we still need the zip-building logic when using the new endpoint? Based on what Wouter told me, we might not need to care anymore and just trigger a download.

LukasStordeur:
Can you also delete the old code related to generating the Zip and remove the JSZip dependencies? :D

changelogs/unreleased/5218-support-archive.yml
```diff
@@ -0,0 +1,6 @@
+change-type: patch
+description: "Update support archive link for v2"
+issue-nr: 5218
+destination-branches:
```
LukasStordeur, line 4: Can you add a section - minor-improvement for this one? It won't be displayed in the outputed changelog otherwise.

LukasStordeur (approved):

PR #5243, opened 2023-11-09
5166 duplicate instance

LukasStordeur:
# Description

Implement the new feature to duplicate instances. 
In the process, the buttons present in the expanded rows have been moved into a kebab-menu at the end of each row. 
The e2e test are using the duplicate instance option whenever a second instance is needed in the test. 

https://github.com/inmanta/web-console/assets/44098050/16d39958-2942-43bd-b767-a1acc3fc3bae

image


src/Slices/DuplicateInstance/UI/DuplicateInstancePage.test.tsx
```diff
@@ -0,0 +1,481 @@
...
+  await act(async () => {
+    await userEvent.click(screen.getByText(words("confirm")));
+  });
+
+  expect(apiHelper.pendingRequests).toHaveLength(1);
+
+  expect(apiHelper.pendingRequests[0]).toEqual({
+    method: "POST",
+    url: `/lsm/v1/service_inventory/${service_entity}`,
+    body: {
```
matborowczyk, line 116: it seems like body of the request is the same here and in the line 214 so we could extract it, but it's hard to compare those from this view so I could me mistaken

matborowczyk (approved):

src/Slices/DuplicateInstance/UI/DuplicateInstancePage.test.tsx
LukasStordeur: they are almost the same, in the second one a value in the embedded entity has been changed.

PR #5263, opened 2023-11-13
Change pagination to patternfly component

### LINKED ISSUE #5185: replace pagination component with PF Pagination Element
PF has their own paginationWidget now. To avoid restyling the dropdown to look like a simple arrow, we should just use the ready-made component. Current implementation is entirely self coded and contains deprecated elements.

matborowczyk:
# Description

* Short description here *

closes #5185 

# Self Check:

- [x] Changelog entry

Screenshot 2023-11-13 at 15 31 00


src/UI/Components/PaginationWidget/Provider.tsx
```diff
@@ -1,7 +1,8 @@
 import React from "react";
+import { Pagination as PaginationComponent } from "@patternfly/react-core";
```
LukasStordeur, line 2: Was the name reassignment really needed?

src/UI/Utils/useFeatures.tsx
```diff
@@ -11,7 +11,7 @@ const extractFeatures = (fileContent) => {
 };
 
 const useFeatures = (): string[] => {
-  const [features, setFeatures] = useState([]);
+  const [features, setFeatures] = useState(["instanceComposer"]);
```
LukasStordeur, line 14: Probably not meant to be comitted?

src/UI/Components/PaginationWidget/Provider.tsx
matborowczyk: We have interface with the same name, so I decided that is better to change this component name than hunt for interface across the app :D

LukasStordeur: Fair enough :D

src/UI/Utils/useFeatures.tsx
```diff
@@ -11,7 +11,7 @@ const extractFeatures = (fileContent) => {
 };
 
 const useFeatures = (): string[] => {
-  const [features, setFeatures] = useState([]);
+  const [features, setFeatures] = useState([""]);
```
LukasStordeur, line 14: It was an empty array initially, maybe best to keep it that way.

matborowczyk: my bad

LukasStordeur (approved):

Issue #5275, opened 2023-11-16
Composer - Disable Deploy button when no changes


LukasStordeur:
When no changes have been made to the model, we should disable the button.

matborowczyk:
Should we recognise  if instance will return to its original attributes? here it's probably easier to implement than in regular Instance Form - but I don't think it will be that easy if user deletes shape(embedded or not) in composer and then creates shapes with the same attributes.

LukasStordeur:
I suppose the easiest path is to enable the button from the moment the user submits any type of form

matborowczyk:
That's right, I did that, but then started thinking if we want go further with that, but we can go back after I'll do everything from current epic

LukasStordeur:
Yeah, I suspect the behavior will change anyway when you start working on persisting the positioning for example.

PR #5306, opened 2023-11-24
Feedback on deploy + block Deploy button when no changes/input + redirect

### LINKED ISSUE #5128: Composer - Add user feedback on deploy
Add toaster on deploy action, informing the user that an order has been placed.

matborowczyk:
# Description

Bundled "Blocking Deploy button" ticket with deploy feedback, currently forms "updates" on confirm even though there is no changes, do we want to block the button, or just in case of no changes then just close the modal? I would personally opt for the second one, but want to read what you think.

**Sidenote:** This bug with suggestions tooltip is due to the me recording mobile devTools version of the app, which is set to 75% of normal size, when app is running "normally" then tooltip is where it should be

closes #5128 #5275 

# Self Check:

- [x] Changelog entry
~~- [ ] Sufficient test cases (reproduces the bug/tests the requested feature)~~ I think that will be covered through e2e

https://github.com/inmanta/web-console/assets/113331659/ea9426c7-6d05-4112-90cc-ae465a9d948d

### DIFF
--- a/src/UI/Components/Diagram/Canvas.tsx
+++ b/src/UI/Components/Diagram/Canvas.tsx
@@ -1,16 +1,21 @@
 import React, { useContext, useEffect, useRef, useState } from "react";
 import "@inmanta/rappid/rappid.css";
+import { useNavigate } from "react-router-dom";
 import { dia } from "@inmanta/rappid";
+import { AlertVariant } from "@patternfly/react-core";
+import { isObject } from "lodash";
 import styled from "styled-components";
-import { ServiceModel } from "@/Core";
+import { objectHasKey, ServiceModel } from "@/Core";
 import { InstanceWithReferences } from "@/Data/Managers/GetInstanceWithRelations/interface";
 import diagramInit, { DiagramHandlers } from "@/UI/Components/Diagram/init";
 import { CanvasWrapper } from "@/UI/Components/Diagram/styles";
 import { DependencyContext } from "@/UI/Dependency";
+import { words } from "@/UI/words";
+import { ToastAlert } from "../ToastAlert";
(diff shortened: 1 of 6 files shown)


src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -181,6 +233,7 @@ const Canvas = ({
         onConfirm={(entity, selected) => {
           if (diagramHandlers) {
             if (cellToEdit) {
+              console.log(cellToEdit);
```
LukasStordeur, line 236: Probably a forgotten one ;)

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -49,27 +63,42 @@ const Canvas = ({
...
-    const mapToArray = Array.from(instancesToSend, (instance) => instance[1]); //only value, the id is stored in the object anyway
-    const topServicesNames = services.map((service) => service.name);
-    const topInstances = mapToArray.filter((instance) =>
-      topServicesNames.includes(instance.service_entity),
-    );
-    const embeddedInstances = mapToArray.filter(
-      (instance) => !topServicesNames.includes(instance.service_entity),
-    );
+    try {
+      const response = await fetch(`${baseUrl}/lsm/v2/order`, {
```
LukasStordeur, line 67: I think we can make a ticket, to at least extract all API interactions into a neat folder. If we want to replace it at a future stage with a better way, it will be easier to handle it.

matborowczyk: Created separate ticket for that

LukasStordeur (approved):

PR #5323, opened 2023-11-28
block editing of certain instances in composer

### LINKED ISSUE #5316: COMPOSER - block edit of certain services
The Order API relies on the PATCH endpoint to update the attributes of a service instance. This means that the Order API only supports updates on instances of service entities that have strict_modifier_enforcement enabled. I wasn't sure whether there was sufficient awareness about this in the front-end team, that's why I wanted to mention it here explicitly.

EXAMPLE:
I got a tree of instances displayed on my canvas in the composer, and some of those instances can't be updated because the don't have the attribute modifier enforced. But if I change some of the attributes of an instance that allows it, and deploy the entire tree, the endpoint will support that.

TASK: Block the edit of instances that don't have the `strict_modifier_enforced` set to true.

matborowczyk:
# Description

Added Blocker, it seems to work in proper direction as shown in the video, as far as I remember there are some issues with the hovering when combining jest and I opt to test as much as possible through E2E and eventually refactor a bit functions from "actions" that it could be possible to write unit tests, but all the visual stuff should go through cypress, and we could then take files tested in that way out of coverage to have more realistic view on what is tested and what not

e2e file change is because it was flaky few times

closes #5316 

# Self Check:

- [x] Changelog entry

https://github.com/inmanta/web-console/assets/113331659/55321ba9-7171-431f-942c-968ce905ccbd


src/UI/Components/Diagram/Canvas.test.tsx
```diff
@@ -747,6 +747,7 @@ describe("Canvas.tsx", () => {
   it("creates valid object for order_api when creating entity from scratch", () => {});
   it("creates valid object for order_api when editing entity", () => {});
 
+  it("WHEN strict_modifier_enforcement is set to false for some services THEN for only these instances and their connections editing/deleting is blocked", () => {});
```
LukasStordeur, line 750: can you make a ticket for those tests so we don't forget about them?

src/UI/Components/Diagram/actions.ts
```diff
@@ -258,6 +271,7 @@ export function appendEmbeddedEntity(
   holderType: string,
   instanceToConnectRelation?: ServiceEntityBlock,
   presentedAttr?: "candidate" | "active",
+  isBlockedFromEditing?: boolean,
```
LukasStordeur, line 274: isEditable?

src/UI/Components/Diagram/actions.ts
```diff
@@ -455,15 +489,20 @@ export function appendColumns(
  * @param {dia.Graph} graph JointJS graph object
  * @param {ServiceEntityBlock} source JointJS shape object
  * @param {ServiceEntityBlock} target JointJS shape object
+ * @param {bolean} isBlocked parameter determining whether we are showing tools for linkView
```
LukasStordeur, line 492:
```suggestion
 * @param {boolean} isBlocked parameter determining whether we are showing tools for linkView
```

LukasStordeur:
I asked the question on slack, but I don't know if deleting a link should be blocked or not.

src/UI/Components/Diagram/Canvas.test.tsx
matborowczyk: Done, added to New and to Epic

src/UI/Components/Diagram/actions.ts
matborowczyk: I prefer to go by triggering only when some instance is blocked from editing, so by default newly created instances doesn't have to have extra parameter in that approach

LukasStordeur: isEditable is for me just meaning the same, it's what's used on other places too. I don't see how changing the name would make it any different when initializing the instances, or how that would mean an additional parameter.

LukasStordeur:
> @LukasStordeur In case of Instances with embedded entities that would introduce issue where user could try to delete or change these embedded instances. I know that its an edge case but still. If deleting is something that is desired for these kind of instances, then I would introduce separate functionality that would delete all of shapes that creates given instance, but would need separate ticket, which I can create and add to the queue

I'm not entirely sure I can follow, but we can elaborate about it in next meeting, can you make a ticket to review the functionality.?

src/UI/Components/Diagram/actions.ts
matborowczyk: For me, semantically `IsEditable === !isBlockedFromEditing`, and to keep in line naming and implementation then I would have to add that every possible shape is editable by default, which is redundant. I can change naming, but then in 6 months it could be confusing that something is set to true but its blocked from editing :)

PR #5340, opened 2023-12-01
Resolve issues with embedded entities in composer

### LINKED ISSUE #5304: Deleting Embedded entities doesn't trigger changes in Instance composer

matborowczyk:
# Description

I had to do deep copies of instances when doing bundling as the changes were copied back to state and that caused wrong state of the instances.
Also now deleting/disconnecting embedded entity trigger correctly "update"

closes #5304 #5338 

# Self Check:

- [x] Changelog entry
- [] Sufficient test cases (reproduces the bug/tests the requested feature)

### DIFF
--- a/src/UI/Components/Diagram/Canvas.tsx
+++ b/src/UI/Components/Diagram/Canvas.tsx
@@ -137,8 +137,8 @@ const Canvas = ({
                 action: "delete",
                 value: null,
                 edit: null,
-                embeddedTo: null,
-                relatedTo: null,
+                embeddedTo: cell.attributes.embeddedTo,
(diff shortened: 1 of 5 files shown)


src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -182,10 +181,11 @@ const Canvas = ({
   }, [instance, services, mainServiceName]);
 
   useEffect(() => {
+    console.log(instancesToSend, bundleInstances(instancesToSend, services));
```
LukasStordeur, line 184: you can remove the log :)

Issue #5356, opened 2023-12-06
COMPOSER : Parsing form response incorrect for Dicts


LukasStordeur:
This is what the web-console sends through the normal instance creation form:
```
{
	"attributes": {
		"auto_update": false,
		"description": "",
		"flavor": "c2m2",
		"influxdb_db": "regression",
		"influxdb_tags": {},
		"inmanta_flavor": "iso7-dev",
		"monitoring": false,
		"name": "test.inmanta.com",
		"network": "local",
		"openstack_url": "node2.ii.inmanta.com",
		"owner": "sander"
	}
}
```
And this is what it sends via the composer:
```
{
	"attributes": {
		"auto_update": false,
		"description": "",
		"flavor": "c2m2",
		"influxdb_db": "regression",
		"influxdb_tags": "{}",
		"inmanta_flavor": "iso5",
		"monitoring": false,
		"name": "test.inmanta.com",
		"network": "local",
		"openstack_url": "node2.ii.inmanta.com",
		"owner": "sander"
	}
}

```
The insfluxdb_tags shouldn't be converted into a string when we send the response to the BE

See conversation : https://inmanta.slack.com/archives/C045Y42N49W/p1701795731291379?thread_ts=1701790480.254009&cid=C045Y42N49W

Issue #5354, opened 2023-12-05
Error in test output, Pagination


LukasStordeur:
![image](https://github.com/inmanta/web-console/assets/44098050/e46b324b-8a40-478b-b393-1312d8ec9e76)

There's a console.error in the tests, about the pagination widget in the Events.test.tsx file.

PR #5371, opened 2023-12-07
Fix parsing dict, upgrade select, fix test error

### LINKED ISSUE #5361: Upgrade Select in Diagram

matborowczyk:
# Description

Sanitise is added only on shapes that go through Form, as that the point where dict are getting "stringified".

closes #5361 #5356 #5354 

# Self Check:

- [x] Changelog entry


src/UI/Components/Diagram/components/FormModal.tsx
```diff
@@ -158,7 +170,7 @@ const FormModal = ({
     const tempPossibleForms = getOptions(services, [
       {
         key: "default_option",
-        value: "Choose a Service",
+        value: undefined,
```
LukasStordeur, line 173: Is this to be able to clear the select? If that's the case, default_option doesn't seem like the right key to me. Also, I wonder if we really do need that option at all

src/UI/Components/Diagram/components/FormModal.tsx
```diff
@@ -225,21 +235,19 @@ const FormModal = ({
...
-            isDisabled={cellView !== null}
           >
             {possibleForms.map(({ key, value }) => (
-              <SelectOption
-                key={key}
-                value={value}
-                isPlaceholder={value === "Choose a Service"}
-              />
+              <SelectOption key={key} value={value}>
+                {value || "Choose a Service"}
```
LukasStordeur, line 249: I think having a cross icon in the select toggle to clear the value is better than having an option displaying as Choose a Service

src/UI/Components/Diagram/components/FormModal.tsx
matborowczyk: After some thought I agree, removed

src/UI/Components/Diagram/components/FormModal.tsx
```diff
@@ -68,14 +71,24 @@ const FormModal = ({
...
+
   const onEntityChosen = useCallback(
-    (
-      _event,
-      value: string | SelectOptionObject,
-      isPlaceholder: boolean | undefined,
-      possibleForms: PossibleForm[],
-    ) => {
-      if (isPlaceholder) {
+    (value: string | number | undefined, possibleForms: PossibleForm[]) => {
```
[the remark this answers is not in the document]
matborowczyk: That was due to the typing that are from the onSelect in the components itself, but changed that now to casing in the component

Issue #5276, opened 2023-11-16
Composer- Button to add is slower than rest of the UI


LukasStordeur:
The Add in Composer button is sometimes displayed after the rest of the UI.

matborowczyk:
closed in #5387

PR #5387, opened 2023-12-13
Issue/5386 embedded entity composer

### LINKED ISSUE #5386: Fix embedded entities is instanceComposer

matborowczyk:
# Description

When preparing for QA session i noticed two bugs which I fixed 

PR also contain E2E which were written to double check if there was no more bugs in core use-cases, which were, so also fixes for those issues
Also fix for delayed "Add in composer" button

closes #5386 #5276 

# Self Check:

- [x] Changelog entry


src/UI/Components/Diagram/helpers.ts
LukasStordeur, line 1: I'm losing myself in the code in this file. There seems to be lots of repetitive pieces of code, and a whole chain of nested if cases. Either it needs a rewrite to be more readable, or add more comments to the new parts you added. Because if we come back to this later on, it will be a major headache to figure everything out again. It's about the entire for loop starting on 286.

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -84,7 +84,7 @@ const Canvas = ({
         setAlertMessage(words("inventory.instanceComposer.success"));
         setTimeout(() => {
```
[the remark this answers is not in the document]
matborowczyk: It was said to get confirmation notification on deploy, so I decided to show it in that view and after second to redirect into service inventory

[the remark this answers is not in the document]
LukasStordeur: Okay, thanks. Can you still add the comment to the code about this decision?

src/UI/Components/Diagram/helpers.ts
```diff
@@ -287,34 +288,52 @@ export const shapesDataTransform = (
...
-            updatedInstance.action !== null;
-
-          if (updatedInstance.action !== "delete") {
-            updated.push(updatedInstance.attributes as InstanceAttributeModel);
+          if (embeddedModel) {
+            const updatedInstance = shapesDataTransform(
+              notMatchingInstances,
+              instance,
+              embeddedModel,
+              true,
```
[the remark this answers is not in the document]
matborowczyk: I don't think it's necessary When we using function recursively so the definition of what's true is just above

src/UI/Components/Diagram/helpers.ts
matborowczyk: valid point, I'll extract it into separate function

src/UI/Components/Diagram/helpers.ts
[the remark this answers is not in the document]
LukasStordeur: am I right if `!!embeddedModel` tells you if it's embedded or not?

src/UI/Components/Diagram/interfaces.ts
```diff
@@ -21,6 +21,7 @@ interface DictDialogData {
 interface Rule {
   name: string;
   type: TypeEnum;
+  attrName?: string; //only used for inter-service relations
```
LukasStordeur, line 24: overall, we try to avoid abbreviations, I believe it's a good practice to maintain.

src/UI/Components/Diagram/init.ts
```diff
@@ -230,7 +227,10 @@ export default function diagramInit(
       );
 
       if (doesTargetHaveRule) {
-        targetCell.addRelation(sourceCell.id as string, sourceName);
+        targetCell.addRelation(
+          sourceCell.id as string,
+          doesTargetHaveRule.attrName as string,
```
LukasStordeur, line 232: `doesTargetHaveRule` is a boolean value for me when I read the variable name. Why is it an object?

src/UI/Components/Diagram/init.ts
```diff
@@ -219,7 +213,10 @@ export default function diagramInit(
       );
 
       if (doesSourceHaveRule) {
-        sourceCell.addRelation(targetCell.id as string, targetName);
+        sourceCell.addRelation(
+          targetCell.id as string,
+          doesSourceHaveRule.attrName as string,
```
LukasStordeur, line 218: same here, this is a boolean name, not inferring it might be an object.

src/UI/Components/Diagram/helpers.ts
```diff
@@ -284,45 +286,50 @@ export const shapesDataTransform = (
...
+      instance.attributes[key] =
+        updated.length === 1 && shouldEmbeddedBeSingular ? updated[0] : updated;
     }
   }
 
   //convert relatedTo property into valid attribute
   if (instance.relatedTo) {
-    instance.relatedTo.forEach((attrName, id) => {
+    instance.relatedTo.forEach((test, testOne) => {
+      console.log(test, testOne);
```
LukasStordeur, line 330: console.log

src/UI/Components/Diagram/helpers.ts
```diff
@@ -388,8 +395,15 @@ export const bundleInstances = (
   const embeddedInstances = deepCopiedMapToArray.filter(
     (instance) => !topServicesNames.includes(instance.service_entity),
   );
+  return topInstances.map((instance) => {
```
LukasStordeur, line 398: topInstances, are those the root instances?

src/UI/Components/Diagram/helpers.ts
```diff
@@ -388,8 +395,15 @@ export const bundleInstances = (
...
+      (service) => service.name === instance.service_entity,
+    );
 
-  return topInstances.map((instance) =>
-    shapesDataTransform(embeddedInstances, instance),
-  );
+    return shapesDataTransform(
+      embeddedInstances,
+      instance,
+      serviceModel as ServiceModel,
```
LukasStordeur, line 406: if you type serviceModel correctly in line 399, you're not supposed to cast it.

src/UI/Components/Diagram/helpers.ts
```diff
@@ -284,45 +286,50 @@ export const shapesDataTransform = (
...
-        instance.attributes[key] = updated;
-      } else {
-        const data = shapesDataTransform(
-          notMatchingInstances,
-          instancesToEmbed[0],
-          true,
-        );
-
-        areEmbeddedEdited = instance.action === null && data.action !== null;
+          shouldEmbeddedBeSingular =
```
LukasStordeur, line 310: I know performance is not much a concern, but allocating a variable like this isn't ideal. It would be more efficient to have a method to which you pass as argument the model and return whether it's singular or not. When I read the flow of code here, it makes me wonder why it has to be done between the two if cases too.

```
const isSingularRelation: boolean = (model: ModelType ) => { 
     !!model.upper_limit && model.upper_limit === 1;
}
```
I don't know by heart which name the Type is supposed to have for the model. But you get my drift. You can then simply pass the method in line 323. And also reuse the method wherever needed. If I remember well, it's a logic that appears now and then, to check if it's one or more relations.

src/UI/Components/Diagram/helpers.ts
matborowczyk: topInstance are non-embedded entities, as they at the top and receive eventual embedded ones

src/UI/Components/Diagram/helpers.ts
matborowczyk: the reason why it is casted as I need to find the serviceModel from the array to pass down into the function, there it becomes `serviceModel | undefined`, I can add if statement and then filter out possible undefined values from array, but theoretically we should have all serviceModels in the first place, that why I chose this solution

src/UI/Components/Diagram/init.ts
```diff
@@ -193,48 +194,56 @@ export default function diagramInit(
...
+        targetCell.addRelation(
+          sourceCell.id as string,
+          targetConnectionRule.attributeName as string,
+        );
         didTargetChanged = true;
+        didConnectionWasSet = true;
       }
     }
 
+    if (!didConnectionWasSet) {
```
LukasStordeur, line 231: You have a lot of repetition in here, which makes it hard to read. Why not extract the logic into a function? I quickly drafted it, you might need to make adjustments and add typings
```
const checkUpdated = (element) => {
    const cell = graph.getCell(
        element.id as dia.Cell.ID,
    ) as ServiceEntityBlock;

    const relations = cell.getRelations();
    const name = cell.getName();

    if (relations) {
        const elementConnectionRule = connectionRules[name].find((rule) => rule.name === name);
        if (elementConnectionRule) {
            cell.addRelation(
                sourceCell.id as string,
                targetConnectionRule.attributeName as string,
              );
            updateInstancesToSend(cell, ActionEnum.UPDATE);
            return; // you don't need further information, so you can early return the function
        }
    }

    if (cell.get("isEmbedded") &&
    cell.get("embeddedTo") !== null) {
        cell.set("embeddedTo", cell.id);
        updateInstancesToSend(cell, ActionEnum.UPDATE);
        return;
    }
}

const source = linkView.model.source();
const target = linkView.model.target();
checkUpdated(source);
checkUpdated(target);

```

src/Data/Managers/GetInstanceWithRelations/QueryManager.ts
```diff
@@ -83,33 +83,42 @@ export function GetInstanceWithRelationsQueryManager(
 
       if (Either.isRight(instanceWithReferences)) {
         if (instanceWithReferences.value.data.referenced_by !== null) {
-          await Promise.all(
+          const resolvedNestedInstances = await Promise.all(
```
LukasStordeur, line 86: There are a few downsides to use promise.all, but I'm not sure it's something we need to care about. 

1. It doesn't timeout, so if one of the operations hangs, it won't abort on it's own. 
2. It's all or nothing, if one element fails, the entire promise will fail. It makes it hard to have individual error-handling.
3. It isn't sequential, but executes all the promises concurrently.

src/UI/Components/Diagram/actions.ts
LukasStordeur, line 1: Looks like this is almost the same as in the init file, I'll make the same comment, extract the repetitive code, to avoid having those boolean variables keeping track of the changes.

src/UI/Components/Diagram/helpers.ts
```diff
@@ -284,47 +286,61 @@ export const shapesDataTransform = (
...
   if (instance.relatedTo) {
-    instance.relatedTo.forEach((attrName, id) => {
+    Array.from(instance.relatedTo).forEach(([id, attributeName]) => {
       if (instance.attributes) {
-        instance.attributes[attrName] = id;
+        const model = serviceModel.inter_service_relations?.find(
+          (relation) => relation.name === attributeName,
+        );
+        if (model) {
+          if (model.upper_limit !== 1) {
```
LukasStordeur, line 331: Use a switch case maybe, to make it more readable?

src/UI/Components/Diagram/init.ts
```diff
@@ -181,66 +181,54 @@ export default function diagramInit(
...
-
-    if (sourceRelations) {
-      const doesSourceHaveRule = connectionRules[sourceName].find(
-        (rule) => rule.name === targetName,
-      );
 
-      if (doesSourceHaveRule) {
-        sourceCell.addRelation(targetCell.id as string, targetName);
-        didSourceChanged = true;
+    const checkUpdated = (sourceElement, targetElement) => {
```
LukasStordeur, line 188: I see it's a bit better, but still confusing, why do we need to execute in both directions in lines 230 and 231? Can you also add some docs strings to the method?

src/UI/Components/Diagram/interfaces.ts
```diff
@@ -21,6 +21,7 @@ interface DictDialogData {
 interface Rule {
   name: string;
   type: TypeEnum;
+  attributeName?: string; //only used for inter-service relations
```
LukasStordeur, line 24: If this is required for inter-service relations, isn't it worth having a type extending the default rule set? I'm no expert in TS, so maybe I'm wrong :D But it seems like this is what has been done so far in the codebase

src/UI/Components/Diagram/actions.ts
```diff
@@ -258,7 +261,7 @@ export function appendInstance(
           if (neighborRelations) {
             const connectedInstance = Array.from(
               neighborRelations,
-              ([id, attrName]) => ({ id, attrName }),
+              ([id, attributeName]) => ({ id, attributeName }),
```
LukasStordeur, line 264: Isn't this the same as in line 238? if yes, maybe worth having it as a variable/function

src/UI/Components/Diagram/interfaces.ts
matborowczyk: but the function that use that interface accept both embedded and inter-service relation, so there is no other place  that use that interface so that would be dubious in that case

LukasStordeur: yes, I understand that, but it means that in the case of an inter-service relation, you have an optional where it would be a required field... Unless it's not required in the inter-service relation, then it's fine. But if it is required, then it biases your code, and enforces you to eliminate undefined cases, or cast values to bypass the linter.. 

Why not go for the kind:"InterserviceRelation" technique that is being used all over the rest of the app? It's just a suggestion

matborowczyk: I see that more like if there is that attributeName then it's inter-service_relation, if not then it's embedded, and it's simple enough.

Again I'm not strongly behind or against one of those solutions so I can adjust if you feel like current solution is unclear, I will take step back to look on the code, to check whether you proposition will improve readability, which could be the case, as I have biased POV at the moment as I stare at it for quite some time :)

src/UI/Components/Diagram/init.ts
matborowczyk: Because we can make the connections from the both sides, so we don't know which instance can get the data about being embedded/relatedTo, and previous version did that both way but there was some repetition, so avoiding repetition we need to check now from both perspectives, otherwise we have identical code but wrapped into extra function body.

I'll add docstring

LukasStordeur: What do you mean, we don't know which instance gets the data, don't you have that information 🤔 You know the two elements that are being concerned by the event, you have the target, and the source, two instances the should get an update. Now, the way you are using  this, you are mixing up source and target... 

I'm lost on the reasoning to be fair, maybe the Kind field in the typing would already resolve some of the complexity of all those if cases throughout the file(s). 

Extracting logic doesn't mean you have to plainly copy some content into a separate method. I made the example previously, as a direction to try to explain what I meant, I didn't have all the pieces of the reasoning. Sometimes it helps to write out in bullet points what you are trying to achieve. 

For me, the two instances get an event, either it's adding something, or removing something. Then I see what kind of thing I need to add or remove, and update accordingly.

src/UI/Components/Diagram/interfaces.ts
LukasStordeur: defining what kind of instance your dealing with, based on a field being present or not, is flakey and is not a clear solution that is maintainable. If someone new comes up in the team, and reads the code, they need explicit explanation to understand what's going on. That's something we don't want.

src/UI/Components/Diagram/init.ts
matborowczyk: This part of code checks which functions should get information about connection, as it's only stored in the children, I don't think that kind would help as theoretically both of them can be the same type and only one should get the data, so that's whats was done here.

We were checking both source and target, as depending from which instance we started it can be swapped on every attempt, that's why there was code repetition, we were checking the same thing on both instances, so as we now removed that into non-repetitive function we need to check it twice, for source and target

src/UI/Components/Diagram/actions.ts
matborowczyk: we check on the same things but the code in those if statement is different, so I don't think extracting here would be just for the sake of extracting

src/UI/Components/Diagram/init.ts
LukasStordeur: But aren't you already doing the check on both directions in lines 213 and further? 
Honestly, I'm drowning in the amount of nested if cases all over, I can't keep track of what's happening. 
Right now, the only thing that got removed were those boolean values, but the repetition and complexity is still there. 

Lets discuss this tomorrow, have a call about this PR, it will be more efficient than discussing here.

src/UI/Components/Diagram/actions.ts
LukasStordeur: I'm not wanting to extract just for the sake of it. Functions have arguments, for minor differences like you have in this file. The element you pass is either a target or a source, but inherently, they are the same thing, an instance. 

Having so many if cases like this makes the code not readable enough. Extracting methods with decent function names can improve readability, and get rid of some complexity. That's why you would extract, not just for the sake of it. 

I'm really trying to follow along what's going on in this file (and the init one), but there's so much look-alike code, that I'm drowning in it. I don't even how I'd try to make changes to this code without breaking my head as to why some if cases are used. Splitting up your code in smaller pieces, that make more sense, is not a bad practice. 

Anyway, let's discuss the PR tomorrow in the meeting.

LukasStordeur (approved):
I'll approve the PR as is now, Link it into the ticket about refactoring so we can go back to some of the comments later.

PR #5393, opened 2023-12-14
Issue/pagination fix iso6

matborowczyk:
# Description

* Short description here *

# Self Check:

- [x] Changelog entry


changelogs/unreleased/pagination-fix-iso6.yml
```diff
@@ -0,0 +1,3 @@
+description: "Fix for pagination in Service Inventory + tests"
```
LukasStordeur, line 1: I think it's not the idea to hide all our changelogs in the output. Only the ones that treat a same subject like the Composer, should be grouped under one visible changelog to avoid having twenty five logs about the same thing. Bugfixes, and other more generic issues can have a full changelog.

changelogs/unreleased/pagination-fix-iso6.yml
```diff
@@ -0,0 +1,3 @@
...
+change-type: patch
+destination-branches: [master, iso6]
```
LukasStordeur, line 3: you can remove master from this one.

LukasStordeur (approved):

PR #5435, opened 2024-01-09
Fix for optional boolean Input

matborowczyk:
# Description

* Short description here *

# Self Check:

- [x] Changelog entry


src/UI/Components/ServiceInstanceForm/Components/BooleanFormInput.tsx
```diff
@@ -64,12 +64,12 @@ export const BooleanFormInput: React.FC<Props> = ({
           <Radio
             isDisabled={shouldBeDisabled}
             isChecked={isChecked === null}
-            onChange={(event, value) => handleInputChange(value, event)}
+            onChange={(event, _value) => handleInputChange("null", event)}
```
LukasStordeur, line 67: But this means you will end up sending the string value instead of the real null 🤔

matborowczyk: the value is being passed through formatter, previously the value was "", I didn't think to much about it, if you want, I can change it.

LukasStordeur: yeah, but a sting null is not the same as a real null 🤔 what do we send back to the BE with this ?

matborowczyk: formatter used for these boolean  will accept everything expect true/false as null, but I changed it anyway to pass regular null

LukasStordeur (approved):
Don't forget to include these changes on the iso6 branch ;)

PR #5456, opened 2024-01-17
Improve Instance Composer Code

### LINKED ISSUE #5443: Refactor Instance Composer functions

matborowczyk:
# Description

* Short description here *

closes #5443 

# Self Check:

- [x] Changelog entry


src/UI/Components/Diagram/init.ts
```diff
@@ -212,10 +212,10 @@ export default function diagramInit(
         );
 
         //if there is corresponding rule we can apply connection and update given service
-        if (cellConnectionRule) {
+        if (cellConnectionRule && cellConnectionRule.attributeName) {
```
LukasStordeur, line 215: I remember the comment about the Typing and how you distinguished between one sort of Entity and another. I still don't agree with using this to know it's an interservice relation.

src/UI/Components/Diagram/components/FormModal.tsx
```diff
@@ -256,7 +256,7 @@ const FormModal = ({
             onOpenChange={(isOpen) => setIsSelectOpen(isOpen)}
             isOpen={isSelectOpen}
             onSelect={(_evt, value) => {
-              onEntityChosen(value as string, possibleForms);
+              onEntityChosen(String(value), possibleForms);
```
LukasStordeur, line 259: so, it can never be a number, null or undefined? 
I'd expect the onEntityChosen to handle the typing instead of mutating the value here 🤔

src/UI/Components/Diagram/actions.ts
```diff
@@ -300,90 +316,33 @@ export function appendEmbeddedEntity(
...
-              instanceAsTable,
-              [instanceToConnectRelation],
-              isBlockedFromEditing,
-            );
-          }
-        }
-      });
-    });
+  /**
+   *  Extracted logic append Embedded entity to reduce repeats of code
```
LukasStordeur, line 329: I don't think this is a good description honestly... the function is called handleSingleEntity, but the description says it's appending an Embedded entity 🤔

src/UI/Components/Diagram/actions.ts
```diff
@@ -35,7 +35,16 @@ export function showLinkTools(
     target.id as dia.Cell.ID,
   ) as ServiceEntityBlock;
 
-  const check = (cellOne: ServiceEntityBlock, cellTwo: ServiceEntityBlock) => {
+  /**
+   * checks if the connection between cells can be deleted
```
LukasStordeur, line 39: deleted, or hidden? method name tells something different.

src/UI/Components/Diagram/actions.ts
```diff
@@ -172,7 +184,10 @@ export function appendInstance(
   const serviceInstance = serviceWithReferences.instance.data;
   const serviceInstanceModel = services.find(
     (model) => model.name === serviceInstance.service_entity,
-  ) as ServiceModel;
+  );
+  if (!serviceInstanceModel) {
+    throw Error("missing Instance Model");
```
LukasStordeur, line 189: use the words dictionary please

src/UI/Components/Diagram/components/FormModal.tsx
matborowczyk: It can never be number, null or undefined, it was said that changing it to default(which was undefined) isn't something wanted there, so all of the select option have string value

src/UI/Components/Diagram/actions.ts
matborowczyk: deleted, as the only tool that is there is to delete connection, if attributes are rw the we are editing instance won't show the tool

LukasStordeur: Yeah, okay, but this should be in your docstrings. The method is called shouldHideLinkTools plural. and your description talks about deleting connections.

src/UI/Components/Diagram/shapes.ts
```diff
@@ -162,10 +162,10 @@ export class ServiceEntityBlock extends shapes.standard.HeaderedRecord {
               ellipsis: true,
             },
           );
-
+          console.log(reproducedDisplayText);
```
LukasStordeur, line 165: console.log

cypress/e2e/scenario-8-instance-composer.cy.js
```diff
@@ -551,8 +551,8 @@ if (Cypress.env("edition") === "iso") {
       );
     });
 
-    it("8.4 edit instances inter-service relation", () => {
-      // Select 'test' environment
+    it.only("8.4 edit instances inter-service relation", () => {
+      // // Select 'test' environment
```
LukasStordeur, line 555: double comment

cypress/e2e/scenario-8-instance-composer.cy.js
```diff
@@ -551,8 +551,8 @@ if (Cypress.env("edition") === "iso") {
       );
     });
 
-    it("8.4 edit instances inter-service relation", () => {
-      // Select 'test' environment
+    it.only("8.4 edit instances inter-service relation", () => {
```
LukasStordeur, line 554: careful to not keep this in the end commit

matborowczyk:
new default size for the entities 
Screenshot 2024-01-18 at 15 31 43

src/UI/Components/Diagram/actions.ts
LukasStordeur: still waiting for a better description :)

src/UI/Components/Diagram/actions.ts
```diff
@@ -434,7 +393,13 @@ export function appendEmbeddedEntity(
         }
       }
     });
-    return [instanceAsTable];
+    return instanceAsTable;
+  }
+
+  if (Array.isArray(entityAttributes)) {
```
LukasStordeur, line 399: How come this can be something else than an array at this point? Genuine question.

src/UI/Components/Diagram/helpers.test.ts
```diff
@@ -95,21 +95,21 @@ describe("createConnectionRules", () => {
           lowerLimit: 1,
           upperLimit: 1,
           modifier: "r",
-          type: "embedded",
+          kind: "embedded",
```
LukasStordeur, line 98: A specific reason why you went for a lower-cased value here? In the rest of the code base the kind field contains a title-cased value.

src/UI/Components/Diagram/actions.ts
matborowczyk: `entityAttributes` can be just one standalone object, `rw_meta` vs `rw_files`, so attributes passed to that function can be both and that's why we differentiate

LukasStordeur: Can you add this piece of information in the docstrings where the argument is defined?

matborowczyk: the type of the attribute is telling the same piece of the information. Do you meant to add an example?

LukasStordeur: No I mean the info about the rw_meta and rw_files 🤔 unless it's already explained in the initial typing

matborowczyk: I wouldn't put some information about attributes from some specific entity as they can change in the future and we will have some outdated information. I think that it's quite well documented already, as `InstanceAttributeModel` is type for actual attributes, and in that function typing says that it's either one specific object, or array of objects

LukasStordeur: yeah, okay, but that info you are referring to seems to be part of some logic, which might be worth keeping 🤔

matborowczyk: But isn't it just clear from the function body itself? the internal function deals with the logic for single embedded entity, and after that function we call it, either in map if those embedded instances are in array, or directly for one single time when it's single embedded instance. And the information that the embedded instance can be as single object, but also in the array is not knowledge specific to the instanceComposer

LukasStordeur: I didn't mean add it here... I meant the little piece of information about what is rw_files and what is rw_meta... You know what it is, I don't. This for me a hint to add it somewhere. Not perse here, just somewhere...

matborowczyk: Oh, okay, I thought you meant here. Anyway still don't think like we should explain attributes from test env services into the frontend docs, primarily because I now mainly the type they have, but that's the information gathered from watching types and looking on service model, but they can change so it would be pointless to track them

LukasStordeur (approved):

PR #5499, opened 2024-02-06
update dependencies part 2

LukasStordeur:
# Description

Second part to resolve remaining dependency security alerts.

### DIFF
--- a/changelogs/unreleased/dep-update.yml
+++ b/changelogs/unreleased/dep-update.yml
@@ -1,0 +1,3 @@
+description: Update dependency resolutions to resolve security alerts.
+change-type: patch
+destination-branches: [master, iso6, iso5]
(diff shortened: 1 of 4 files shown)


matborowczyk (approved):

changelogs/unreleased/dep-update.yml
```diff
@@ -0,0 +1,3 @@
+description: Update dependency resolutions to resolve security alerts.
+change-type: patch
+destination-branches: [master, iso6, iso5]
```
matborowczyk, line 3: There is no need to update iso5 as it's dropped out of the support

matborowczyk: and more importantly not used by any of the customers

LukasStordeur: Sorry, I missed those when using the template to make the changelog file. Will also update the changelog template so this doesn't happen anymore

Issue #5310, opened 2023-11-24
Extract - Api calls into separate folder


matborowczyk:

matborowczyk:
Calls from Instance Composer won't be part of Query Manager, we can extract them into one place for time until we decide which tools are gonna be used to manage queries from now on - January hackathon

PR #5514, opened 2024-02-09
Add description to the call and extract it from the canvas

### LINKED ISSUE #5469: [Composer] Add default description
When doing the deploy call to the order API, atm there's no default description. This results in having an order in the table that has no readable description. 

Suggested description: "Requested with Instance Composer"

matborowczyk:
# Description

It's bundled as it was targeting the same piece of code

closes #5469 #5310 

# Self Check:

- [x] Changelog entry
~~- [] Sufficient test cases (reproduces the bug/tests the requested feature)~~
~~- [ ] Correct, in line with design~~

### DIFF
--- a/changelogs/unreleased/5469-composer-description.yml
+++ b/changelogs/unreleased/5469-composer-description.yml
@@ -1,0 +1,4 @@
+description: "Add description to the order api request and extract the call from the Canvas component"
+issue-nr: 5469
+change-type: patch
+destination-branches: [master, iso7]
--- a/changelogs/unreleased/5469-composer-description.yml
+++ b/changelogs/unreleased/5469-composer-description.yml
@@ -2,3 +2,5 @@ description: "Add description to the order api request and extract the call from
 issue-nr: 5469
 change-type: patch
 destination-branches: [master, iso7]
+sections:
+  minor-improvement: "{{description}}"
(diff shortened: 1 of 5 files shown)


src/UI/Components/Diagram/api/orderRequest.ts
```diff
@@ -0,0 +1,20 @@
+import { words } from "@/UI/words";
+import { InstanceForApi } from "../interfaces";
+
+export const sendOrder = (
```
LukasStordeur, line 4: PostOrder (based on the semantics of API requests)

Can you add docstrings?

LukasStordeur (commented):
Could you add a test case for the method? sendOrder/postOrder

changelogs/unreleased/5469-composer-description.yml
```diff
@@ -0,0 +1,4 @@
+description: "Add description to the order api request and extract the call from the Canvas component"
+issue-nr: 5469
+change-type: patch
+destination-branches: [master, iso7]
```
LukasStordeur, line 4: missing section

matborowczyk: I removed description from that as that doesn't influence end-user

LukasStordeur: It doesn't always need to affect the end user to be part of the changelog. The lack of a description was visible in the order view.

LukasStordeur (approved):

PR #5522, opened 2024-02-12
Remove feature flag for composer

### LINKED ISSUE #5470: [Composer] Remove feature flag for composer
Currently, there's no documentation on how to activate the composer for local development. It is by default enabled on the local-setup, but when we proxy, the local folder structure is overridden, and the config from the local-setup ignored. 

It would be an improvement to maybe work with a ENV variable to toggle the feature on local development. 

- [ ] add documentation how the featureFlag can be used on local development when proxying.
- [ ] try to adjust the current implementation to use ENV variables.

matborowczyk:
# Description

closes #5470

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5470-remove-featureflag.yml
+++ b/changelogs/unreleased/5470-remove-featureflag.yml
@@ -1,5 +1,5 @@
 description: Remove Instance Composer feature flag and make it accessible to all users
-issue-nr: 5512
+issue-nr: 5470
 change-type: patch
 destination-branches: [master, iso7]
 sections:
--- a/changelogs/unreleased/5470-remove-featureflag.yml
+++ b/changelogs/unreleased/5470-remove-featureflag.yml
@@ -1,4 +1,4 @@
-description: Remove Instance Composer feature flag and make it accessible to all users
+description: Remove The Composer feature flag as it is not in beta version anymore
 issue-nr: 5470
 change-type: patch
 destination-branches: [master, iso7]
--- a/changelogs/unreleased/5470-remove-featureflag.yml
+++ b/changelogs/unreleased/5470-remove-featureflag.yml
@@ -1,4 +1,4 @@
-description: Remove The Composer feature flag as it is not in beta version anymore
+description: Smart Composer is now general available
 issue-nr: 5470
 change-type: patch
 destination-branches: [master, iso7]
--- a/src/Core/Domain/Route.ts
+++ b/src/Core/Domain/Route.ts
@@ -51,7 +51,7 @@ const kinds = [
 
 export type RouteKind = (typeof kinds)[number];
 
-export type RestrictedRouteKind = "InstanceComposer" | "InstanceComposerEditor";
+export type RestrictedRouteKind = "";
 
 export const isValidKind = (value: string): value is RouteKind =>
   kinds.includes(value as RouteKind);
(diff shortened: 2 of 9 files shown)


changelogs/unreleased/5470-remove-featureflag.yml
```diff
@@ -0,0 +1,6 @@
+description: Remove Instance Composer feature flag and make it accessible to all users
```
LukasStordeur, line 1: I wouldn't say all users because this isn't available on OSS for example. 
Perhaps saying the Composer is not in beta version anymore might be better?

matborowczyk: Right, forgot about oss. I'll change that, it's bold, but I guess we are out of beta if we are making it accessible by default

src/Slices/ServiceInventory/UI/Components/RowActionsMenu/RowActions.tsx
```diff
@@ -64,9 +62,6 @@ export const RowActions: React.FunctionComponent<InstanceActionsProps> = ({
     // -100 for each default option present above the first drilldown that is enabled.
     // diagnose and duplicate are always enabled, so this means it starts at -200
     let insetHeight = -200;
-    if (features && features.includes("instanceComposer") && !editDisabled) {
-      insetHeight = insetHeight - 100;
-    }
     if (!editDisabled) {
       insetHeight = insetHeight - 100;
```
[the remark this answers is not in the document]
matborowczyk: Right below it is the code that does the same, and deleted part required "instanceComposer" feature flag which won't be there from now on.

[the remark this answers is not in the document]
LukasStordeur: yeah, no, that's the thing, previously, when you had the composer option present and enabled, it would deduce 100. Now the option is there by default, which means you'll have an issue with the offset in the dropdown if you open one of the drilldown menus. So you should deduce 200 in line 66, because now, we have two options that can be disabled/enabled. It's -100 for each enabled option.

src/Core/Domain/Route.ts
```diff
@@ -51,7 +51,7 @@ const kinds = [
 
 export type RouteKind = (typeof kinds)[number];
 
-export type RestrictedRouteKind = "InstanceComposer" | "InstanceComposerEditor";
+export type RestrictedRouteKind = "";
```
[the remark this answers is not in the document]
matborowczyk: Added

src/Slices/ServiceInventory/UI/Components/RowActionsMenu/RowActions.tsx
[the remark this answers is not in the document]
matborowczyk: Oh, I misunderstood it, changed!

changelogs/unreleased/5470-remove-featureflag.yml
matborowczyk: @bartv do you have any opinion on that?

bartv: I would write: Smart Composer is now general available.

matborowczyk: Changed

LukasStordeur (approved):

PR #5553, opened 2024-02-21
Make Composer available only from root

### LINKED ISSUE #5471: [Composer] Only make Composer accessible from Root Instance
When you create multiple instances and dependencies, you get redirected to the Service Inventory, which could perfectly remain empty if the user didn't create an instance of that type. 

Maybe it would be nicer to either redirect to the OrdersPage or the Service Catalog.

matborowczyk:
# Description

* Short description here *

closes #5471 

# Self Check:

- [x] Changelog entry
Screenshot 2024-02-21 at 21 54 43
Screenshot 2024-02-21 at 21 54 30


src/Slices/ServiceInventory/UI/Components/RowActionsMenu/RowActions.tsx
```diff
@@ -166,11 +168,11 @@ export const RowActions: React.FunctionComponent<InstanceActionsProps> = ({
               service: instance.service_entity,
               instance: instance.id,
             })}
-            isDisabled={editDisabled}
+            isDisabled={editDisabled || composerDisabled}
```
LukasStordeur, line 171: Do we want to disable it or hide it 🤔 I think hiding it would be more meaningful. There's no case where the button might go from disabled to enabled in the same view, unlike the edit button for example that can become enabled because of a state change.

src/UI/Components/Diagram/Mock.ts
```diff
@@ -3820,7 +3827,8 @@ export const childModel: ServiceModel = {
   },
   config: {},
   service_identity: "name",
-
+  owner: "parent-service",
+  owned_entities: [],
```
LukasStordeur, line 3831: perhaps would be valuable to have a mock that has owned_entities and use it in a test case to cover that scenario?

src/Slices/InstanceComposerEditor/Core/helper.ts
```diff
@@ -0,0 +1,26 @@
...
+  mainService: ServiceModel,
+): ServiceModel[] => {
+  const relatedServices = services.filter((service) =>
+    mainService?.owned_entities.includes(service.name),
+  );
+  const unrelatedServices = services.filter(
+    (service) => !mainService?.owned_entities.includes(service.name),
+  );
+
+  const nestedRelatedServices = relatedServices
```
LukasStordeur, line 21:
```
const nestedRelatedServices = relatedServices
  .flatMap((service) => filterServices(unrelatedServices, service));
```

src/Slices/InstanceComposerEditor/Core/helper.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @param {ServiceModel[]} services - An array of ServiceModel objects to filter.
+ * @param {ServiceModel} mainService - root service to filter owned services by it.
+ * @returns {ServiceModel[]} An array of owned ServiceModel objects
+ */
+export const filterServices = (
+  services: ServiceModel[],
+  mainService: ServiceModel,
+): ServiceModel[] => {
+  const relatedServices = services.filter((service) =>
```
LukasStordeur, line 14: I would personally be more defensive. Especially since this is a recursive method 
```

export const filterServices = (
  services: ServiceModel[] = [],
  mainService: ServiceModel | undefined,
): ServiceModel[] => {
  if (!mainService || !mainService.owned_entities || !Array.isArray(services)) {
    return [];
  } 
// rest of the code
```

src/Slices/InstanceComposerEditor/Core/helper.ts
```diff
@@ -0,0 +1,26 @@
...
+    mainService?.owned_entities.includes(service.name),
```
LukasStordeur, line 15: If `mainService`  doesn't have the property, the expression` mainService?.owned_entities` will also evaluate to undefined, and trying to call `includes` on undefined will result in a runtime error.

`Array.isArray(mainService.owned_entities) && mainService.owned_entities.includes(service.name)` Could be a more defensive alternative.

src/Slices/InstanceComposerEditor/Core/helper.ts
```diff
@@ -0,0 +1,26 @@
...
+  );
+  const unrelatedServices = services.filter(
+    (service) => !mainService?.owned_entities.includes(service.name),
```
LukasStordeur, line 18: Same observation here, the fact you are needing to use the optional chaining operator is Typsecript telling you there is a loophole where the value could potentially be null or undefined. 

I wouldn't mind it if it was used in the template (most often Angular practice I believe) but here, you're chaining a function to it that could cause a runtime error

`(service) =>
      !Array.isArray(mainService.owned_entities) || !mainService.owned_entities.includes(service.name),`

LukasStordeur:
I believe your screenshots are from your previous PR :)

matborowczyk:
> I believe your screenshots are from your previous PR :)

they are correct, were, as we want to hide the buttons

src/UI/Components/Diagram/Mock.ts
matborowczyk: it's covered in the test third test case

src/Slices/InstanceComposerEditor/Core/helper.ts
matborowczyk: changed

src/Slices/InstanceComposerEditor/Core/helper.ts
matborowczyk: Added

matborowczyk:
Screenshot 2024-02-22 at 15 59 39
Screenshot 2024-02-22 at 15 59 24

LukasStordeur (approved):

PR #5565, opened 2024-02-27
Add highlighter for loose embedded entities

### LINKED ISSUE #5366: Composer : Handle lose elements on canvas
Add user feedback concerning lose elements on the canvas.

matborowczyk:
# Description

Added functionality with some minor improvements

closes #5366 

# Self Check:

- [x] Changelog entry

https://github.com/inmanta/web-console/assets/113331659/317a9ba9-717d-4ff7-b037-536f910d1862

### DIFF
--- a/src/UI/Components/Diagram/Canvas.tsx
+++ b/src/UI/Components/Diagram/Canvas.tsx
@@ -62,11 +62,16 @@ const Canvas = ({
     );
     if (eventData.kind === "remove") {
       setLooseEmbedded((prevSet) => {
-        prevSet.delete(eventData.id);
-        return new Set(prevSet);
+        const newSet = new Set(prevSet);
+        newSet.delete(eventData.id);
+        return newSet;
       });
     } else {
-      setLooseEmbedded((prevSet) => new Set(prevSet.add(eventData.id)));
+      setLooseEmbedded((prevSet) => {
+        const newSet = new Set(prevSet);
+        newSet.add(eventData.id);
+        return newSet;
+      });
     }
   };
 
--- a/src/UI/Components/Diagram/Canvas.tsx
+++ b/src/UI/Components/Diagram/Canvas.tsx
@@ -35,6 +35,7 @@ const Canvas = ({
   const environment = environmentHandler.useId();
   const baseUrl = urlManager.getApiUrl();
   const canvas = useRef<HTMLDivElement>(null);
+  const [looseEmbedded, setLooseEmbedded] = useState<Set<string>>(new Set());
   const [alertMessage, setAlertMessage] = useState("");
   const [alertType, setAlertType] = useState(AlertVariant.danger);
   const [isFormModalOpen, setIsFormModalOpen] = useState(false);
@@ -53,6 +54,21 @@ const Canvas = ({
   const url = routeManager.useUrl("Inventory", {
     service: mainServiceName,
   });
+
+  const handleLooseEmbeddedEvent = (event) => {
+    const customEvent = event as CustomEvent;
+    const eventData: { kind: "remove" | "add"; id: string } = JSON.parse(
+      customEvent.detail,
+    );
+    if (eventData.kind === "remove") {
+      setLooseEmbedded((prevSet) => {
+        prevSet.delete(eventData.id);
+        return new Set(prevSet);
+      });
+    } else {
+      setLooseEmbedded((prevSet) => new Set(prevSet.add(eventData.id)));
+    }
+  };
 
   const handleDictEvent = (event) => {
     const customEvent = event as CustomEvent;
@@ -194,10 +210,12 @@ const Canvas = ({
   useEffect(() => {
     document.addEventListener("openDictsModal", handleDictEvent);
     document.addEventListener("openEditModal", handleEditEvent);
+    document.addEventListener("looseEmbedded", handleLooseEmbeddedEvent);
 
     return () => {
       document.removeEventListener("openDictsModal", handleDictEvent);
       document.removeEventListener("openEditModal", handleEditEvent);
+      document.addEventListener("looseEmbedded", handleLooseEmbeddedEvent);
     };
   }, []);
 
@@ -262,7 +280,9 @@ const Canvas = ({
         }}
         serviceName={mainServiceName}
         handleDeploy={handleDeploy}
-        isDeployDisabled={instancesToSend.size < 1 || !isDirty}
+        isDeployDisabled={
+          instancesToSend.size < 1 || !isDirty || looseEmbedded.size > 0
+        }
       />
       <CanvasWrapper id="canvas-wrapper">
         <div className="canvas" ref={canvas} />
--- a/src/UI/Components/Diagram/halo.ts
+++ b/src/UI/Components/Diagram/halo.ts
@@ -1,5 +1,5 @@
 import { dia, highlighters, ui } from "@inmanta/rappid";
-import { checkIfConnectionIsAllowed } from "./helpers";
+import { checkIfConnectionIsAllowed, toggleLooseElement } from "./helpers";
 import { ActionEnum, ConnectionRules } from "./interfaces";
 import { ServiceEntityBlock } from "./shapes";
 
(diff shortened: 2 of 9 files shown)


src/UI/Components/Diagram/halo.ts
```diff
@@ -123,7 +130,7 @@ const createHalo = (
     const shapes = paper.findViewsInArea(area);
 
     shapes.map((shape) => {
-      highlighters.mask.remove(shape), "available-to-connect";
+      dia.HighlighterView.get(shape, "available-to-connect")?.remove();
```
LukasStordeur, line 133: I'm not a fan of optional chaining to call methods on the potentially undefined/null variable. In this case, it might return an undefined (perhaps that's what you need), and also fail silently. Which can make it hard to debug at some point.

src/UI/Components/Diagram/helpers.ts
```diff
@@ -440,3 +440,39 @@ export const findCorrespondingId = (
...
+ * Toggle the highlighting of a loose element in a diagram cell view.
+ * @param {dia.CellView} cellView - The cell view containing the element.
+ * @param {"add" | "remove"} kind - The action to perform, either "add" to add highlighting or "remove" to remove highlighting.
+ * @returns {void}
+ */
+export const toggleLooseElement = (
+  cellView: dia.CellView,
+  kind: "add" | "remove",
+): void => {
+  switch (kind) {
```
LukasStordeur, line 454: when having a switch case, it is always safe practice to have a default case

src/UI/Components/Diagram/helpers.ts
```diff
@@ -440,3 +440,39 @@ export const findCorrespondingId = (
...
+        className: "loose_element-highlight",
+        attrs: {
+          "stroke-opacity": 0.5,
+          "stroke-width": 5,
+          filter: "drop-shadow(3px 5px 2px rgb(0 0 0 / 0.4))",
+        },
+      });
+      break;
+    case "remove":
+      dia.HighlighterView.get(cellView, "loose_element")?.remove();
```
LukasStordeur, line 467: same remark as higher up, this might end up in failing silently and not calling the method. Is this a behavior that you wanted?

src/UI/Components/Diagram/styles.ts
```diff
@@ -102,6 +102,10 @@ export const CanvasWrapper = styled.div`
     stroke: var(--pf-v5-global--palette--light-green-300);
     fill: var(--pf-v5-global--palette--light-green-300);
   }
+  .joint-loose_element-highlight {
+    stroke: var(--pf-v5-global--palette--red-100);
```
LukasStordeur, line 106: you picked the stronger ones for the green, perhaps that's why it appears to look saturated or muddy for the red.

LukasStordeur:
About the colors, when it has both states, available and errored, it should not display both (gives a dirty green halo). 
Also, the red doesn't look really good imo.

src/UI/Components/Diagram/halo.ts
matborowczyk: I'll change it, although, here it is possible undefined/null, but that mean there is no highlighter given highlighter which is fine, as we are iterating through all of instances that are on the canvas.

src/UI/Components/Diagram/helpers.ts
matborowczyk: Similarly to comment above, theoretically it's possible to get undefined/null here, if there won't be highlighterView that would mean there is no red border(in this case, green in other place). Which here is undesired but that would meant that it was renamed in the line above.

src/UI/Components/Diagram/styles.ts
matborowczyk: changed and made border smaller

src/UI/Components/Diagram/helpers.ts
matborowczyk: Here it's either of those, with default to no action, which would be the case here with, or without default. I don't know if my thought process is correct here.

src/UI/Components/Diagram/helpers.ts
matborowczyk: I don't know why i chose switch over if in that case though, I can switch(no pun intended) it in that case(no pun intended as well) :)

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -54,6 +55,21 @@ const Canvas = ({
     service: mainServiceName,
   });
 
+  const handleLooseEmbeddedEvent = (event) => {
+    const customEvent = event as CustomEvent;
```
LukasStordeur, line 59: can't you type the argument instead?

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -54,6 +55,21 @@ const Canvas = ({
...
+    const eventData: { kind: "remove" | "add"; id: string } = JSON.parse(
+      customEvent.detail,
+    );
+    if (eventData.kind === "remove") {
+      setLooseEmbedded((prevSet) => {
```
LukasStordeur, line 64: you are returning a new Set to avoid mutating the previous state, but by doing the delete action, you are mutating. 
```
setLooseEmbedded((prevSet) => {
  const newSet = new Set(prevSet);
  newSet.delete(eventData.id);
  return newSet;
});
```
You can do it this way to avoid mutating the previous state.

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -54,6 +55,21 @@ const Canvas = ({
...
+        prevSet.delete(eventData.id);
+        return new Set(prevSet);
+      });
+    } else {
+      setLooseEmbedded((prevSet) => new Set(prevSet.add(eventData.id)));
```
LukasStordeur, line 69: same here as above. It's always best to prefer immutability principles to prevent side effects on the long run. Mutating the data should be done deliberately and for specific reasons, which is not the case here I believe.

src/UI/Components/Diagram/halo.ts
```diff
@@ -113,6 +120,14 @@ const createHalo = (
             filter: "drop-shadow(3px 5px 2px rgb(0 0 0 / 0.4))",
           },
         });
+        const looseElementHighlight = dia.HighlighterView.get(
+          element,
+          "loose_element",
+        );
+
+        if (looseElementHighlight) {
+          V(looseElementHighlight.el).addClass("-hidden");
```
LukasStordeur, line 129: What is V ? I see the import, but what does it do?

src/UI/Components/Diagram/helpers.test.ts
```diff
@@ -1349,3 +1350,51 @@ describe("bundleInstances", () => {
     expect(bundledInstances).toEqual([coreCopy]);
   });
 });
+
+describe("toggleLooseElement", () => {
+  it("dispatch an proper event with id when called", () => {
```
LukasStordeur, line 1355:
```suggestion
  it("dispatch a proper event with id when called", () => {
```

src/UI/Components/Diagram/helpers.test.ts
```diff
@@ -1349,3 +1350,51 @@ describe("bundleInstances", () => {
...
+    toggleLooseElement(paper.findViewByModel(entity), "remove");
+    expect(
+      dia.HighlighterView.get(paper.findViewByModel(entity), "loose_element"),
+    ).toBeNull();
+    expect((dispatchEventSpy.mock.calls[1][0] as CustomEvent).detail).toEqual(
+      JSON.stringify({ kind: "remove", id: entity.id }),
+    );
+  });
+
+  it("appends and removes a highlighter object onto entity", () => {
```
LukasStordeur, line 1383:
```suggestion
  it("appends and removes a highlighted object from an entity", () => {
```

src/UI/Components/Diagram/helpers.ts
LukasStordeur: yes, but the default case can also simply do a break. I think, and that was also a rule at my previous company; that having the default case statement present in the switch case is best practice. The typing says it should never get there, but typing is not bulletproof.  

Adding a default case is a form of defensive programming. It helps catch unexpected or unforeseen scenarios. Even if, based on the current understanding of the code, it seems like the switch statement should cover all possible cases, future changes, refactoring, or external factors might introduce new cases.

it can just do a break, or throw an error. Whichever you choose.

src/UI/Components/Diagram/Canvas.tsx
matborowczyk: unfortunately AddEventListener accept function where its argument it of type Event, which is missing property "detail" that stores, tried to override type in the listener itself but it throws out error, even though CustomEvent extends Event interface

src/UI/Components/Diagram/halo.ts
matborowczyk: it's tool from jointjs library, [helper for svgs ](https://resources.jointjs.com/docs/jointjs/v4.0/vectorizer.html#V.prototype.addClass)

I remove that, as later on labels branch reminded myself about classList.add/remove (when tried to use it first I saw that it's read-only and switched for that without trying these functions out).

src/UI/Components/Diagram/Canvas.tsx
LukasStordeur: right... forgot Typescript was sometimes complicated.

LukasStordeur (approved):

Issue #5531, opened 2024-02-15
Support suggested values in inventory fields


bartv:
The goal of this ticket is to implemented suggested values based on information added to the service catalog. The frontend base model now has a number of examples in the basic-service. This can be used both in text fields, select fields and lists. There are two versions for this

1. With literal values. In the example project you will find this value:
    ```json
    {
        "attribute_annotations": {
            "web_suggested_values": {
                "type": "literal",
                "values": [
                    "192.168.10.1/24",
                    "192.168.10.2/24",
                    "192.168.10.3/24"
                ]
            }
        },
        "default_value": null,
        "default_value_set": false,
        "description": "The IP-address/netmask to assign to the given VLAN interface.",
        "modifier": "rw+",
        "name": "address_r1",
        "type": "string",
        "validation_parameters": null,
        "validation_type": "pydantic.IPvAnyInterface"
    }
    ```

2. A reference to a parameter value on the api:
   ```json
    {
        "attribute_annotations": {
            "web_suggested_values": {
                "parameter_name": "web-address_r2",
                "type": "parameters"
            }
        },
        "default_value": null,
        "default_value_set": false,
        "description": "The IP-address/netmask to assign to the given VLAN interface.",
        "modifier": "rw+",
        "name": "address_r2",
        "type": "string",
        "validation_parameters": null,
        "validation_type": "pydantic.IPvAnyInterface"
    }
   ```

In the parameter API you have the following two examples (both attached to a field like shown a bove). The data is always stored in metadata because only that field accepts json:

1. With only the values in a list:
    ```json
    {
        "value": "metadata",
        "source": "user",
        "metadata": {
            "values": [
                "192.168.10.1/24",
                "192.168.10.2/24",
                "192.168.10.3/24"
            ]
        }
    }
    ```

2. With values and labels:
   ```json
    {
        "value": "metadata",
        "source": "user",
        "metadata": {
            "values": [
                {"label": "ip1", "value": "192.168.10.1/24"},
                {"label": "ip2", "value": "192.168.10.2/24"},
                {"label": "ip3", "value": "192.168.10.3/24"}
            ]
        }
    }
   ```

Option 2 is a bit of a tricky one because it makes sense for select boxes but probably less for normal fields and list. An option is to use the values anyway when fields do not support labels.

PR #5581, opened 2024-03-05
5531 add support for form-suggestions

LukasStordeur:
# Description

The ServiceInstanceForm now handles suggested values, literal ones and parameters. The suggestions are both available in the Composer and Normal Form. 

MSW has been added to the project to handle mocking API requests. [documentation](https://mswjs.io/docs/api/http-response)
React-Queries are now handled in a new folder labeled V2. The idea is to use this location for all queries we will add/edit. 

There was also a bug resolution for the InlineInput Element, by adding typing, it turned out the event and value for that onChange event were inversed, causing the app to break when you were in expert mode and trying to update a dictList using a TextListInput.

closes *#5531*

https://github.com/inmanta/web-console/assets/44098050/3da1706f-3e39-4b7d-9645-50874757b498

https://github.com/inmanta/web-console/assets/44098050/2691d757-c569-494b-a9c2-65662ecae5b4


src/Data/Managers/V2/FormSuggestions/useSuggestions.ts
```diff
@@ -0,0 +1,80 @@
...
+        };
+      },
+    };
+  }
+
+  const baseUrlManager = new PrimaryBaseUrlManager(
+    globalThis.location.origin,
+    globalThis.location.pathname,
+  );
+  const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
```
matborowczyk, line 45: I think we could reuse urlManager here
`const { urlManager } = useContext(DependencyContext);`
`const baseUrl = urlManager.getApiUrl();`

src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx
```diff
@@ -79,6 +95,26 @@ export const FieldInput: React.FC<Props> = ({
...
+    if (!isLoading && !error) {
+      // if the data is of type array, we can use it directly
+      if (Array.isArray(data)) {
+        setSuggestionsList(data);
+      } else if (
+        data &&
+        data.metadata &&
+        data.metadata.values &&
+        Array.isArray(data.metadata.values) &&
+        // TODO: remove this when the API returns a fixed format.
```
matborowczyk, line 108: Should we assign to to its own ticket?

src/Data/Managers/V2/FormSuggestions/useSuggestions.ts
LukasStordeur: I wanted to avoid having the Dependency context here to keep it more simple. Adding the context would mean we need to inject it in the tests too, but if you think it is best to use it, I can change it.

src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx
LukasStordeur: I'll make a ticket, yes

LukasStordeur: https://github.com/inmanta/web-console/issues/5582

src/Data/Managers/V2/FormSuggestions/useSuggestions.ts
matborowczyk: That's a fair point, I thought about having not to repeat that in every query. I don't have strong preference though, It's a trade-off and we can stand on that site, especially if we gonna remove old queries in some time in future, there is a high chance that whole `urlManager` will be redundant.

LukasStordeur: it's two or three lines, I think for now, having it dependency-less is best, we can revisit this later. Sadly, react-query doesn't have a way to set it in the queryClient configuration, which would have been handy.

src/UI/Components/ServiceInstanceForm/Components/SuggestionsPopover.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+  useEffect(() => {
+    window.addEventListener("keydown", handleMenuKeys);
+    window.addEventListener("click", handleClickOutside);
+    return () => {
+      window.removeEventListener("keydown", handleMenuKeys);
+      window.removeEventListener("click", handleClickOutside);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isOpen]);
```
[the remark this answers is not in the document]
LukasStordeur: hmm, is it? The events only need to be added if the component is shown, but I can try without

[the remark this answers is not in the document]
matborowczyk: now there is no dependency array, I can't visualize the whole flow, it could be fine, but it could run on every change.

I had empty array in mind. but if that works then it's fine :D

matborowczyk (approved):

src/UI/Components/ServiceInstanceForm/Components/SuggestionsPopover.tsx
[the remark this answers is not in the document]
LukasStordeur: true, hmm, it does work though, but I'll add the array with the handlers as dependency.

PR #5599, opened 2024-03-12
Add database based login form

### LINKED ISSUE #5558: Add database based login form
For the local user setup we introduced [a new setting](https://github.com/inmanta/inmanta-ui/pull/536) in the auth dict in config.js For the keycloak/oidc based auth the following dict will be returned:

```js
window.auth = {
    'method': 'oidc',
    'realm': 'inmanta',
    'url': 'http://127.0.0.1:8080',
    'clientId': 'inmantaso'
};
```

When local auth is used, we keep the same bearer based mechanism but the authz will need to happen local. The config.js will be:

```js
  window.auth = {
      'method': 'database'
  };
```

The logic that needs to be applied is the following:

- When window.auth is set and the method is oidc, use the current keycloak method. If method is database the webconsole needs to use it's own logic
- If the backend returns 401, show the login page. [For example this one](https://www.patternfly.org/components/login-page/#loginpage)
- It should use the /api/v2/login 
  image
- This call will return a valid token that has to be used in the authentication header like "Bearer $token"
- The frontend needs to ensure that the bearer token is always passed to the backend
- If the token expires, it will generate a 401 and the login page should be shown again
- Logout means no longer setting the authentication header

Currently, the username is not encoded in the token. This will be fixed in https://github.com/inmanta/inmanta-core/pull/7244 When this is merged, the token will contain a "sub" claim (key) that contains the username for in the top bar.

The token can be decoded with for example https://github.com/auth0/jwt-decode

The local-setup has now a start:localauth target. This will setup the orchestrator with ssl and local auth. The initial user is admin with password adminadmin

matborowczyk:
# Description

closes #5558 

# Self Check:

- [ ] Changelog entry

https://github.com/inmanta/web-console/assets/113331659/1987f951-62b4-443b-b209-d8fbf270c972


LukasStordeur:
can we show another message in case the user is unauthenticated? Like, a please wait you are being redirected? Since we know the error message before it hits the screen, I think it might be worth the effort to look further into it.

src/Data/API/BaseApiHelper.ts
```diff
@@ -9,12 +9,14 @@ import {
...
 } from "@/Core";
 import { words } from "@/UI/words";
+import { getCookie } from "../Common/CookieHelper";
 import { BigIntJsonParser } from "./BigIntJsonParser";
 
 export class BaseApiHelper implements ApiHelper {
   jsonParser = new BigIntJsonParser();
   constructor(
     private readonly baseUrl: string = "",
+    private readonly shouldAuthLocally: boolean = false,
```
LukasStordeur, line 19: why are you passing the shouldAuthLocally? I can't see it being used in the code below.

src/UI/Components/Diagram/Canvas.test.tsx
```diff
@@ -413,7 +413,10 @@ describe("Canvas.tsx", () => {
       await user.click(copyButton);
     });
 
-    const clipboardText = await navigator.clipboard.readText();
+    const clipboardItems = await navigator.clipboard.read();
```
LukasStordeur, line 416: what does this change relates to?

src/UI/Components/ErrorView/ErrorView.tsx
```diff
@@ -25,27 +28,54 @@ export const ErrorView: React.FC<Props> = ({
...
-          <Button variant="primary" onClick={retry}>
-            {words("retry")}
-          </Button>
-        )}
-      </EmptyStateFooter>
-    </EmptyState>
-  </Delayed>
-);
+}) => {
+  const { authController } = useContext(DependencyContext);
```
LukasStordeur, line 32: Why here? And not in the baseApiHelper? Now you are adding an additional dependency into this file, making it high cohesion instead of the opposite. 

I understand better why you have a glitch now.
The useEffect will run after the initial mount of the component, and then update accordingly. This takes a fragment of a second, but enough time to actually display the error view instead of redirecting. 

The redirection should be done as high level as possible, right when the response lands into the code.

src/UI/Root/Components/Header/Actions/LocalProfie/Profile.tsx
LukasStordeur, line 1: There's a typo in the name of the folder. You wrote LocalProfie, instead of LocalProfile.

src/Injector.tsx
```diff
@@ -91,11 +91,12 @@ export const Injector: React.FC<React.PropsWithChildren<Props>> = ({
         environmentHandler,
         authHelper,
         archiveHelper,
-        keycloakController,
+        authController,
       }}
     >
       <UpdateBanner apiHelper={apiHelper} />
       {children}
+      <ReactQueryDevtools initialIsOpen={false} />
```
LukasStordeur, line 99: Why did you add it a second time? I believe we only need it on the root component, no?

src/UI/Root/Root.tsx
```diff
@@ -34,39 +35,49 @@ export const Root: React.FC = () => {
...
-          </SearchSanitizer.Provider>
-        </Initializer>
+        <SearchSanitizer.Provider>
+          <Routes>
+            <Route path="console/login" element={<LoginPage />} />
+            <Route
+              path="*"
+              element={
+                <Initializer>
+                  <Routes>
```
LukasStordeur, line 45: Is this on purpose? To have two levels or  Routes nested? I find it confusing.

src/UI/Utils/ErrorBoundary.test.tsx
```diff
@@ -12,9 +13,11 @@ test("GIVEN ErrorBoundary WHEN no error is thrown THEN error page should NOT be
   const explode = false;
 
   const component = (
-    <ErrorBoundary>
-      <span data-testid="bomb">{explode ? <Bomb /> : "normal text"}</span>
-    </ErrorBoundary>
+    <MemoryRouter>
```
LukasStordeur, line 16: The ErrorBoundary should by design be encapsulating the entire application. Having it wrapped in a MemoryRouter defeats the purpose of the element, even if it is in a test case. You shouldn't have anything around it.

LukasStordeur (commented):
You mixed both authentication ways into a single controler, I find the approach rather confusing. For the sake of separating concerns, it might have been more adequate to have two separate handlers that do not mingle. It makes it easier to maintain/replace/adjust.

The auth object is present on the window, you should be able to access it as soon as the app is hitting the browser. (This might mean you need to extend the typing because of Typescript) But this means you know at the root level, when you are initializing your dependencies if you are going to need a localAuth, or attempt the keycloak one. 

There are very few comments in the code too, it would be nice to have more of them, at least on methods and components declaration. 

Lastly, I noticed you opted for the cookies instead of the sessionStorage, can you stave your choice? Both are valid, I'd just like to know your reasoning, and then we can make up and adr for it.

jest.polyfills.js
```diff
@@ -6,11 +6,24 @@
...
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable import/order */
 const { TextDecoder, TextEncoder, ReadableStream } = require("node:util");
 
 Object.defineProperties(globalThis, {
   TextDecoder: { value: TextDecoder },
   TextEncoder: { value: TextEncoder },
   ReadableStream: { value: ReadableStream },
 });
+const { Blob, File } = require("node:buffer");
```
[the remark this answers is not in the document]
matborowczyk: Withou these polyfills mutations wasn't working properly in jest, and I spent hours to find out why, and finally I found that this is hard requirement for msw v2

src/Injector.tsx
matborowczyk: Forgot to remove that, after merging. removed

src/Data/Auth/PrimaryAuthController.ts
```diff
@@ -26,11 +32,25 @@ export class PrimaryKeycloakController implements KeycloakController {
   }
 
   private getConfig(): KeycloakConfig {
-    return (
-      this.externalConfig || {
+    if (this.externalConfig && this.externalConfig?.method === "oidc") {
```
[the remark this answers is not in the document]
matborowczyk: previously we were returning either keycloak config provided by config.js file or this default one, now config provided can vary as it can also mention database, to keep login the same way I had to add this additional check

matborowczyk:
> can we show another message in case the user is unauthenticated? Like, a please wait you are being redirected? Since we know the error message before it hits the screen, I think it might be worth the effort to look further into it.

It's being shown "Redirecting to log in page shortly..." (0:03)

src/UI/Utils/ErrorBoundary.test.tsx
matborowczyk: We are using navigate in the errorBoundary to redirect, putting Router inside cause error, so only thing I can do is to change into `window.location.replace()`

src/UI/Root/Components/Header/Actions/LocalProfie/Profile.tsx
matborowczyk: It had to be there already, fixed

src/UI/Components/Diagram/Canvas.test.tsx
matborowczyk: polyfills, it broke that part of the text

src/Data/API/BaseApiHelper.ts
```diff
@@ -33,26 +35,32 @@ export class BaseApiHelper implements ApiHelper {
...
-        : {}),
+      ...(Authorization ? { Authorization } : {}),
     };
   }
 
   private formatError(message: string, response: Response): string {
     let errorMessage = message;
-
-    if (this.keycloak && (response.status === 401 || response.status === 403)) {
+    if (response.status === 401 || response.status === 403) {
```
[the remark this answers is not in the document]
matborowczyk: the glitch was caused by Initializer being in wrong place. I can move the logic here, I guess.

src/Data/API/BaseApiHelper.ts
```diff
@@ -33,26 +35,32 @@ export class BaseApiHelper implements ApiHelper {
 
   private getHeaders(environment?: string): Record<string, string> {
     const { keycloak } = this;
+    const jwt = getCookie("inmanta_user");
```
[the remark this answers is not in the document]
matborowczyk: Ad1) This cookie set only when user use database login, now as I think about it ,the only question is, whether it is possible that end-user toggle between authorisations? @bartv 

Ad2) I'll extract it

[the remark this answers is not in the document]
bartv: It is not possible to switch. However, just like with keycloak we need to always pass along the token.

matborowczyk:
> You mixed both authentication ways into a single controler, I find the approach rather confusing. For the sake of separating concerns, it might have been more adequate to have two separate handlers that do not mingle. It makes it easier to maintain/replace/adjust.

I personally think that is fine to have them both together, as both things are responsible for authentication and if auth is used it's either one or the other.

> The auth object is present on the window, you should be able to access it as soon as the app is hitting the browser. (This might mean you need to extend the typing because of Typescript) But this means you know at the root level, when you are initializing your dependencies if you are going to need a localAuth, or attempt the keycloak one.

Isn't that handled currently in AuthController?

> There are very few comments in the code too, it would be nice to have more of them, at least on methods and components declaration.

Could you point parts that aren't clear? I added docstrings to the functions but I think that rest is self-explanatory, but I can be mistaken

> I also don't see any E2E tests. Either make a ticket, or add them :)

Forgot about, I'll either create separate ticket or wait with merge and do it after I resolve notification bug.


> Lastly, I noticed you opted for the cookies instead of the sessionStorage, can you stave your choice? Both are valid, I'd just like to know your reasoning, and then we can make up an adr for it.

as far as I know cookies are safer to store jwt that sessionStorage

jest.polyfills.js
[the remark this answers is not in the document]
LukasStordeur: okay, for me they were breaking a lot of things, but if you got it working, great!

src/Data/API/BaseApiHelper.ts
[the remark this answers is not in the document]
LukasStordeur: I'd still appreciate if it was extracted. I think it makes it clearer and even if theoretically it would never end up in the loophole, it's always best to avoid creating them when we can.

src/Data/API/BaseApiHelper.ts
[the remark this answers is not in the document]
LukasStordeur: but this is the closest level you could possibly handle the error, so it would make sense to actually handle the responses here, not in the error-view.

src/UI/Utils/ErrorBoundary.test.tsx
LukasStordeur: The ErrorBoundary is for unhandled errors. It gets displayed only in those cases, and shouldn't be responsible for redirecting the user. To me, it means the logic has flaws if you end up here. This isn't resolving the issue, but hacking around.

src/Data/API/BaseApiHelper.ts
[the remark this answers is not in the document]
matborowczyk: extracted

Issue #5532, opened 2024-02-15
Introduce web_presentation annotation for documentation


bartv:
The goal of this feature is to allow the model developer to provide information to the user/operator in the web console. The current proposal is to use the following in the model:

```
    string documentation
    lsm::attribute_modifier documentation__modifier="r"
    dict documentation__annotations = {
        "web_presentation": "documentation",
        "web_title": "Documentation",
        "web_icon": "fa-info-circle",
        "web_content_type": "text/markdown"
    }
```

<details>
<summary>The json equivalent</summary>

```json
{
    "attribute_annotations": {
        "web_content_type": "text/markdown",
        "web_icon": "fa-info-circle",
        "web_presentation": "documentation",
        "web_title": "Documentation"
    },
    "default_value": null,
    "default_value_set": false,
    "description": null,
    "modifier": "r",
    "name": "documentation",
    "type": "string",
    "validation_parameters": null,
    "validation_type": null
}
```

</details

This results in an additional tab in expanded row (before the others).

![image](https://github.com/inmanta/web-console/assets/10819/cbed641d-40e4-4b73-8115-2a8bc3bb6335)

The support content types could/should be:

- text/plain
- text/html (do we want to do this??)
- text/markdown

For text/markdown we also want to support the mermaid extension so that we can include diagrams.

Discussion: do we use "tab" for the web_presentation annotation?

This also means we need to be able to render markdown in the browser. [markdown-it ](https://github.com/markdown-it/markdown-it)(also used in the python world) seems to be one of the most complete and most used implementations:

For mermaid this is mermaid-js and there seems to be a number of plugins available. The original implementation of the plugin seems to be no longer maintained. However this fork seems to be more active: https://github.com/wekan/markdown-it-mermaid#readme

The following is some example code that you can use to test (without the backslashes):

```
# This is some more information

\```mermaid
  graph TD;
      A-->B;
      A-->C;
      B-->D;
      C-->D;
\```
```

PR #5613, opened 2024-03-19
5532 markdown presentation

LukasStordeur:
# Description

closes : #5532 

The attribute annotations are now supported by the Service Inventory table.
When an attribute has the correct type of annotation, it will be rendered as a link button in the attributeTreeTable and upon being clicked on, it will open the matching tab containing the documentation. 

Mermaid has been blocked from major updates and locked onto version 9. 
The structuredClone method is not yet supported by JsDom and Jest. Adding the needed mocks would be a lot of work and counter productive. Version 9 of mermaid will still have all the needed security updates for now. 

https://github.com/inmanta/web-console/assets/44098050/e639d996-9bd0-4fb7-b4c5-86404ef6b55d

https://github.com/inmanta/web-console/assets/44098050/5fe70936-7707-47f4-a314-71934c6f36af


src/Slices/ServiceInventory/UI/Tabs/Tabs.tsx
```diff
@@ -24,14 +29,27 @@ export enum TabKey {
 }
 
 interface Props {
-  activeTab: TabKey;
-  setActiveTab: (tabKey: TabKey) => void;
+  activeTab: TabKey | string;
```
matborowczyk, line 32: I think it would be helpful to add info why we escape enum into general type, at first I was confused but then I checked into DocumentationTab and realised that it's probably due to the fact that we can't foresee every possibility there. But to realise that I had to look into another file

src/UI/Components/TreeTable/TreeTable.test.tsx
```diff
@@ -73,6 +79,65 @@ test("TreeTable 1st level of nested property can be expanded", async () => {
...
+  const documentationButton = screen.getByRole("button", {
+    name: "documentation",
+  });
+  expect(documentationButton).toBeInTheDocument();
+
+  act(() => {
+    fireEvent.click(documentationButton);
+  });
+  // expect the setTab from the TreeTable component method to have been called
+  expect(setTab).toHaveBeenCalled();
```
matborowczyk, line 138: ui for the tabs is there, so assertion that the tab content will be visible should also work, right? If that's the case then it would be better to assert that instead just the call, but I can be mistaken as there are plenty of components, helpers, creators for instance services view

src/UI/Components/TreeTable/TreeRow/TreeRowView.tsx
```diff
@@ -1,74 +1,116 @@
...
     case "Flat":
       return (
         <StyledTr aria-label={`Row-${row.id}`}>
           <Td dataLabel={row.primaryCell.label}>
             <Indent level={0} noToggle>
-              {row.primaryCell.value}
+              {annotations?.web_presentation !== "documentation" ? (
+                row.primaryCell.value
+              ) : (
+                <DocumentationCell
```
matborowczyk, line 71: I think that the gain for this kind of components is minuscule. it doesn't save that much lines of code there, there is nothing extra happening inside of the component itself, and it's not reused anywhere else.

That doesn't mean that I want to remove it, just sharing my opinion

src/Slices/ServiceInventory/UI/Tabs/Tabs.tsx
```diff
@@ -128,3 +178,80 @@ const configTab = (
...
 });
+
+/**
+ * Creates an array of documentation tab descriptors.
+ *
+ * @param row - The row object.
+ * @param service - The service model.
+ * @returns An array of documentation tab descriptors.
+ */
+const documentationTab = (
```
matborowczyk, line 189: cosmetic things, but it should be documentationTabs

src/UI/Components/TreeTable/TreeRow/TreeRowView.tsx
LukasStordeur: I separated it because when I see ternaries with big blobs of code, I lose sight of what I'm reading. Naming the component also adds readability.

src/UI/Components/TreeTable/TreeTable.test.tsx
LukasStordeur: Actually, the UI for the tabs is in a higher level. This TreeTable is inside a tab and only shows the attribute table with the collapsible/expandable attributes. Hence why the check if the setTab method is being called instead :)

src/Slices/ServiceInventory/UI/Tabs/Tabs.tsx
LukasStordeur: added in the docstrings

src/Slices/ServiceInventory/UI/Tabs/Tabs.tsx
LukasStordeur: updated

matborowczyk (approved):

PR #5626, opened 2024-03-22
Add feature flags for order view, resource discovery, and composer***

LukasStordeur:
# Description

Update tests and dependencies for new feature flags
Update navigation and page components to use feature flags
Update instance composer and orders pages to check feature flag before rendering

closes: #5619


src/Data/Common/PrimaryFeatureManager.ts
```diff
@@ -46,34 +66,92 @@ export class PrimaryFeatureManager implements FeatureManager {
     return this.getExtensions().includes(extension);
   }
 
+  private isFeaterEnabled(featureSlice: string): boolean {
```
matborowczyk, line 69: Feature?

matborowczyk: Also, docstrings are missing, as they are added everywhere else in the class

matborowczyk (approved):

PR #5632, opened 2024-03-27
5554 improve add button composer

### LINKED ISSUE #5554: [Composer] replace Add Instance button with Action button
https://www.patternfly.org/components/menus/menu-toggle#split-button-toggle-with-action

Replace the Add Instance button when more than one option is possible with a Split button toggle. 
Possible cases will be Add in Composer, Add Instance, and later Add with Json Editor.

The default should be the form.

LukasStordeur:
# Description

close #5554 

- Updated the button to add instances on the Service Inventory view. If the composer is enabled, it will now appear as a action toggle.
- Updated the custom Link component that use React-Router-Dom to allow to hide the outline when we use it in menu-items or inside buttons. It now supports the variant attribute which will default to "default".
- Updated which attribute state the documentation is prioritizing to display the markdown. Candidate > Active > Rollback.

https://github.com/inmanta/web-console/assets/44098050/6b4a56cc-61b2-4b1f-8fa3-db2295fe18ee

### DIFF
--- a/src/Slices/ServiceInventory/UI/Tabs/Tabs.tsx
+++ b/src/Slices/ServiceInventory/UI/Tabs/Tabs.tsx
@@ -235,17 +235,17 @@ const documentationTab = (
 const getAttributeValue = (attributeName: string, row: Row) => {
   if (
     row.attributes &&
-    row.attributes.active &&
-    row.attributes.active[attributeName]
-  ) {
-    return row.attributes.active[attributeName];
-  }
-  if (
-    row.attributes &&
     row.attributes.candidate &&
     row.attributes.candidate[attributeName]
   ) {
     return row.attributes.candidate[attributeName];
+  }
+  if (
+    row.attributes &&
+    row.attributes.active &&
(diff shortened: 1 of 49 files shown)


src/Slices/ServiceInventory/UI/Tabs/Tabs.tsx
```diff
@@ -235,17 +235,17 @@ const documentationTab = (
 const getAttributeValue = (attributeName: string, row: Row) => {
   if (
     row.attributes &&
-    row.attributes.active &&
```
matborowczyk, line 238: I just want to know why the order is reversed? Not saying that is wrong but wonder, should candidate be default displayed value here?

matborowczyk (approved):

src/Slices/ServiceInventory/UI/Tabs/Tabs.tsx
LukasStordeur: Bart requested it :D

PR #5639, opened 2024-03-27
Refactor SingleTextSelect and InstanceCellButton components

### LINKED ISSUE #5561: [Bug] filtering for inter-service-relation links
https://inmanta.slack.com/archives/C045Y42N49W/p1708685908082859

Can be reproduced in the local setup by creating two instances, one child, one parent, and then clicking on the link in the attributes of the child-service forwarding to the filtered view of the parent-services

LukasStordeur:
# Description

closes #5561 

Repair the forwarding to a filtered vue on the service inventory table
Repair the displayName for the create Instance form for Single Selects. (the problem didn't occur in the multi-select)

https://github.com/inmanta/web-console/assets/44098050/eab2b64b-8e14-44cd-b673-3d6e8f9e94fb

https://github.com/inmanta/web-console/assets/44098050/a7d7aaf7-2d52-4fc3-b821-0ed14c5d80cd

### DIFF
--- a/src/UI/Components/SingleTextSelect/SingleTextSelect.tsx
+++ b/src/UI/Components/SingleTextSelect/SingleTextSelect.tsx
@@ -119,17 +119,22 @@ export const SingleTextSelect: React.FC<Props> = ({
   }, [options]);
 
   useEffect(() => {
-    setInputValue(selected || "");
+    setInputValue(getDisplayValue(selected));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [selected]);
 
   const onToggleClick = () => {
     setIsOpen(!isOpen);
   };
 
-  const getDisplayValue = (value: string | number | undefined) => {
+  const getDisplayValue = (value: string | null) => {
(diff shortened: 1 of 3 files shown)


src/UI/Components/SingleTextSelect/SingleTextSelect.tsx
```diff
@@ -145,7 +150,7 @@ export const SingleTextSelect: React.FC<Props> = ({
       default:
         if (value && value !== "no results") {
           setFilterValue("");
-          setInputValue(getDisplayValue(value) as string);
+          setInputValue(getDisplayValue(value as string));
```
matborowczyk, line 153: Instead of casting it as a string we could  use `String(value)`. Light suggestions as I'm fine with both ways.

matborowczyk (approved):

src/UI/Components/SingleTextSelect/SingleTextSelect.tsx
LukasStordeur: updated in other places in the file too where it was being casted

PR #5641, opened 2024-03-28
Update Canvas and Page components

LukasStordeur:
# Description

Hide the label-toggle (commented out the code to keep it for further use if we need it)
Add titles to the pages
Increase size of the handler icons
Update add-instance button to have text

https://github.com/inmanta/web-console/assets/44098050/c5f744f1-0fab-4783-bf22-21d5db9c595a


src/UI/Components/Diagram/components/Toolbar.tsx
```diff
@@ -56,14 +55,12 @@ const Toolbar = ({
                   src={entityIcon}
                   alt="Create new entity icon"
                   aria-label="new-entity-icon"
-                />
+                />{" "}
```
matborowczyk, line 58:
```suggestion
                />
```

matborowczyk (approved):

Issue #4970, opened 2023-07-14
deleted workaround for related entities query manager


matborowczyk:

matborowczyk:
Current QueryManager for Related Entities  used two calls in recurrence chain.

`Query.SubQuery<"GetInstanceWithRelations">` doesn't have logic to get referenced_by correctly, `Query.SubQuery<"GetServiceInstance">` does, but it has to have service_entity name which is missing when going down to recurrence tree.

When first one will be updated, then we can simplify query to only `Query.SubQuery<"GetInstanceWithRelations">`

matborowczyk:
GET call `lsm/v1/service_inventory` still doesn't return referenced ids for given service instance

PR #5672, opened 2024-04-10
Add persisting coordinates to the backend

### LINKED ISSUE #5273: Composer - Persist position of elements on the canvas

matborowczyk:
# Description

* Short description here *

closes #5273 #4970 

# Self Check:

- [x] Changelog entry

https://github.com/inmanta/web-console/assets/113331659/8180bb06-efa0-4f37-9b6a-0c8af965a480

https://github.com/inmanta/web-console/assets/113331659/7c049e6a-cbc9-47e4-b7e4-27b6b74187fe


src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
```diff
@@ -0,0 +1,82 @@
...
+    const response = await fetch(
+      `${baseUrl}/lsm/v1/service_inventory?service_id=${instanceId}&include_deployment_progress=false&exclude_read_only_attributes=false&include_referenced_by=true&include_metadata=false`,
+      {
+        headers,
+      },
+    );
+    handleAuthorization(response);
+
+    if (!response.ok) {
+      throw new Error("Failed to fetch instance");
```
LukasStordeur, line 45: I'd provide more information in the error-handling. Since this is a recursive, and relies on Promise.all, you have no way to know on what part the code failed.

src/UI/Components/Diagram/Canvas.test.tsx
```diff
@@ -277,7 +283,7 @@ beforeEach(() => {
   });
 });
 
-describe("Canvas.tsx", () => {
+describe.only("Canvas.tsx", () => {
```
LukasStordeur, line 288: I think this is a left over from running your test locally :D

src/UI/Components/Diagram/Canvas.test.tsx
```diff
@@ -603,4 +609,48 @@ describe("Canvas.tsx", () => {
     await deleteAndAssert("child-service", 2, 1);
     await deleteAndAssert("child_container", 1, 0);
   });
+
+  it.only("sends request with correct data to the backend when instance is being deployed", async () => {
```
LukasStordeur: here too

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -55,6 +61,11 @@ const Canvas = ({
     service: mainServiceName,
   });
 
+  /**
+   * Handles the event triggered when there is loose embedded entity on the canvas.
+   *
+   * @param {Event} event - The event object.
```
LukasStordeur, line 69:
```suggestion
   * @param {CustomEvent} event - The event object.
```

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -75,50 +86,45 @@ const Canvas = ({
     }
   };
 
+  /**
+   * Handles the event triggered when user want to see the dictionary properties of an entity.
+   *
+   * @param {Event} event - The event object.
```
LukasStordeur, line 94:
```suggestion
   * @param {CustomEvent} event - The event object.
```

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -75,50 +86,45 @@ const Canvas = ({
...
+   */
   const handleDictEvent = (event) => {
     const customEvent = event as CustomEvent;
     setDictToDisplay(JSON.parse(customEvent.detail));
   };
+
+  /**
+   * Handles the event triggered when user want to edit an entity.
+   *
+   * @param {Event} event - The event object.
```
LukasStordeur, line 104:
```suggestion
   * @param {CustomEvent} event - The event object.
```

src/UI/Components/Diagram/init.ts
```diff
@@ -314,7 +319,13 @@ export default function diagramInit(
       isMainInstance: boolean,
     ) => {
       appendInstance(paper, graph, instance, services, isMainInstance);
-
+      const savedInstancesCoordinates = localStorage.getItem(
```
LukasStordeur, line 366: weren't we going to use the metadata endpoint for this? I believe the BE was ready too. See : [slides planning](https://docs.google.com/presentation/d/15X82_MMk13Qs5R8eUncrAZjeq-BW_yXobZdOaDmUT74/edit#slide=id.g23ae9aa45fc_1_10)

src/UI/Components/Diagram/Canvas.test.tsx
matborowczyk: right, removed :D

src/UI/Components/Diagram/Canvas.test.tsx
matborowczyk: removed

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
```diff
@@ -0,0 +1,82 @@
...
+  );
+  const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
+
+  /**
+   * Fetches a service instance with its related instances recursively.
+   * @param instanceId - The ID of the instance to fetch.
+   * @returns An object containing the fetched instance and its related instances.
+   * @throws Error if the instance fails to fetch.
+   */
+  const fetchService = async (instanceId: string) => {
```
LukasStordeur, line 32: If you make instanceId optional, and fallback on the id passed to the hook, you don't need that extra method on the line 65. 

```suggestion
  const fetchService = async (instanceId: string = id) => {
```

And on line 50 you can keep:  
`const nestedInstance = await fetchService(relatedId);` and it will override the id in the next recursion.

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Custom hook to fetch the parameter from the API once.
+     * @returns The result of the query, including the parameter data.
+     */
+    useOneTime: () =>
+      useQuery({
+        queryKey: ["get_instance_with_relations-one_time", id, environment],
+        queryFn: fetchServiceWithRelations,
+        retry: false,
+        select: (data) => data,
```
LukasStordeur, line 79: data.data?
missing return type too.

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
```diff
@@ -0,0 +1,82 @@
...
+  const fetchServiceWithRelations = async () => {
+    return await fetchService(id);
+  };
+
+  return {
+    /**
+     * Custom hook to fetch the parameter from the API once.
+     * @returns The result of the query, including the parameter data.
+     */
+    useOneTime: () =>
```
LukasStordeur, line 74: typing.
This is how I added the correct typing 
` useOneTime: (): UseQueryResult<ServiceInstanceModel, Error> =>`

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
LukasStordeur, line 32: missing return type

src/Data/Managers/V2/GetMetadata/useGetMetadata.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    return response.json();
+  };
+
+  return {
+    /**
+     * Custom hook to fetch the metadata.
+     * @returns The result of the query.
+     */
+    useOneTime: () =>
```
LukasStordeur, line 57: see comment about the return typing

src/Data/Managers/V2/GetMetadata/useGetMetadata.ts
```diff
@@ -0,0 +1,72 @@
...
+  const baseUrlManager = new PrimaryBaseUrlManager(
+    globalThis.location.origin,
+    globalThis.location.pathname,
+  );
+  const { createHeaders, handleErrors } = useFetchHelpers();
+  const headers = createHeaders(environment);
+  const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
+
+  const getMetadata = async (): Promise<{
+    data: string | undefined;
```
LukasStordeur, line 31: maybe would be good to have a dedicated type for this? Like we have in most cases?

src/Data/Managers/V2/PostMetadata/usePostMetadata.ts
```diff
@@ -0,0 +1,52 @@
...
+  const { createHeaders, handleErrors } = useFetchHelpers();
+  const headers = createHeaders(environment);
+  const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
+
+  /**
+   * Posts metadata.
+   *
+   * @param info - The metadata information to post.
+   */
+  const postMetadata = async (info: {
```
LukasStordeur, line 25: Can we have a type for this? I think it might be handy to be able to reuse this type when you're using the hook.

src/Data/Managers/V2/usePostOrder/usePostOrder.ts
```diff
@@ -0,0 +1,43 @@
+import { useMutation } from "@tanstack/react-query";
+import { PrimaryBaseUrlManager, words } from "@/UI";
+import { InstanceForApi } from "@/UI/Components/Diagram/interfaces";
+import { useFetchHelpers } from "../helpers";
+
+/**
+ * Custom hook for sending an order batch from Instance COmposer.
+ * @returns {Mutation} The mutation object for sending an order.
+ */
+export const usePostOrder = (environment: string) => {
```
LukasStordeur, line 10: You added the return type in the docstrings, can you also add it to the method?

src/UI/Components/InstanceProvider/InstanceProvider.tsx
```diff
@@ -1,34 +1,77 @@
...
-import { DependencyContext } from "@/UI";
-import { RemoteDataView } from "@/UI/Components";
+import { useGetInstanceWithRelations } from "@/Data/Managers/V2/GetInstanceWithRelations";
+import { useGetMetadata } from "@/Data/Managers/V2/GetMetadata";
+import { DependencyContext, words } from "@/UI";
+import { ErrorView, LoadingView } from "@/UI/Components";
 import Canvas from "@/UI/Components/Diagram/Canvas";
 
+/**
+ * Renders the InstanceProvider component.
```
LukasStordeur, line 10: Can you add briefly the purpose of the component/or elaborate.

src/Data/Managers/GetInstanceWithRelations/interface.ts
```diff
@@ -14,8 +14,7 @@ export interface GetInstanceWithRelationsManifest {
 }
 
 export interface InstanceWithReferences {
```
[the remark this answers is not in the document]
matborowczyk: added + moved interface to the v2 as I forgot to delete this unused manager

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
matborowczyk: added

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
matborowczyk: removed that, because we do transformation inside of the fetch function so that was unnecessarily there

src/Data/Managers/V2/GetMetadata/useGetMetadata.ts
matborowczyk: added

src/Data/Managers/V2/GetMetadata/useGetMetadata.ts
matborowczyk: added

src/Data/Managers/V2/PostMetadata/usePostMetadata.ts
matborowczyk: added

src/Data/Managers/V2/usePostOrder/usePostOrder.ts
```diff
@@ -0,0 +1,45 @@
+import { UseMutationResult, useMutation } from "@tanstack/react-query";
+import { PrimaryBaseUrlManager, words } from "@/UI";
+import { InstanceForApi } from "@/UI/Components/Diagram/interfaces";
+import { useFetchHelpers } from "../helpers";
+
+/**
+ * Custom hook for sending an order batch from Instance COmposer.
```
LukasStordeur, line 7:
```suggestion
 * React Query hook for sending an order batch from Instance Composer.
```

src/UI/Components/InstanceProvider/InstanceProvider.tsx
```diff
@@ -1,34 +1,79 @@
...
+ * It serves purpose to not provide all the neccesary data for the Canvas at the same time to avoid unnecessary rerenders,
```
LukasStordeur, line 11:
```suggestion
 * It serves purpose to not provide all the necessary data for the Canvas at the same time to avoid unnecessary rerenders,
```

The text confuses me a bit, why would it prevent serving necessary data?

src/UI/Components/InstanceProvider/InstanceProvider.tsx
```diff
@@ -1,34 +1,79 @@
...
+    instanceVersion,
+    "coordinates",
+  ).useOneTime();
+
+  if (isLoading || isMetadataLoading) {
+    return <LoadingView aria-label="instance_composer_editor-loading" />;
+  }
+
+  if (isError || isMetadataError) {
+    const errorMessage = error
```
LukasStordeur, line 51: Please, we should avoid chaining ternaries, it makes code not so readable.

src/UI/Components/InstanceProvider/InstanceProvider.tsx
```diff
@@ -1,34 +1,79 @@
...
-  const { queryResolver } = useContext(DependencyContext);
-  const [data] = queryResolver.useOneTime<"GetInstanceWithRelations">({
-    kind: "GetInstanceWithRelations",
-    id: instanceId,
-  });
+}> = ({ services, mainServiceName, instanceId, editable = false }) => {
+  const { environmentHandler } = useContext(DependencyContext);
+  const environment = environmentHandler.useId();
+
+  const { data, isError, isLoading, error, refetch } =
```
LukasStordeur, line 29: Perhaps it would have been more readable to not opt for deconstruction when having two React Queries in the same component.

It would avoid confusion, because now, you're explicitly renaming the metadata query deconstruction, but not the one for the instance.

I had the same issue in my current wip and solved it like this to keep it neat.
image

src/Data/Managers/V2/GetMetadata/useGetMetadata.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @param service_id - The ID of the service.
+ * @param instanceVersion - The version of the service instance.
+ * @param key - The key for the metadata.
+ * @returns An object containing the custom hook.
+ */
+export const useGetMetadata = (
+  environment: string,
+  service_entity: string,
+  service_id: string,
+  instanceVersion: ParsedNumber | undefined,
```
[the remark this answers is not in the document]
matborowczyk: if we not specify the instance version the query will return metadata for the latest version possible. in the current usage we get the version from the fetched data and according to the typing it can be already there undefined, so either way it moving it to the end is fine I guess.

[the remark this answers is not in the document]
LukasStordeur: Okay, I think making it optional and moving it to the end makes most sense then.

src/Data/Managers/V2/PostMetadata/usePostMetadata.ts
```diff
@@ -0,0 +1,54 @@
...
+import { useMutation } from "@tanstack/react-query";
+import { PrimaryBaseUrlManager } from "@/UI";
+import { useFetchHelpers } from "../helpers";
+
+interface PostMetadataInfo {
+  service_entity: string;
+  service_id: string;
+  key: string;
+  body: {
+    current_version: number;
```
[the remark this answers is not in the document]
matborowczyk: we are passing number, but according to the types that are already in the web-console we receive ParsedNumber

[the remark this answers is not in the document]
LukasStordeur: Why are we sending back a number instead of ParsedNumber?

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
```diff
@@ -0,0 +1,91 @@
...
+  const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
+
+  /**
+   * Fetches a service instance with its related instances recursively.
+   * @param instanceId - The ID of the instance to fetch.
+   * @returns An object containing the fetched instance and its related instances.
+   * @throws Error if the instance fails to fetch.
+   */
+  const fetchService = async (
+    instanceId: string,
```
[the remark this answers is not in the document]
matborowczyk: had to miss addressing that, but I found a way to change that to something close that you suggested,

src/Data/Managers/V2/PostMetadata/usePostMetadata.ts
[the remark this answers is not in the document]
matborowczyk: Nevermind I changed because we are not transforming that version in any way so the type should be the same, my mistake

src/Data/Managers/V2/GetMetadata/useGetMetadata.ts
```diff
@@ -0,0 +1,79 @@
+import { UseQueryResult, useQuery } from "@tanstack/react-query";
+import { ParsedNumber } from "@/Core";
+import { PrimaryBaseUrlManager } from "@/UI";
+import { useFetchHelpers } from "../helpers";
+
+interface Metadata {
+  data: string | undefined;
+}
+
+interface GetMetadataHook {
```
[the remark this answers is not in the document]
LukasStordeur: I used the Get keyword in my wip, I think it makes most sense since this is the return signature of the useGetSomething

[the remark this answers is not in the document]
matborowczyk: changed instanceWithRelationName to keep with that naming

src/Data/Managers/V2/RemoveUser/useRemoveUser.ts
```diff
@@ -7,7 +11,12 @@ import { useFetchHelpers } from "../helpers";
  *
  * @returns {Mutation} - The mutation object provided by `useMutation` hook.
  */
-export const useRemoveUser = () => {
+export const useRemoveUser = (): UseMutationResult<
```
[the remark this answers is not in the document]
matborowczyk: trying anything more generic, or without these parameters leads to typeError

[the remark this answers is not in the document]
LukasStordeur: okay

src/UI/Components/InstanceProvider/InstanceProvider.tsx
```diff
@@ -1,34 +1,78 @@
...
-import { RemoteDataView } from "@/UI/Components";
+import { useGetInstanceWithRelations } from "@/Data/Managers/V2/GetInstanceWithRelations";
+import { useGetMetadata } from "@/Data/Managers/V2/GetMetadata";
+import { DependencyContext, words } from "@/UI";
+import { ErrorView, LoadingView } from "@/UI/Components";
 import Canvas from "@/UI/Components/Diagram/Canvas";
 
+/**
+ * Renders the InstanceProvider component.
+ * It serves purpose to provide all the necessary data for the Canvas at the same time to avoid unnecessary rerenders,
```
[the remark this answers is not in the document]
matborowczyk: I think that this component gives us great control what we want to display and there is no need to share anything inside of the canvas itself beside service models. If comes to the performance I don't know what we can gain there, but If you have ideas then we can definitely talk about it.

cypress/e2e/scenario-8-instance-composer.cy.js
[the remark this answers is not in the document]
LukasStordeur: image
Can you also improve the coverage of the React Query hooks you added?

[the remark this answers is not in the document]
matborowczyk: There is no direct test scenario, yet scenarios for editing instances have to interact with entities based on the position of them in the canvas from creating so if for some reason positioning won't be saved test will fail as we will fallback to default sorting which will cause rearrangement and cypress won't find the entities  based on given coordinates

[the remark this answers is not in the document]
LukasStordeur: the current scenario's all use the default positioning, which so far seemed to be pretty consistent, I'd suggest moving one block in the create and then reuse the coordinates in the edit.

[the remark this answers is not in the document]
matborowczyk: moved embedded-service-entity-extra to non-default place

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+        retry: false,
+      },
+    },
+  });
+  return ({ children }) => (
+    <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+  );
+};
+
+test("when instance returned has referenced other instance, then query will fetch both of them", async () => {
```
LukasStordeur, line 48:
```suggestion
test("when the returned instance has referenced instances, then the query will fetch both of them", async () => {
```
 Both of what? I'm not certain to exactly understand what you're talking about? Is it two referenced instances?

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  expect(result.current.data).toBeDefined();
+  expect(result.current.data?.instance.id).toEqual("test_id");
+  expect(result.current.data?.relatedInstances).toHaveLength(1);
+  expect(result.current.data?.relatedInstances[0].instance.id).toEqual(
+    "test_mpn_id",
+  );
+});
+
+test("when instance returned hasn't referenced other instance, then query will fetch only one instance", async () => {
```
LukasStordeur, line 66: Can you make the text clearer? 
one instance, which one? the root instance?

LukasStordeur: Also,  i'm confused because the mock handler is returning exactly the same HTTP response... how is this working then? I'd expect it to be different, considering you're passing another id?

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * React Query hook to fetch an instance with its related instances from the API.
+ * @param {string} id - The ID of the instance to fetch.
+ * @param {string} environment - The environment in which we are looking for instances.
+ * @returns  {GetInstanceWithRelationsHook} An object containing a custom hook to fetch the instance with its related instances.
+ */
+export const useGetInstanceWithRelations = (
+  id: string,
```
LukasStordeur, line 26: instanceId, I don't think we use simply id in other places for this 🤔

src/Data/Managers/V2/GetMetadata/useGetMetadata.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    },
+  );
+
+  await waitFor(() => expect(result.current.isSuccess).toBe(true));
+
+  expect(result.current.data).toBeDefined();
+  expect(result.current.data).toEqual("metadata_string");
+});
+
+test("when servers return 404 query shouldn't throw error, but return undefined instead", async () => {
```
LukasStordeur, line 52: Why let the hook return undefined? 
In my opinion, the query should return the error state, and whichever component using the query should handle the result of the query.  That's what we do everywhere, I believe?

src/Data/Managers/V2/GetMetadata/useGetMetadata.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    () => useGetMetadata("env", "service_entity", "id", "key", 5).useOneTime(),
+    {
+      wrapper: createWrapper(),
+    },
+  );
+
+  // wait until the query has transitioned to success state
+  await waitFor(() => expect(result.current.isSuccess).toBe(true));
+
+  expect(result.current.data).not.toBeDefined();
```
LukasStordeur, line 75: Even when you don't explicitly return undefined in your hook, in case the server responds with an error state, it will remain undefined. I don't get why you want to make an explicit exception. 
your `if (query.data)` will remain the same, whether your return undefined for it or not.

src/Data/Managers/V2/PostMetadata/usePostMetadata.ts
```diff
@@ -0,0 +1,56 @@
...
+interface PostMetadataInfo {
+  service_entity: string;
+  service_id: string;
+  key: string;
+  body: {
+    current_version: ParsedNumber;
+    value: string;
+  };
+}
+/**
```
LukasStordeur, line 15: new line before comment

src/Data/Managers/V2/GetMetadata/useGetMetadata.ts
```diff
@@ -0,0 +1,79 @@
...
+        queryKey: [
+          "get_metadata",
+          service_entity,
+          service_id,
+          key,
+          environment,
+          instanceVersion,
+        ],
+        retry: false,
+        enabled: instanceVersion !== undefined,
```
LukasStordeur, line 75: what happens when the instanceVersion doesn't exist? Does the query not fire, or how does it work concretely? Is this something we need to test for?

src/Data/Managers/V2/PostOrder/usePostOrder.ts
```diff
@@ -0,0 +1,45 @@
...
+  const headers = createHeaders(environment);
+  const baseUrlManager = new PrimaryBaseUrlManager(
+    globalThis.location.origin,
+    globalThis.location.pathname,
+  );
+  const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
+
+  /**
+   * Sends an order with the provided bundle instances.
+   * @param {InstanceForApi[]} serviceOrderItems - The bundled instances to include in the order.
```
LukasStordeur, line 23: Why not adjust the name of the typing too? To keep it consistent. Doesn't it match the typing used in the Order Overview page by any hasard?

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -77,53 +83,54 @@ const Canvas = ({
...
+   * @param {CustomEvent} event - The event object.
+   */
   const handleEditEvent = (event) => {
     const customEvent = event as CustomEvent;
     setCellToEdit(customEvent.detail);
     setIsFormModalOpen(true);
   };
 
+  /**
+   * Handles the filtering of the unchanged entities and sending bundled instances to the backend.
```
LukasStordeur, line 108:
```suggestion
   * Handles the filtering of the unchanged entities and sending serviceOrderItems to the backend.
```

src/UI/Components/Diagram/helpers.ts
```diff
@@ -529,3 +531,71 @@ export const toggleLooseElement = (
     }),
   );
 };
+
+/**
+ * Gets the coordinates of all cells in the graph.
```
LukasStordeur, line 536: Someone not familiar with jointJS might wonder what the graph is. Maybe specify that this is the canvas we are talking about?

src/UI/Components/Diagram/helpers.ts
```diff
@@ -529,3 +531,71 @@ export const toggleLooseElement = (
...
+    .map((cell) => ({
+      id: cell.id,
+      name: cell.attributes.entityName,
+      attributes: cell.attributes.instanceAttributes,
+      coordinates: cell.attributes.position,
+    }));
+};
+
+/**
+ * Applies coordinates to cells in the graph.
```
LukasStordeur, line 554: same here

src/UI/Components/Diagram/helpers.ts
```diff
@@ -529,3 +531,71 @@ export const toggleLooseElement = (
...
+/**
+ * Moves a cell away from any cells it is colliding with.
+ *
+ * @param {dia.Graph} graph - The graph containing the cell.
+ * @param {dia.Cell} cell - The cell to move.
+ */
+export const moveCellFromColliding = (graph: dia.Graph, cell: dia.Cell) => {
+  let isColliding = false;
+  do {
+    const elementsUnder = graph
```
LukasStordeur, line 589: what's an elementsUnder?

src/UI/Components/InstanceProvider/InstanceProvider.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    await waitFor(() => {
+      expect(screen.getByText("Loading")).toBeInTheDocument();
+    });
+
+    await waitFor(() => {
+      expect(screen.getByLabelText("Composer-Container")).toBeInTheDocument();
+    });
+  });
+
+  it("should render error view when is there is error returned from the instance endpoint", async () => {
```
LukasStordeur, line 72:
```suggestion
  it("should render the ErrorView when there is an error returned from the instance endpoint", async () => {
```
I don't see the instance endpoint being handled, only the whole inventory.

src/UI/Components/InstanceProvider/InstanceProvider.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    await waitFor(() => {
+      expect(
+        screen.getByText(
+          "The following error occured: Failed to fetch instance with id: id",
+        ),
+      ).toBeInTheDocument();
+    });
+  });
+
+  it("should render error view when is there is error returned from the metadata endpoint", async () => {
```
LukasStordeur, line 98:
```suggestion
  it("should render the ErrorView when there is an error returned from the metadata endpoint", async () => {
```
Also, you're talking about general errors, but in fact, you made an exception for the 404 (which I still don't think is right) But the descriptive text of your test is not accurate.

LukasStordeur (changes requested):

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.test.tsx
matborowczyk: We return two different responses depending on the Id, but as we are depending on the same endpoint query(recursion) there is if statement inside

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.ts
matborowczyk: we use instanceId below, but I changed places then

src/Data/Managers/V2/GetMetadata/useGetMetadata.test.tsx
matborowczyk: If some instances won't have coordinates then they won't be available in the composer, I think we want to avoid that, also if for some reason coordinates fails to save on creation that would be the same case there.

src/Data/Managers/V2/PostOrder/usePostOrder.ts
matborowczyk: it is close, it's missing two properties, should I extend that interface?

src/UI/Components/Diagram/helpers.ts
matborowczyk: changed to `elementsUnderTheCell`

LukasStordeur: But what is underTheCell then ? Is it like, connected to it? A decendant? Is it visually underneath ?

src/UI/Components/Diagram/helpers.ts
matborowczyk: graph isn't exactly the canvas, should I add link to the documentation there?

src/Data/Managers/V2/PostOrder/usePostOrder.ts
LukasStordeur: you can, yes, either you extend the interface and name it ComposerServiceOrderItems, or you add the missing properties as optional to the existing interface. Which properties are missing?

src/UI/Components/Diagram/helpers.ts
matborowczyk: The docstrings explains that this function moves cell away from any cell it is colliding with, so yes, visually

LukasStordeur: Then overlapping elements is probably the term you need :D or simply collidingElements

src/UI/Components/Diagram/helpers.ts
LukasStordeur: that would be nice

src/Data/Managers/V2/PostOrder/usePostOrder.ts
matborowczyk: related to the composer, so I will extend the interface

src/UI/Components/InstanceProvider/InstanceProvider.test.tsx
matborowczyk: This is the correct endpoint, just all other data is being send through parameters, I'll rename test case though.

src/UI/Components/InstanceProvider/InstanceProvider.test.tsx
matborowczyk: This is the correct endpoint, just all other data is being send through parameters, I'll rename test case though.

src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -297,7 +322,7 @@ const Canvas = ({
         }
         editable={editable}
       />
-      <CanvasWrapper id="canvas-wrapper">
+      <CanvasWrapper id="canvas-wrapper" aria-label="Composer-Container">
```
LukasStordeur, line 325: For accessibility, adding an aria-label to a div or a span, requires you to define a role. If the aria-label was added for testing purposes, please use the data-testid instead.

src/UI/Components/Diagram/helpers.ts
```diff
@@ -257,17 +259,17 @@ const doesElementIsEmbeddedWithExhaustedConnections = (
  * Function that will merge state from Instance Composer to proper object for order_api endpoint.
  * Instance composer state is being split into multiple objects that could be embedded into other available, so we need to recursively
  * go through all of them to group, and sort them
- * @param {InstanceForApi[]} instances all of the instances that were created/edited in the instance, not including the one passed in second parameter
- * @param {InstanceForApi} instance instance which is used taken into consideration as the parent of the possible embedded
+ * @param {ComposerServiceOrderItem[]} instances all of the instances that were created/edited in the instance, not including the one passed in second parameter
+ * @param {ComposerServiceOrderItem} instance instance which is used taken into consideration as the parent of the possible embedded
```
LukasStordeur, line 263: The comment is confusing me. Can you update it? Is instance then the right naming ? Seems not specific enough based on the comment

src/UI/Components/InstanceProvider/InstanceProvider.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+  server.listen();
+});
+// Reset any request handlers that we may add during the tests,
+// so they don't affect other tests.
+afterEach(() => server.resetHandlers());
+// Clean up after the tests are finished.
+afterAll(() => server.close());
+
+describe("UserManagementPage", () => {
+  it("should render loading view when is there is no response from the endpoints yet,", async () => {
```
LukasStordeur, line 46:
```suggestion
  it("should render LoadingView when there are no responses from the endpoints yet,", async () => {
```

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { renderHook, waitFor } from "@testing-library/react";
+import { HttpResponse, http } from "msw";
+import { setupServer } from "msw/node";
+import { testInstance } from "@/UI/Components/Diagram/Mock";
+import { useGetInstanceWithRelations } from "./useGetInstanceWithRelations";
+
+export const server = setupServer(
+  http.get("/lsm/v1/service_inventory", async (params) => {
+    if (params.request.url.match(/test_id/)) {
```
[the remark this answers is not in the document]
matborowczyk: I don't think I can avoid if statement, as we are testing if the recurrent call will be fired correctly, so we need two objects to make distinction between them to be sure that one object isn't assigned twice. I can add comments though if you find it not readable though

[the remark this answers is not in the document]
LukasStordeur: Of course you can avoid it. You can setupServer in each test case. Or do the server.use method like you did in another test if you want to keep the beforeach/aftereach test loops.

src/Data/Managers/V2/GetInstanceWithRelations/useGetInstanceWithRelations.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+        retry: false,
+      },
+    },
+  });
+  return ({ children }) => (
+    <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+  );
+};
+
+test("if the fetched instance has referenced other instance, then query will returns given instance with that related instance", async () => {
```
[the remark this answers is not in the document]
matborowczyk: change the depth to only first one, as that's what we will also do in the V2

src/Data/Managers/V2/GetMetadata/useGetMetadata.ts
```diff
@@ -0,0 +1,79 @@
...
+  const getMetadata = async (): Promise<Metadata> => {
+    const response = await fetch(
+      baseUrl +
+        `/lsm/v1/service_inventory/${service_entity}/${service_id}/metadata/${key}?current_version=${instanceVersion}`,
+      {
+        headers,
+      },
+    );
+
+    // If the value isn't set for the given key, we receive a 404 which shouldn't break the flow of the application, as the endpoint doesn't serve critical data.
```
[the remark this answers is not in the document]
matborowczyk: That's the last place to intercept response status, later we have to base on error message which theoretically could change at any time and block view of the composer. Anyway I'll get rid of it as I found out that query for service instance also can return the metadata, which I missed in the waterfall of parameter it accepts

src/Data/Store/ServiceInstanceSliceWithRelations.ts
```diff
@@ -1,6 +1,6 @@
 import { Action, action } from "easy-peasy";
 import { RemoteData } from "@/Core";
-import { InstanceWithReferences } from "../Managers/GetInstanceWithRelations/interface";
+import { InstanceWithReferences } from "../Managers/V2/GetInstanceWithRelations";
```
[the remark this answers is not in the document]
matborowczyk: changed

cypress/e2e/scenario-8-instance-composer.cy.js
```diff
@@ -1,778 +1,827 @@
...
-
-      // Check if only one row has been deleted to the table.
-      cy.get('[aria-label="InstanceRow-Intro"]', { timeout: 90000 }).should(
-        "have.length",
-        5,
-      );
-    });
-  });
-}
+//TODO: tests are commented out as there is ongoing redesign of the instance composer which will affect how the user interact with it and how e2e test should look like
```
LukasStordeur, line 1: Just make certain they are only commented out on ISO8

src/UI/Components/Diagram/actions.ts
```diff
@@ -227,12 +231,12 @@ export function appendInstance(
...
   serviceWithReferences.relatedInstances.forEach((relatedInstance) => {
     const isInstanceMain = false;
-    const cellAdded = graph.getCell(relatedInstance.instance.data.id);
+    const cellAdded = graph.getCell(relatedInstance.id);
     if (!cellAdded) {
       appendInstance(
         paper,
         graph,
-        relatedInstance,
+        { instance: relatedInstance, relatedInstances: [] },
```
LukasStordeur, line 239: hmm, this is a little confusing, can you add a comment about the reasoning?

cypress/e2e/scenario-8-instance-composer.cy.js
matborowczyk: iso7 does have this scenario intact

src/UI/Components/Diagram/actions.ts
matborowczyk: changed typing and that resolved the issue there

LukasStordeur (approved):

PR #5681, opened 2024-04-11
Fix enum field issue

### LINKED ISSUE #5676: BUG - duplicate form doesn't retain preselected values
https://inmanta.slack.com/archives/C045Y42N49W/p1712754080954529

LukasStordeur:
# Description

I replaced the select as it was before, whilst keeping the original type of the options. Previously the values were stringified in the Object.keys mapping. Now that the values keep their typing, we can safely remove the logic in the mapping prepping the body to send to the api that was previously assuming the value would always be a string in case of enums. The issue wasn't specific to the duplicate form.

close #5676

### DIFF
--- a/src/UI/Components/ServiceInstanceForm/Components/SelectFormInput.tsx
+++ b/src/UI/Components/ServiceInstanceForm/Components/SelectFormInput.tsx
@@ -22,6 +22,21 @@ interface Props {
   shouldBeDisabled?: boolean;
   handleInputChange: (value) => void;
 }
+/**
+ * A form input component for managing a select input.
+ * @component
+ *
+ * @param {Props} props - The props for the SelectFormInput component.
+ *  @prop {Record<string, string | ParsedNumber>} options - The options for the select input.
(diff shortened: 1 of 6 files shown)


src/UI/Components/ServiceInstanceForm/Components/SelectFormInput.tsx
```diff
@@ -28,15 +32,50 @@ export const SelectFormInput: React.FC<Props> = ({
...
+      isSelected:
+        selectOptions.length === 1 || options[option] === attributeValue,
     };
   });
 
+  const onToggleClick = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const onSelect = (
```
matborowczyk, line 51: component and its internal functions could get docstrings to align with the new habits

LukasStordeur: added

matborowczyk (approved):

PR #5698, opened 2024-04-22
Resolve filtering issue in the resource details

### LINKED ISSUE #5697: Removing log filter in resource details send invalid request
https://inmanta.slack.com/archives/C045Y42N49W/p1713514277206439

matborowczyk:
# Description

* Short description here *

closes #5697 

# Self Check:

- [ ] Changelog entry

https://github.com/inmanta/web-console/assets/113331659/79a5ee17-afd9-4252-a5aa-259a63a549da

### DIFF
--- a/src/Slices/ResourceDetails/UI/Tabs/LogTab/Controls/LogLevelFilter.tsx
+++ b/src/Slices/ResourceDetails/UI/Tabs/LogTab/Controls/LogLevelFilter.tsx
@@ -21,7 +21,7 @@ export const LogLevelFilter: React.FC<Props> = ({ filter, setFilter }) => {
     });
 
   const onSelect = (selection) => {
-    update(selection === null ? undefined : selection);
+    update(selection === null || selection === "" ? undefined : selection);
   };
 
   const deleteChip = () =>
--- a/changelogs/unreleased/5697-resouce-details-filtering-bug.yml
+++ b/changelogs/unreleased/5697-resouce-details-filtering-bug.yml
@@ -1,0 +1,6 @@
+description: "Resolve bug in resource details filtering, where removing log filters would send invalid request."
+issue-nr: 5697
+change-type: patch
+destination-branches: [master, iso7]
+sections:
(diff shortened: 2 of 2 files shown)


src/Slices/ResourceDetails/UI/Tabs/LogTab/Controls/LogLevelFilter.tsx
```diff
@@ -21,7 +21,7 @@ export const LogLevelFilter: React.FC<Props> = ({ filter, setFilter }) => {
     });
 
   const onSelect = (selection) => {
-    update(selection === null ? undefined : selection);
+    update(selection === null || selection === "" ? undefined : selection);
```
LukasStordeur, line 24: image

That's the error from BE, so how is setting it to an empty string solve the problem since the error explicitly says it can't be an empty string? I'm confused

matborowczyk: but what I added is opposite of the setting it to empty string. We use some general purpose TextSelect, and it was simpler to add that condition here that to change given component.

The issue is that when deleting chosen option from select, internal function set selection to empty string and that's why it sent invalid request

LukasStordeur: So, setting it to an empty string removes the param from the API-call? If yes, okay, this can be seen as a solution. 

Otherwise, I'd go for updating the getUrl method to make sure that the query doesn't keep empty filterings, may they be empty strings; null; undefined for the loglevels. 

`web-console\src\Slices\ResourceDetails\Data\ResourceLogs\getUrl.ts`

matborowczyk: Currently yes, if that general purpose component set the value to empty string the param will be removed from the api-call

LukasStordeur (approved):

PR #5758, opened 2024-05-22
Add functionality to filter instances through labels

### LINKED ISSUE #5710: in the service instance view, make it possible to click on the labels or pie chart to set the filter
image

matborowczyk:
# Description

on the video when clicking on "warning" and "info" labels it seems like nothing happens, or the filters are removed, it's due to the lack of the labels in the service that corresponds to given category.

closes #5710 

# Self Check:

- [ ] Changelog entry

https://github.com/inmanta/web-console/assets/113331659/e93d3d35-26ff-490e-be1f-659f6563b086


LukasStordeur:
As much as I like native JS approaches, it doesn't align much with how the app is working. Using React and co, I believe using the document to add eventListeners goes against the best practices. I get we use it when we don't have an option, or if we need to work around React, but this is not the case now. 

Why not use the context provider of the Service Inventory and sharing it with the chart? 

Also, I'm not a fan of having the "empty" states doing nothing.  I'd at least expect the chips to be added or have some sort of user feedback instead of a blank action.

A tooltip might be interesting to indicate that there's an action there.

matborowczyk:
> As much as I like native JS approaches, it doesn't align much with how the app is working. Using React and co, I believe using the document to add eventListeners goes against the best practices. I get we use it when we don't have an option, or if we need to work around React, but this is not the case now.
> 
> Why not use the context provider of the Service Inventory and sharing it with the chart?
> 
> Also, I'm not a fan of having the "empty" states doing nothing. I'd at least expect the chips to be added or have some sort of user feedback instead of a blank action.
> 
> A tooltip might be interesting to indicate that there's an action there.

Right, I didn't think of the context, somehow I forgot about it totally. using event listeners was was simplest and "fastest" way to get the data from there to filters but I'll change that.

I am aware about empty states being an issue, as I forgot about context, the data that I had didn't allow me to know beforehand about whether the service have any filters that fit under this category.

I'll move those into some local context to provide better and two-way communication between those two places.

LukasStordeur (commented):
I'm confused, why does warning and info have no filters? Some others do have filters when there are no matching instances. This behavior feels buggy to me. 

Perhaps this feature should have been discussed further before development

src/Slices/ServiceInventory/UI/ServiceInventory.tsx
```diff
@@ -75,54 +94,72 @@ export const ServiceInventory: React.FunctionComponent<{
...
-                  instances={instances}
-                  serviceEntity={service}
-                  sort={sort}
-                  setSort={setSort}
-                />
-              ) : (
-                <EmptyView
-                  message={words("inventory.empty.message")(serviceName)}
-                  aria-label="ServiceInventory-Empty"
+    <LabelContext.Provider
```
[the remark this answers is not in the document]
matborowczyk: Good point, I merged them together

matborowczyk:
https://github.com/inmanta/web-console/assets/113331659/25b4201a-5674-46eb-b685-7717f46fe481

src/Slices/ServiceInventory/UI/ServiceInventory.tsx
```diff
@@ -43,6 +42,31 @@ const PreppedServiceInventory: React.FC<{ service: ServiceModel }> = ({
   />
 );
 
+interface ServiceInventoryContextProps {
```
LukasStordeur, line 45: any objections to put the labels into an object? Feels like it make sense to group them to me.

matborowczyk: grouped

src/Slices/ServiceInventory/UI/Components/RowActionsMenu/DestroyAction/DestroyAction.test.tsx
```diff
@@ -41,7 +41,19 @@ function setup() {
...
+          <ServiceInventoryContext.Provider
+            value={{
+              labels: {
+                danger: [],
+                warning: [],
+                success: [],
+                info: [],
+                no_label: [],
+              },
+              onClick: jest.fn(),
```
LukasStordeur, line 53: is the onClick not part of the labels logic?

matborowczyk: It's filtering callback, so it uses labels but per se I would keep it separated, I can change the name though to be more descriptive

LukasStordeur: yes, because now it's not clear what the onClick does or with what it belongs. And for me it relates to the labels, it's when you click on a label/graph section.

LukasStordeur: if it makes more sense you can also rename the group to : `labelFiltering`.

src/Slices/ServiceInventory/UI/ServiceInventory.tsx
LukasStordeur, line 1: Can you also add new-lines between the different sections? Makes it a bit more readable and coherent :D

src/Slices/ServiceInventory/UI/ServiceInventory.tsx
```diff
@@ -43,6 +42,35 @@ const PreppedServiceInventory: React.FC<{ service: ServiceModel }> = ({
...
```
LukasStordeur, line 45: Props if these are local props.

src/Slices/ServiceInventory/UI/GetInstancesContext.tsx
LukasStordeur, line 1: Why did you remove it instead of adding the labels to it?  I don't mind the name change if we keep it consistent.

src/Slices/ServiceInventory/UI/Components/RowActionsMenu/DestroyAction/DestroyAction.test.tsx
matborowczyk: changed

src/Slices/ServiceInventory/UI/GetInstancesContext.tsx
matborowczyk: Actually I have no idea, I think it was faster to add refetch to the "ServiceInventoryContext" than other way around as I made them at first independent and separate. That's my guess after holiday :D

src/Slices/ServiceInventory/UI/ServiceInventory.tsx
```diff
@@ -43,6 +42,35 @@ const PreppedServiceInventory: React.FC<{ service: ServiceModel }> = ({
...
+      warning: [],
+      success: [],
+      info: [],
+      no_label: [],
+    },
+    onClick: (_label) => {
+      // Default to no-op
+      return;
+    },
+    refetch: () => {
```
LukasStordeur: why not like this, that's how it was in the initial Context
```
onClick: (_label) => null;
refetch: () => null;
```

src/Slices/ServiceInventory/UI/ServiceInventory.tsx
LukasStordeur, line 1: Still missing some comments in this file :)

src/Slices/ServiceInventory/UI/ServiceInventory.tsx
LukasStordeur: my bad, I'll place it on the right lines again, because it still applies ;)

src/Slices/ServiceInventory/UI/ServiceInventory.tsx
```diff
@@ -43,6 +42,37 @@ const PreppedServiceInventory: React.FC<{ service: ServiceModel }> = ({
...
+}
+
+export const ServiceInventoryContext = createContext<Props>({
+  labelFiltering: {
+    danger: [],
+    warning: [],
+    success: [],
+    info: [],
+    no_label: [],
+    onClick: (_label) => {
```
LukasStordeur, line 64: onClick: (_label) => null;
refetch: () => null;

LukasStordeur (approved):

PR #5770, opened 2024-05-28
add column for managed resources in resource discovery page

### LINKED ISSUE #5765: Resource Discovery - Add column is_managed resource
Resource Discovery endpoint is now providing a link to the managed resource if available. The managed resource id needs to be extracted from the provided link.

LukasStordeur:
# Description

Added a column to display the link to the managed resource if available. When no link is available, a dash is displayed. 
The screenshot doesn't yet have managed resources. 

 closes #5765 

image

### DIFF
--- a/src/Slices/ResourceDiscovery/Core/Query.ts
+++ b/src/Slices/ResourceDiscovery/Core/Query.ts
@@ -31,6 +31,7 @@ export interface Manifest {
 
 export interface DiscoveredResource {
   discovered_resource_id: string;
+  managed_resource_uri: string | null;
   values: unknown;
 }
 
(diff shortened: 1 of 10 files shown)


src/Slices/ResourceDiscovery/Core/Query.ts
```diff
@@ -31,6 +31,7 @@ export interface Manifest {
 
 export interface DiscoveredResource {
   discovered_resource_id: string;
+  managed_resource_uri: string | null;
```
matborowczyk, line 34: just out of curiosity, in case of no uri won't that value be undefined? I'm not sure, but often that's the case with the responses from the backend

matborowczyk (approved):

src/Slices/ResourceDiscovery/Core/Query.ts
LukasStordeur: No, it is null :D

PR #5777, opened 2024-05-30
Issue/5766 improve filtering get instance form

### LINKED ISSUE #5766: Replace filter.name with new filtering on GET instances
GET instances is now enhanced with the possibility to filter on the `service_identity_attribute_value` which can either be a displayName or an id. 

Update the form to use this filtering to GET the Inter-Service-Relations.

LukasStordeur:
# Description

closes #5766 

The dropdown displayed the UUID previously when editing an instance. It does show the correct name now, and the filtering works whether the user types in a UUID or an instance name. 

https://github.com/inmanta/web-console/assets/44098050/05fe55d8-1bb3-4fd1-acd5-acc1d0e7d51f


src/UI/Components/SingleTextSelect/SingleTextSelect.tsx
```diff
@@ -114,15 +114,14 @@ export const SingleTextSelect: React.FC<Props> = ({
       ]);
     } else {
       setSelectOptions(options);
+
+      if (inputValue && inputValue === selected) {
+        setInputValue(getDisplayValue(selected));
+      }
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [options]);
```
matborowczyk, line 123: isn't selected here required? I'm out of the flow here so if the result is correct then it's fine :)

LukasStordeur: tested and indeed, also works. I had thought it would chain renders but it's fine.

matborowczyk (approved):

matborowczyk (approved):

PR #5789, opened 2024-05-31
Auth appended to the routing, no_auth + database_auth injected

### LINKED ISSUE #5759: Authentication refactor

matborowczyk:
# Description

Auth appended to the routing, no_auth injected, database_auth partially injected.

Request for the review as proposed to just verify if the approach is correct with what was agreed

closes #5759 

# Self Check:

- [ ] Changelog entry


src/Data/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+import KeycloakAuthContext from "./Providers/KeycloakAuthContext";
+import NoAuthProvider from "./Providers/NoAuthProvider";
+
+interface AuthProviderProps {
+  children: React.ReactNode;
+  config: undefined | AuthConfig | LocalConfig;
+  keycloakUrl: string | undefined;
+}
+
+const AuthProvider = ({ children, config, keycloakUrl }: AuthProviderProps) => {
```
bartv, line 13: Why can't this just be a function with typed arguments?

src/Data/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
```
bartv, line 1: I would like to see at least a comment (and maybe also better naming) to make it clear that this is the factory function

src/Data/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return <DatabaseAuthProvider>{children}</DatabaseAuthProvider>;
+  }
+  if (config?.method === "oidc") {
+    return (
+      <KeycloakAuthContext config={{ ...config, url: keycloakUrl }}>
+        {children}
+      </KeycloakAuthContext>
+    );
+  }
+  return <NoAuthProvider>{children}</NoAuthProvider>;
```
bartv, line 24: Why not a switch case and auth is the default case?

src/Data/Auth/Providers/AuthContext.ts
```diff
@@ -0,0 +1,16 @@
+import { createContext } from "react";
```
bartv, line 1: is "Props" the convention in ts for interfaces?
This also needs a comment that this is the interface.

Interfaces define the "contract" to the other code. So here it needs a clear documentation of what the semantics are of each of these functions

src/Data/Auth/Providers/DatabaseAuthProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+  getCookie,
+  removeCookie,
+} from "../../Common/CookieHelper";
+import { AuthContext } from "./AuthContext";
+
+interface AuthProviderProps {
+  children: React.ReactNode;
+}
+
+const DatabaseAuthProvider = ({ children }: AuthProviderProps) => {
```
bartv, line 14: Why use the context and provider indirection? If the provider has the same interface you only construct this one and the context is the same for all of them?

src/Data/Managers/V2/helpers/useHandleErrors/useHandleErrors.ts
```diff
@@ -6,37 +6,15 @@ import { useContext } from "react";
 import { DependencyContext } from "@/UI";
 
 export const useHandleErrors = () => {
```
LukasStordeur, line 8: Is it really necessary now to have a hook for this?

src/UI/Root/Root.tsx
```diff
@@ -31,14 +30,14 @@ export const Root: React.FC = () => {
   const GlobalStyleProxy: any = GlobalStyles;
 
   return (
-    <QueryClientProvider client={queryClient}>
+    <>
       <ReactQueryDevtools initialIsOpen={false} />
       <GlobalStyleProxy />
       <KeycloakProvider>
```
LukasStordeur, line 36: Since you have the new authProvider, `<KeycloakProvider>` can be removed.

src/index.tsx
```diff
@@ -2,10 +2,12 @@ import "@patternfly/react-core/dist/styles/base.css";
 import "@/Core/Language/Extensions";
 import React from "react";
 import { Flex } from "@patternfly/react-core";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { StoreProvider } from "easy-peasy";
 import { createRoot } from "react-dom/client";
 import { getStoreInstance } from "@/Data";
 import { Root } from "@/UI/Root";
+import AuthProvider from "./Data/Auth/AuthProvider";
```
LukasStordeur, line 10: `import { AuthProvider } from "./Data/Auth/AuthProvider";`

(After you removed the default export)

src/Data/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+  keycloakUrl: string | undefined;
+}
+
+const AuthProvider = ({ children, config, keycloakUrl }: AuthProviderProps) => {
+  if (config?.method === "database") {
+    return <DatabaseAuthProvider>{children}</DatabaseAuthProvider>;
+  }
+  if (config?.method === "oidc") {
+    return (
+      <KeycloakAuthContext config={{ ...config, url: keycloakUrl }}>
```
LukasStordeur, line 19: Should also be a Provider

src/Data/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+        {children}
+      </KeycloakAuthContext>
+    );
+  }
+  return <NoAuthProvider>{children}</NoAuthProvider>;
+};
+
+export default AuthProvider;
```
LukasStordeur, line 27: No default exports

src/Data/Auth/Providers/DatabaseAuthProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+  };
+
+  return (
+    <AuthContext.Provider value={{ user, login, logout, updateUser, getToken }}>
+      {children}
+    </AuthContext.Provider>
+  );
+};
+
+export default DatabaseAuthProvider;
```
LukasStordeur, line 40: no default exports

src/Data/Auth/Providers/NoAuthProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+          return "No token";
+        },
+      }}
+    >
+      {children}
+    </AuthContext.Provider>
+  );
+};
+
+export default NoAuthProvider;
```
LukasStordeur, line 23: No export defaults

src/Data/Auth/Providers/KeycloakAuthContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const getName = useCallback((): string | undefined => {
+    return keycloakInstance &&
+      keycloakInstance.profile &&
+      keycloakInstance.profile.username
+      ? keycloakInstance.profile.username
+      : undefined;
+  }, [keycloakInstance]);
+
+  const [user, setUser] = useState<string | undefined>(undefined);
```
LukasStordeur, line 21: Typically, the useStates are placed as first declarations

src/Data/Auth/Providers/KeycloakAuthContext.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState, useEffect, useCallback, useMemo } from "react";
+import Keycloak, { KeycloakConfig } from "keycloak-js";
+import { AuthContext } from "./AuthContext";
+
+interface AuthProviderProps {
+  children: React.ReactNode;
+  config: KeycloakConfig;
+}
+
+const KeycloakAuthProvider = ({ children, config }: AuthProviderProps) => {
```
LukasStordeur, line 10:
```suggestion
const KeycloakAuthProvider: React.FC<React.PropsWithChildren<Props>> = ({ config, children }: AuthProviderProps) => {
```

LukasStordeur (commented):
I see you're mixing the namings of the files and the folders. 
image

src/UI/Root/Root.tsx
matborowczyk: Yes that's the idea, but I don't want to remove everything at once as that would be too much to handle at once :D

src/Data/Managers/V2/helpers/useHandleErrors/useHandleErrors.ts
matborowczyk: removed

src/Data/Auth/Providers/KeycloakAuthContext.tsx
matborowczyk: changed

src/Data/Auth/Providers/KeycloakAuthContext.tsx
matborowczyk: removed that totally, I will verify the implementation though, as I suspect that keycloak could require some tweaks anyway

src/Data/Auth/AuthProvider.tsx
matborowczyk: Correct me if I understood question wrong, but as we auth providers return context component as a wrapper we have to follow the React's convention for deconstructed arguments

src/Data/Auth/Providers/DatabaseAuthProvider.tsx
matborowczyk: We create context in the AuthContext.ts file, which then is accessed in the application through provider, where we also create distinction in the logic between each implementation. I don't know if I explained it clearly enough.

src/Data/Auth/AuthProvider.tsx
matborowczyk: Added

src/Data/Auth/Providers/AuthContext.ts
matborowczyk: Yes, "Props" it is a convention here. Added documentation

src/Data/Auth/Providers/AuthContext.ts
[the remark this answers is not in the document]
matborowczyk: It got adjusted accordingly, I think

LukasStordeur, line 1: image

The folder structure is still wrong.

src/Data/Auth/Providers/AuthContext.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Interface for the authentication context properties.
+ */
```
LukasStordeur, line 5: provider, not context. We need to stay clear because the two terms start to mingle in my head. 
The Context returns a CreateContext that consumes a Provider. 

This means that this file should look like so (minus the comments) instead:
```ts
import { createContext } from "react";

// or AuthProviderInterface, whichever is fine for me.
interface GetAuthProvider {
  getUser: () => void;
  login: () => void;
  logout: () => void;
  updateUser: (user: string, token: string) => void;
  getToken: () => string | null;
}

export const GetAuthContext = createContext<GetAuthProvider>({
  getUser: () => undefined,
  login: () => {},
  logout: () => {},
  updateUser: (_user: string, _token: string) => {},
  getToken: () => null, // or the return "no token"; if you consider the default being without authentication.  although I'm not certain that would be needed, or make sense to return "no token" instead of null
});

```

src/Data/Auth/Providers/KeycloakAuthProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const login = () => {
+    keycloakInstance.clearToken();
+    keycloakInstance.login();
+  };
+
+  /**
+   * Get the access token of the currently logged-in user.
+   * @returns The access token of the currently logged-in user, or null if no user is logged in.
+   */
+  const getToken = () => {
```
LukasStordeur, line 49: missing return type method

src/Data/Auth/Providers/NoAuthProvider.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const NoAuthProvider: React.FC<React.PropsWithChildren> = ({
+  children,
+}) => {
+  return (
+    <GetAuthContext.Provider
+      value={{
+        ...defaultAuthContext,
+        getToken: () => {
+          return "No token";
```
LukasStordeur, line 12: If you have refactored the file with the interface, you'll see that you won't need the override of the value here and can simply either return the `<GetAuthContext.Provider>{children}</GetAuthContext.Provider>`
or do that in the switch case and remove this file; choice is yours on that. I don't mind having it in a separate file for matter of cleaner cut.

src/UI/Dependency/Dependency.tsx
```diff
@@ -52,6 +57,7 @@ export const DependencyContext = createContext<Dependencies>({
   authHelper: new DummyAuthHelper(),
   archiveHelper: new DummyArchiveHelper(),
   authController: new DummyAuthController(),
+  useAuth: defaultAuthContext,
```
LukasStordeur, line 60: I take you'll rename it to `authHelper`, or `authManager `to stay in line with the rest after? useAuth hints at a hook, which can be confusing

src/UI/Dependency/Dependency.tsx
```diff
@@ -87,6 +94,7 @@ export const DependencyProvider: React.FC<{
       authHelper: authHelper || new DummyAuthHelper(),
       archiveHelper: archiveHelper || new DummyArchiveHelper(),
       authController: authController || new DummyAuthController(),
+      useAuth: useAuth || defaultAuthContext,
```
LukasStordeur, line 97: defaultAuthContext not needed anymore if it's for free in the initialization in the `authContext.ts` file

LukasStordeur (commented):
I didn't focus on missing docstrings for now :)

src/Data/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface Props {
+  config: undefined | AuthConfig | LocalConfig;
+  keycloakUrl: string | undefined;
+}
+
+/**
+ * AuthProvider component renders the appropriate authentication provider based on the provided config
+ *
+ * @param config - The authentication configuration.
```
bartv, line 15: This "interface" should be 100% generic. I see config which makes me assume that this is the configuration provided by the backend but then there is still a keycloak specific parameter. This should not be the case:

1. It should be clear what `config` is and I cannot deduce that from the code, the naming or the comments
2. We should not have any authentication method specific parameters in here anymore.

src/Data/Auth/Providers/NoAuthProvider.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { GetAuthContext, defaultAuthContext } from "./AuthContext";
+
```
bartv, line 3:
```suggestion
/**
 * Component that implements a null authentication provider when no authentication is enabled.
 */
```

src/Data/Auth/Providers/AuthContext.ts
```diff
@@ -0,0 +1,51 @@
...
+   * Function to get the current user's token.
+   * @returns The current user's token, or null if no user is logged in.
+   */
+  getToken: () => string | null;
+}
+
+/**
+ * The default values for the authentication context.
+ */
+export const defaultAuthContext: GetAuthProvider = {
```
bartv, line 40: Why have a default provider when there is also the NoAuth provider?

src/Data/Auth/Providers/AuthContext.ts
```diff
@@ -0,0 +1,51 @@
+import { createContext } from "react";
+
+/**
+ * Interface for the authentication context properties.
+ */
+export interface GetAuthProvider {
```
bartv, line 6: I don't understand why an interface can have an action verb (get) in it's name. It describes how to interact with code that implements this.

src/Data/Managers/V2/AddUser/useAddUser.ts
```diff
@@ -39,7 +40,9 @@ export const useAddUser = () => {
       body: JSON.stringify(orderBody),
       headers,
     });
-    handleAuthorization(response);
+    if (response.status === 401) {
```
bartv, line 43: Why do this everywhere manually? Either there is a method to hook into fetch when the status code is 401 or create a function that does this.

or use the mechanism of before and create a function that handles the response.

src/UI/Dependency/Dependency.tsx
matborowczyk: Yes, I planned to use the authController, after I remove old one and free the name, but I can change it to any of those right now

src/UI/Dependency/Dependency.tsx
matborowczyk: `dependencies: Partial<Dependencies>` indicates that any of dependencies properties can be undefined so it's required for the sake of the typescript(I think also tests would react differently, but that I didn't check, and I'm not 100% sure)

LukasStordeur: If you update` authContext.ts` based on the example I gave you, useAuth will never be undefined because it has a default initialization, which is technically the noAuth one.

matborowczyk: but undefined that is addressed here doesn't come from the `useAuth` itself but from the `Partial<Dependencies>` . `authContext` file got correct type

src/Data/Managers/V2/AddUser/useAddUser.ts
matborowczyk: Moved it into handleErrors hook

src/Data/Auth/Providers/AuthContext.ts
matborowczyk: Renamed

src/Data/Auth/Providers/AuthContext.ts
matborowczyk: default context variable is the default values for the creation of the contexts, as well as in the noAuthProvider or test suites that's why it got separated from the `createContext()`

src/Data/Auth/Providers/NoAuthProvider.tsx
matborowczyk: As far as I know it's not possible to remove assignment of the "value" property. the interface assigned to the context is correct, but still errors is there:
`Property 'value' is missing in type '{ children: ReactNode; }' but required in type 'ProviderProps<AuthContextInterface>'`

src/Data/Auth/Providers/AuthContext.ts
LukasStordeur: Okay, but in that case you can simply import the context instead of reinitializing it by importing the defaultAuthContext.

Whether you have a specific object for it or handle it like so:

```ts
import { createContext } from "react";

interface  AuthProviderInterface {
  getUser: () => void;
  login: () => void;
  logout: () => void;
  updateUser: (user: string, token: string) => void;
  getToken: () => string | null;
}

export const AuthContext= createContext<AuthProviderInterface >({
  getUser: () => undefined,
  login: () => {},
  logout: () => {},
  updateUser: (_user: string, _token: string) => {},
  getToken: () => null, // or the return "no token"; if you consider the default being without authentication.  although I'm not certain that would be needed, or make sense to return "no token" instead of null
});
```

You should be able to use it like this since it is already instanciated. You don't need to reset the defaults or make sure it has defaults because it already has it built in when you declared it. 
```ts
  const authContext= useContext(AuthContext);
  // rest of code
  return (
     <AuthContext.Provider value={authContext}>
         {children}
     </AuthContext.Provider>
  )
```

src/Data/Auth/Providers/NoAuthProvider.tsx
LukasStordeur: I don't get why you need to override the defaults for no auth when no auth is the default. You can use the context as is in that case.
```ts
  const authContext= useContext(AuthContext);
  // rest of code
  return (
     <AuthContext.Provider value={authContext}>
         {children}
     </AuthContext.Provider>
  )
```

src/Data/Auth/Providers/AuthContext.ts
matborowczyk: Oh, didn't think of that, it felt wrong to call context in the component that will provide it to the application. variable itself will stay as it's required in the test files

src/Data/Auth/KeycloakProvider.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import { Bullseye, Spinner } from "@patternfly/react-core";
+import { ReactKeycloakProvider } from "@react-keycloak/web";
+import Keycloak, { KeycloakConfig } from "keycloak-js";
+
+export interface AuthConfig extends KeycloakConfig {
+  method: "oidc";
+}
+
+export interface LocalConfig {
```
LukasStordeur, line 10: Why is this one in the KeycloakProvider?

src/Data/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const AuthProvider: React.FC<React.PropsWithChildren<Props>> = ({
+  children,
+  config,
+}) => {
+  switch (config?.method) {
+    case "database":
+      return <DatabaseAuthProvider>{children}</DatabaseAuthProvider>;
+    case "oidc":
+      return (
+        <KeycloakProvider config={config}>
```
LukasStordeur, line 26: This doesn't make sense to me. Any reason why you wrapped it like this instead of having the entire implementation in one file?

src/Data/Auth/KeycloakProvider.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import { Bullseye, Spinner } from "@patternfly/react-core";
+import { ReactKeycloakProvider } from "@react-keycloak/web";
+import Keycloak, { KeycloakConfig } from "keycloak-js";
+
+export interface AuthConfig extends KeycloakConfig {
```
LukasStordeur, line 6: Why is the AuthConfig extending KeycloakConfig instead of being generic?

src/Data/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { AuthConfig, KeycloakProvider, LocalConfig } from "./KeycloakProvider";
+import { DatabaseAuthProvider } from "./PartialProviders/DatabaseAuthProvider";
```
LukasStordeur, line 3: PartialProviders? Providers was just fine as folder name. You just need to put the right files under the right name. I'm not even sure what I need to understand under PartialProviders.

src/Data/Managers/V2/helpers/useHelpers.ts
```diff
@@ -0,0 +1,42 @@
+import { useContext } from "react";
+import { DependencyContext } from "@/UI";
+
+export const useHelpers = () => {
```
LukasStordeur, line 4: useHelper is not specific enough in this case. These are helpers that are meant to help with http requests. I'd opt for a better naming such as `useHttpHelpers` or something similar.

src/Data/Managers/V2/helpers/useHelpers.ts
```diff
@@ -0,0 +1,42 @@
+import { useContext } from "react";
+import { DependencyContext } from "@/UI";
+
```
LukasStordeur, line 3: can you add comments

src/Slices/Agents/UI/Agents.test.tsx
```diff
@@ -9,9 +9,9 @@ import {
   QueryResolverImpl,
   getStoreInstance,
   CommandResolverImpl,
-  KeycloakAuthHelper,
   QueryManagerResolverImpl,
   CommandManagerResolverImpl,
+  defaultAuthContext,
```
LukasStordeur, line 14: I don't get it, defaultAuthContext is just an object at this point, it's nothing more. Why aren't you using the actual context in the tests?

LukasStordeur: Why don't you access the context directly in the concerned places instead? That's the whole point of having a context. 

You shouldn't need to do this in the CommandManagerResolverImpl 
image

but access the authContext in the different managers if they require it. We go for a context to be able to avoid drilling down.

src/Test/Inject/TestInjector.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { useContext } from "react";
+import { AuthContext } from "@/Data/Auth/AuthContext";
+import { Dependencies, DependencyProvider } from "@/UI";
+
+type Props = {
+  // Define your component props here
+  dependencies: Partial<Dependencies>;
+};
+
+export const TestInjector: React.FC<React.PropsWithChildren<Props>> = ({
```
LukasStordeur, line 10: comments please, what's the purpose of this injector? I see you're only using it in a few tests concerning authentication.

src/UI/Root/Components/PrivateRoute/PrivateRoute.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { DependencyContext } from "@/UI/Dependency";
+import { Initializer } from "../Initializer";
+
+/**
+ * Renders a private route component that checks if the user is authenticated before rendering the child components.
+ * If the user is not authenticated, it triggers login flow.
+ *
+ * @returns The rendered private route component.
+ */
+const PrivateRoute = () => {
```
LukasStordeur, line 12: `RouterOutlet` is a more conventional naming I believe. 
I'm a bit confused in the way you're using it in the Root file, is it common to have a Route with an element containing the outlet?

src/UI/Root/Root.test.tsx
```diff
@@ -63,15 +67,26 @@ function setup() {
...
-    </MemoryRouter>
+    <QueryClientProvider client={queryClient}>
+      <MemoryRouter initialEntries={["/"]}>
+        <StoreProvider store={store}>
+          <DependencyProvider
+            dependencies={{
+              ...dependencies,
+              queryResolver,
+              environmentHandler,
+              authHelper: {
```
LukasStordeur, line 78: Why wouldn't it be part of the mocked dependencies ? You'd avoid the need to repeat importing the defaultAuthContext all over the place.

src/Data/Managers/V2/helpers/useHelpers.ts
bartv: I would put in the name what it is a helper for. The "use" part I also do not get but that might be something that is specific to the js/ts ecosystem.

As far as I can tell this is a helper for `fetch` so I would name this `fetchHelpers` or if the `use` has a meaning: `useFetchHelpers`

src/Data/Auth/AuthProvider.tsx
matborowczyk: I tried it, but when it was in one file I couldn't reach the keycloak instance from the keycloakprovider. The logic has to be one level below initialisation of the context I guess

src/Data/Auth/KeycloakProvider.tsx
matborowczyk: What we receive from the config.js file is the KeyclokaConfig extended by one extra property. Should I rename it?

src/Data/Auth/KeycloakProvider.tsx
matborowczyk: That was the top place that requires it, and when I had it in AuthProvider file it created circular dependency and I didn't feel like type.ts file is required here, but I can be mistaken.

src/Slices/Settings/UI/Tabs/Tabs.tsx
```diff
@@ -23,7 +23,7 @@ export const Tabs: React.FC<Props> = ({ activeTab, setActiveTab }) => {
   const { authHelper, environmentHandler } = useContext(DependencyContext);
   const tokenTooltipRef = useRef<HTMLElement>();
   const selected = environmentHandler.useSelected();
-  const tokenTabDisabled = authHelper.isDisabled();
+  const tokenTabDisabled = !authHelper.getUser();
```
bartv, line 26: This is a bit weird because now you assume that the contract of "getUser" is that authentication is disabled. This is what makes code brittle: you have implicit contracts between modules.

There are two options:
1. You make this explicit in the documentation of getUser so that if at some point we change this (maybe we can have auth without a user as some point)  it is documented
2. You add an explicit method to the interface for checking if auth is disabled.

I am a strong proponent for 2.

src/UI/Root/Components/PrivateRoute/PrivateRoute.tsx
bartv: Do you have a reference somewhere to "outlet"? I have never heard that term in the context of software.

bartv (changes requested):

matborowczyk:
@bartv regarding outlet - https://reactrouter.com/en/main/components/outlet

bartv:
> @bartv regarding outlet - https://reactrouter.com/en/main/components/outlet

that is a weird name :-)

src/Data/Auth/Providers/AuthContext.ts
LukasStordeur: Yes, then you can make a ticket and reference this PR//Comment.

src/Data/Auth/KeycloakProvider.tsx
LukasStordeur: Moving an export to another file should not be triggering circular dependency. If you have a circular dep issue, you might have forgotten to remove the import statement if you moved it to the file where it's used or adjust it.

src/Slices/Settings/UI/Tabs/Tabs.tsx
matborowczyk: I followed here logic that was in the previous authHelper. I can add explicit method, but should the logic behind it stay the same?

src/Data/Auth/KeycloakProvider.tsx
LukasStordeur: yes, rename it please

src/Data/Managers/V2/helpers/useHelpers.ts
LukasStordeur: @bartv  the 'use' keyword is part of the react ecosystem and are used to identify hooks.

> Hooks are defined using JavaScript functions, but they represent a special type of reusable UI logic with restrictions on where they can be called.

Based on the definition of hooks, this helper doesn't need to be considered one I believe. `httpHelpers` or `fetchHelpers` might be more accurate.

src/Slices/Agents/UI/Agents.test.tsx
LukasStordeur: Okay, make a ticket for it. But I had hoped that the refactor would allow the auth to be standalone and thus meant that we could remove the complexer logic relying on that.

src/UI/Root/Components/PrivateRoute/PrivateRoute.tsx
LukasStordeur: RouterOutlet, or Outlet has its origins in Angular, but is also part of React-Router-Dom now, and the term has become more widely used in Frontend tech.  

An Outlet renders the matching child route of a parent route. 
https://reactrouter.com/en/main/components/outlet

src/Slices/Settings/UI/Tabs/Tabs.tsx
LukasStordeur: +1 on option 2

src/Data/Auth/AuthProvider.tsx
LukasStordeur: A full application in React could hypothetically be written in one single component, we are splitting it up for readability and organization, but technically speaking, everything that can be achieved with multiple components can be done in one. 

If you can't figure it out, which is fine, make a ticket for it, and at least, place all the elements/wrappers/parts concerning Keycloak in on single file. You should only call one Provider for Keycloak here.

matborowczyk: Of course it could be done, but then there would be no need for useContext. And as far as I know and searched context provider has to be one level higher that the component where we want to use given context.

I'll merge the files.

src/Data/Auth/KeycloakProvider.tsx
matborowczyk: it does, at least madge complains when the interface is in the authprovider, app and tests works just fine

LukasStordeur: I believe it's only used in the authProvider component, so either;
1.  don't export it and declare it as an interface inside AuthProvider.
2.  since we are using a switch case on the config.method, perhaps this distinction isn't needed anymore and we can go for a more generic type which has the different possibilities for '`method : "oidc" | "database" | "other" | undefined`

matborowczyk: It's reused, and in case of "oidc" the rest of the config is being passed to the keycloakProvider to create keycloak instance, I'll put it to the types.ts file to have it in "more correct" place

src/UI/Root/Root.test.tsx
matborowczyk: it all over the place as the commandResolvers require them, but I added it anyway, because I forgot it there

src/Data/Auth/AuthProvider.tsx
matborowczyk: changed, and removed url parameter

src/UI/Root/Root.tsx
```diff
@@ -1,18 +1,17 @@
...
-import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
 import { LoginPage } from "@/Slices/Login";
 import { DependencyContext } from "@/UI/Dependency";
 import { SearchSanitizer } from "@/UI/Routing";
 import { GlobalStyles } from "@/UI/Styles";
 import { NotFoundPage } from "@S/NotFound/UI";
-import { KeycloakProvider, PageFrame, Initializer } from "./Components";
+import { PageFrame } from "./Components";
+import RouteOutlet from "./Components/RouteOutlet/RouteOutlet";
```
LukasStordeur, line 10: Why is RouterOutlet not inside the Routing folder?

LukasStordeur (commented):
I'd like to see screenshots of the passing E2E tests for OSS, the keycloak and database flows if possible.

src/Slices/Agents/UI/Agents.test.tsx
matborowczyk: https://github.com/inmanta/web-console/issues/5840

LukasStordeur:
> I had to remove email, first name & last name input part from the keycloak scenario to make it working Screenshot 2024-06-21 at 15 14 09 Screenshot 2024-06-21 at 15 15 27

Okay, still would like the OSS tests later if you have the time :)

LukasStordeur:
image
oss-tests results

src/Data/Auth/types.ts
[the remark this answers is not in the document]
matborowczyk: I also don't understand why madge is raising this error, I'll add it to the task about investigation alternatives for madge

src/Data/Auth/Providers/KeycloakAuthProvider.tsx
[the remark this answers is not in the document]
matborowczyk: That wasn't working, that was my initial idea, as I wrote either here or through slack it will allow you to log in, and token can be accessed but for some reason it won't update user details so some parts of the ui would be like there is no auth at all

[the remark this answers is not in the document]
LukasStordeur: Can you name the inner provider something else than AuthProvider then? It's confusing.

src/UI/Root/Root.tsx
matborowczyk: changed

src/UI/Root/Components/Header/EnvSelector/EnvSelector.tsx
```diff
@@ -93,14 +82,14 @@ export const EnvSelector: React.FC<Props> = ({
               {words("home.navigation.button")}
             </DropdownItem>
           </Tooltip>
-          {authController.isEnabled() && (
+          {authHelper.getUser() && (
```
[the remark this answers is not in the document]
matborowczyk: changed

src/UI/Root/Components/Header/EnvSelector/EnvSelector.tsx
```diff
@@ -45,30 +44,20 @@ export const EnvSelector: React.FC<Props> = ({
           aria-label={toggleText}
           isFullHeight
           onClick={() => setIsOpen(!isOpen)}
-          icon={authController.isEnabled() ? <UserCircleIcon /> : null}
+          icon={authHelper.getUser() ? <UserCircleIcon /> : null}
```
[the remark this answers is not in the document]
matborowczyk: changed

src/Slices/ServiceInventory/UI/Spec/ServiceInventoryPrepper.tsx
```diff
@@ -99,19 +100,21 @@ export class ServiceInventoryPrepper {
...
-            commandResolver,
-            environmentModifier: new MockEnvironmentModifier(),
-            environmentHandler,
-          }}
-        >
-          <StoreProvider store={store}>
-            <ServiceInventory serviceName={service.name} service={service} />
-          </StoreProvider>
-        </DependencyProvider>
+        <AuthProvider config={undefined}>
```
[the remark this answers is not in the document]
matborowczyk: I follow what the real scenario would be, as config theoretically would be undefined, not null

src/Test/Inject/TestInjector.tsx
```diff
@@ -0,0 +1,41 @@
...
+ *     <AuthProvider config={config}>
+ *      <TestInjector dependencies={dependencies}>
+ *       <Login />
+ *     </TestInjector>
+ *    </AuthProvider>
+ *   );
+ * };
+ * ```
+ */
+export const TestInjector: React.FC<React.PropsWithChildren<Props>> = ({
```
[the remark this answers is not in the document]
matborowczyk: I'll rename it as I don't know in what direction we will go if it comes to refactor

src/Slices/Login/Page.tsx
```diff
@@ -15,33 +14,18 @@ import logo from "@images/logo.svg";
  * @returns {React.FunctionComponent} The rendered component.
  */
 export const Login: React.FunctionComponent = () => {
```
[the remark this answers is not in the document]
matborowczyk: Adjusted

src/Data/Managers/V2/helpers/useFetchHelpers.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  it("should return headers without Authorization Token when authController is disabled", () => {
+    const wrapper = setup();
+
+    const { result } = renderHook(() => useFetchHelpers().createHeaders(), {
+      wrapper,
+    });
+    expect(result.current.get("Authorization")).toEqual(null);
+  });
+
+  it("should return headers without Authorization Token when authHelper hook return undefined", () => {
```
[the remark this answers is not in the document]
matborowczyk: when there is no authentication, we don't set authorization token so the value will be undefined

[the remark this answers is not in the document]
LukasStordeur: getToken should return null instead of undefined, like we do everywhere else when instantiating a function like that. You do it for getUser, why make it different for getToken?

src/Slices/ServiceInventory/UI/ServiceInventory.test.tsx
```diff
@@ -111,6 +111,7 @@ function setup(service = Service.a, pageSize = "") {
   );
   const component = (
     <MemoryRouter initialEntries={[`/?env=aaa${pageSize}`]}>
+      {/* <AuthProvider config={undefined}> */}
```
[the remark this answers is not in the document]
matborowczyk: removed

src/Data/Common/CookieHelper.ts
```diff
@@ -34,7 +34,7 @@ export const getCookie = (cookieName: string): string | null => {
     //slice of the amount that equals to cookieName + '=' sign
     return searchedCookie.slice(cookieName.length + 1);
   } else {
-    return null;
+    return undefined;
```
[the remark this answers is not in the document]
matborowczyk: as I wrote below to keep it in line with keycloak.token, it felt natural to go like this, but I'll change that

src/Data/Auth/Providers/KeycloakAuthProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+   * Clear the token and initiate the Keycloak login flow.
+   */
+  const login = (): void => {
+    keycloak.clearToken();
+    keycloak.login();
+  };
+
+  /**
+   * Get the access token of the currently logged-in user.
+   * @returns The access token of the currently logged-in user, or undefined if no user is logged in.
```
LukasStordeur, line 41:
```suggestion
   * @returns {string|null} The access token of the currently logged-in user, or null if no user is logged in.
```

src/Data/Auth/Providers/KeycloakAuthProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * KeycloakProvider component provides authentication functionality using Keycloak, by using Instance provider by 3rd party library and passing it to  our auth context implementation.
+ * It's done in that manner as fetching the context has to be done one level lower that the provider itself.
+ */
+const KeycloakProvider: React.FC<React.PropsWithChildren> = ({ children }) => {
+  const { keycloak } = useKeycloak();
+  /**
+   * Get the username of the currently logged-in user.
+   * @returns The username of the currently logged-in user, or undefined if no user is logged in.
```
LukasStordeur, line 16:
```suggestion
   * @returns  {string|null} The username of the currently logged-in user, or null if no user is logged in.
```

src/Slices/Login/Page.tsx
```diff
@@ -1,47 +1,31 @@
...
 import { DependencyContext, words } from "@/UI";
 import { UserCredentialsForm } from "@/UI/Components/UserCredentialsForm";
 import logo from "@images/logo.svg";
 
 /**
  * Login component.
- * This component is responsible for rendering the login pop-up form and handling the login process
- * @note This is being used only when local authentication is enabled.
+ * This component is responsible for rendering the login page.
+ * @note This is being used only when databse authentication is enabled.
```
LukasStordeur, line 13:
```suggestion
 * @note This is being used only when database authentication is enabled.
```

src/Slices/Login/Page.tsx
```diff
@@ -15,33 +14,18 @@ import logo from "@images/logo.svg";
...
-  const [isLoginOpen, setIsLoginOpen] = useState(false);
   const { mutate, isError, error, isSuccess, isPending, data } = useLogin();
 
   useEffect(() => {
     if (isSuccess) {
-      createCookie("inmanta_user", data.data.token, 1);
-      //reload the page to avoid possible error view due to the 401 calls done in the background
-      navigate(0);
-      authController.setLocalUserName(data.data.user.username);
+      authHelper.updateUser(data.data.user.username, data.data.token);
```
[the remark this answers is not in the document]
LukasStordeur: hmm, okay, curious.

src/Slices/ServiceInventory/UI/Spec/ServiceInventoryPrepper.tsx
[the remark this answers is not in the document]
LukasStordeur: in these cases, the attribute is more often made optional and not passed at all when it's not set, but it's fine for now.

src/Slices/ServiceInventory/UI/Spec/ServiceInventoryPrepper.tsx
```diff
@@ -99,19 +100,21 @@ export class ServiceInventoryPrepper {
...
+              commandResolver,
+              environmentModifier: new MockEnvironmentModifier(),
+              environmentHandler,
+            }}
+          >
+            <StoreProvider store={store}>
+              <ServiceInventory serviceName={service.name} service={service} />
+            </StoreProvider>
+          </DependencyProvider>
+        </AuthProvider>{" "}
```
LukasStordeur, line 117:
```suggestion
        </AuthProvider>
```

src/UI/Root/Components/Header/EnvSelector/EnvSelector.tsx
```diff
@@ -45,30 +44,20 @@ export const EnvSelector: React.FC<Props> = ({
...
           isFullHeight
           onClick={() => setIsOpen(!isOpen)}
-          icon={authController.isEnabled() ? <UserCircleIcon /> : null}
+          icon={authHelper.isDisabled() ? null : <UserCircleIcon />}
         >
-          {authController.isEnabled() ? (
-            <StyledDiv>
+          <StyledDiv>
+            <div>
+              {authHelper.isDisabled() ? null : (
```
LukasStordeur, line 51: `{!authHelper.isDisabled() && <StyledText>{authHelper.getUser()}</StyledText>}`

src/UI/Root/Root.tsx
```diff
@@ -1,18 +1,17 @@
...
 import { Navigate, Route, Routes } from "react-router-dom";
-import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
 import { LoginPage } from "@/Slices/Login";
 import { DependencyContext } from "@/UI/Dependency";
 import { SearchSanitizer } from "@/UI/Routing";
 import { GlobalStyles } from "@/UI/Styles";
 import { NotFoundPage } from "@S/NotFound/UI";
-import { KeycloakProvider, PageFrame, Initializer } from "./Components";
+import RouteOutlet from "../Routing/RouteOutlet/RouteOutlet";
```
LukasStordeur, line 9: Why have a separate folder for RouteOutlet? 
And if you really want a separate folder, it should use an index file.

src/UI/Routing/RouteOutlet/RouteOutlet.tsx
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  return (
+    <Initializer>
+      <Outlet />
+    </Initializer>
+  );
+};
+
+export default RouteOutlet;
```
LukasStordeur, line 27: avoid export defaults

src/UI/Root/Components/Header/EnvSelector/EnvSelector.tsx
```diff
@@ -45,30 +44,20 @@ export const EnvSelector: React.FC<Props> = ({
...
           isFullHeight
           onClick={() => setIsOpen(!isOpen)}
-          icon={authController.isEnabled() ? <UserCircleIcon /> : null}
+          icon={!authHelper.isDisabled() ? <UserCircleIcon /> : null}
         >
-          {authController.isEnabled() ? (
-            <StyledDiv>
+          <StyledDiv>
+            <div>
+              {!authHelper.isDisabled() ? (
```
LukasStordeur, line 51: {!authHelper.isDisabled() && <StyledText>{authHelper.getUser()}</StyledText>}
      
There's no need for the ternary here. You changed the wrong line, there was nothing wrong with line 47.

src/UI/Routing/RouteOutlet/index.ts
LukasStordeur, line 1: I still don't see the point in having a specific folder for the RouteOutlet. But fine.

matborowczyk: That's the convention for all of the folders

LukasStordeur (approved):

bartv (approved):

PR #5816, opened 2024-06-12
Feature/5763 instance editor

### LINKED ISSUE #5763: Implement Editor to Form
Enhance the Forms to edit/create instances with the possibility to have a code editor like Monaco.

LukasStordeur:
# Description
 
Implementation for the React-Monaco-Editor. 
The form now supports toggling between the normal form, and the JSON-Editor with linting.
The form-toggle is only enabled when the editor contains parseable and valid JSON according to the provided JSON-schema.
Only caveat is for integers or floats when the value is empty, it is displayed/parsed as an empty string. The value will then be considered invalid integer.

The Editor doesn't render properly in JSDOM, Hence the coverage will be done in the E2E tests.

closes #5763 

https://github.com/inmanta/web-console/assets/44098050/54838fe8-7b34-4466-9da7-75c308d950ae


src/UI/Components/JSONEditor/JSONEditor.tsx
```diff
@@ -0,0 +1,146 @@
...
+   * @param markers - The markers from the monaco editor.
+   * It will set the errors state with the error messages.
+   * see : https://microsoft.github.io/monaco-editor/typedoc/interfaces/editor.IMarker.html
+   *
+   * @return {void}
+   */
+  const handleOnValidate: OnValidate = (markers) => {
+    if (markers.length === 0) {
+      setErrors([]);
+    }
```
matborowczyk, line 61: nothing big but we could do else there as if length is 0 then there is no markers to iterate through

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
```diff
@@ -150,18 +176,43 @@ export const ServiceInstanceForm: React.FC<Props> = ({
...
+        <ToggleGroupItem
+          text="Form"
+          key={0}
+          buttonId="formButton"
+          isSelected={isForm}
+          isDisabled={!isEditorValid}
+          onChange={() => setIsForm(true)}
+        />
+        <ToggleGroupItem
+          text="JSON-Editor"
```
matborowczyk, line 189: is that displayed? if so we can put it to the words.ts

src/UI/Components/JSONEditor/JSONEditor.tsx
LukasStordeur: done

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
LukasStordeur: done

matborowczyk (approved):

PR #5845, opened 2024-06-27
Issue/5818 jwt auth

### LINKED ISSUE #5818: New AUTH method
The goal is to add support for `jwt` authentication method. This method does not require the frontend to do anything except indicating that a user is logged in. The backend service handles the authentication.

The current user can be fetched using `/api/v2/current_user`   

It returns a 404 when it is not possible to determine which user is logged in.

```json
{
  "message": "Request or referenced resource does not exist: No current user found, probably an API token is used."
}
```

matborowczyk:
# Description

closes #5818 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/5818-jwt-auth.yml
+++ b/changelogs/unreleased/5818-jwt-auth.yml
@@ -1,0 +1,6 @@
+description: Refactor Authentication to simplify the flow and make it more maintainable and adjustable for new authentication providers
+issue-nr: 5759
+change-type: patch
+destination-branches: [master, iso7]
+sections:
+  minor-improvement: "{{description}}"
--- a/changelogs/unreleased/5818-jwt-auth.yml
+++ b/changelogs/unreleased/5818-jwt-auth.yml
@@ -1,5 +1,5 @@
-description: Refactor Authentication to simplify the flow and make it more maintainable and adjustable for new authentication providers
-issue-nr: 5759
+description: Add JWT authentication method to the application
+issue-nr: 5818
 change-type: patch
 destination-branches: [master, iso7]
 sections:
--- a/src/Data/Auth/Providers/JwtAuthProvider.tsx
+++ b/src/Data/Auth/Providers/JwtAuthProvider.tsx
@@ -1,10 +1,11 @@
 import React, { useContext, useEffect, useState } from "react";
 import { useGetCurrentUser } from "@/Data/Managers/V2/GetCurrentUser";
 import { AuthContext } from "../AuthContext";
+
 /**
- * Component that implements a null authentication provider when no authentication is enabled.
+ * Component that implements a authentication provider when jwt authentication is enabled.
  */
-export const JwtProvider: React.FC<React.PropsWithChildren> = ({
+export const JwtAuthProvider: React.FC<React.PropsWithChildren> = ({
   children,
 }) => {
   const [user, setUser] = useState<string | null>(null);
@@ -19,6 +20,7 @@ export const JwtProvider: React.FC<React.PropsWithChildren> = ({
       setUser(data?.username);
     }
   }, [data, isSuccess]);
+
   return (
     <AuthContext.Provider value={{ ...authContext, getUser, isDisabled }}>
       {children}
--- a/src/Data/Auth/Providers/JwtAuthProvider.tsx
+++ b/src/Data/Auth/Providers/JwtAuthProvider.tsx
@@ -4,6 +4,8 @@ import { AuthContext } from "../AuthContext";
 
 /**
  * Component that implements a authentication provider when jwt authentication is enabled.
+ * In practice it fetches username from the backend because whole authentication flow is managed by 3rd party outside of the scope of our application.
+ * User will open application already authorized and from our perspective the flow is almost identical as in case of no authorization.
  */
 export const JwtAuthProvider: React.FC<React.PropsWithChildren> = ({
   children,
--- a/src/Data/Auth/Providers/JwtProvider.tsx
+++ b/src/Data/Auth/Providers/JwtProvider.tsx
@@ -1,5 +1,5 @@
 import React, { useContext, useEffect, useState } from "react";
-import { useGetCurrentUser } from "@/Data/Managers/V2/RemoveUser";
+import { useGetCurrentUser } from "@/Data/Managers/V2/GetCurrentUser";
 import { AuthContext } from "../AuthContext";
 /**
  * Component that implements a null authentication provider when no authentication is enabled.
--- a/src/Data/Auth/Providers/JwtProvider.tsx
+++ b/src/Data/Auth/Providers/JwtProvider.tsx
@@ -1,0 +1,27 @@
+import React, { useContext, useEffect, useState } from "react";
+import { useGetCurrentUser } from "@/Data/Managers/V2/RemoveUser";
+import { AuthContext } from "../AuthContext";
+/**
+ * Component that implements a null authentication provider when no authentication is enabled.
+ */
+export const JwtProvider: React.FC<React.PropsWithChildren> = ({
+  children,
+}) => {
+  const [user, setUser] = useState<string | null>(null);
+  const { data, isSuccess } = useGetCurrentUser().useOneTime();
+  const authContext = useContext(AuthContext);
+
+  const getUser = (): string | null => user;
+  const isDisabled = () => true;
+
+  useEffect(() => {
+    if (isSuccess) {
+      setUser(data?.username);
+    }
+  }, [data, isSuccess]);
+  return (
+    <AuthContext.Provider value={{ ...authContext, getUser, isDisabled }}>
+      {children}
+    </AuthContext.Provider>
+  );
+};
--- a/src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
+++ b/src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
@@ -2,6 +2,11 @@ import { useQuery } from "@tanstack/react-query";
 import { PrimaryBaseUrlManager } from "@/UI";
 import { useFetchHelpers } from "../helpers";
 
+interface LoggedUser {
+  data: {
+    username: string;
+  };
+}
 /**
  * Custom hook for getting a current user from the server.
  *
@@ -16,11 +21,7 @@ export const useGetCurrentUser = () => {
 
   const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
 
-  const currentUserOrder = async (): Promise<{
-    data: {
-      username: string;
-    };
-  }> => {
+  const currentUserOrder = async (): Promise<LoggedUser> => {
     const response = await fetch(baseUrl + `/api/v2/current_user/`, {
       headers: createHeaders(),
     });
@@ -33,7 +34,7 @@ export const useGetCurrentUser = () => {
   return {
     /**
      * Custom hook to fetch the user information from the API once.
-     * @returns The result of the query, including the current user information.
+     * @returns {UseQueryResult<LoggedUser, Error>} The result of the query, including the current user information.
      */
     useOneTime: () =>
       useQuery({
--- a/src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
+++ b/src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
@@ -3,9 +3,9 @@ import { PrimaryBaseUrlManager } from "@/UI";
 import { useFetchHelpers } from "../helpers";
 
 /**
- * Custom hook for removing a user from the server.
+ * Custom hook for getting a current user from the server.
  *
- * @returns {Mutation} - The mutation object provided by `useMutation` hook.
+ * @returns {Query} - An object containing a custom hook to fetch user information.
  */
 export const useGetCurrentUser = () => {
   const { createHeaders, handleErrors } = useFetchHelpers();
@@ -16,12 +16,6 @@ export const useGetCurrentUser = () => {
 
   const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
 
-  /**
-   * Deletes a user from the server.
-   *
-   * @param {string} username - The username of the user to be removed.
-   * @returns {Promise<{}>} - A promise that resolves when the user is successfully removed.
-   */
   const currentUserOrder = async (): Promise<{
     data: {
       username: string;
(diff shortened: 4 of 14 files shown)


src/Data/Auth/Providers/JwtProvider.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useContext, useEffect, useState } from "react";
+import { useGetCurrentUser } from "@/Data/Managers/V2/GetCurrentUser";
+import { AuthContext } from "../AuthContext";
+/**
+ * Component that implements a null authentication provider when no authentication is enabled.
+ */
+export const JwtProvider: React.FC<React.PropsWithChildren> = ({
```
LukasStordeur, line 7: JwTAuthProvider, all others contain the word auth in there.

src/Data/Auth/Providers/JwtProvider.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useContext, useEffect, useState } from "react";
+import { useGetCurrentUser } from "@/Data/Managers/V2/GetCurrentUser";
+import { AuthContext } from "../AuthContext";
+/**
+ * Component that implements a null authentication provider when no authentication is enabled.
```
LukasStordeur, line 5: update docstring please

src/Data/Auth/Providers/JwtProvider.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useContext, useEffect, useState } from "react";
+import { useGetCurrentUser } from "@/Data/Managers/V2/GetCurrentUser";
+import { AuthContext } from "../AuthContext";
+/**
```
LukasStordeur, line 4: missing new line before comment

src/Data/Auth/Providers/JwtProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  const getUser = (): string | null => user;
+  const isDisabled = () => true;
+
+  useEffect(() => {
+    if (isSuccess) {
+      setUser(data?.username);
+    }
+  }, [data, isSuccess]);
+  return (
```
LukasStordeur, line 22: new line before return
```suggestion
  
  return (
```

src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
```diff
@@ -0,0 +1,51 @@
+import { useQuery } from "@tanstack/react-query";
+import { PrimaryBaseUrlManager } from "@/UI";
+import { useFetchHelpers } from "../helpers";
+
+/**
+ * Custom hook for removing a user from the server.
+ *
+ * @returns {Mutation} - The mutation object provided by `useMutation` hook.
+ */
+export const useGetCurrentUser = () => {
```
LukasStordeur, line 10: Why is it calle Get when the docstrings tell it's for removing a user from the server?

src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
```diff
@@ -0,0 +1,51 @@
...
+  const { createHeaders, handleErrors } = useFetchHelpers();
+  const baseUrlManager = new PrimaryBaseUrlManager(
+    globalThis.location.origin,
+    globalThis.location.pathname,
+  );
+
+  const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
+
+  /**
+   * Deletes a user from the server.
```
LukasStordeur, line 20: ??

src/UI/Root/Components/Header/EnvSelector/EnvSelector.tsx
```diff
@@ -44,11 +44,11 @@ export const EnvSelector: React.FC<Props> = ({
           aria-label={toggleText}
           isFullHeight
           onClick={() => setIsOpen(!isOpen)}
-          icon={!authHelper.isDisabled() ? <UserCircleIcon /> : null}
+          icon={!authHelper.getUser() ? <UserCircleIcon /> : null}
         >
           <StyledDiv>
             <div>
-              {!authHelper.isDisabled() && (
+              {!authHelper.getUser() && (
```
LukasStordeur, line 51: checking if you don't have a user to then call the user, doesn't make sense at all.

LukasStordeur (changes requested):
I don't see any tests mocking the situation, or testing the provider itself.
Nearly all your docstrings are copy-pasted, can you please read through them and update them?

changelogs/unreleased/5818-jwt-auth.yml
```diff
@@ -0,0 +1,6 @@
+description: Refactor Authentication to simplify the flow and make it more maintainable and adjustable for new authentication providers
```
LukasStordeur, line 1: ? This is the wrong changelog, please update

src/Data/Auth/Providers/JwtAuthProvider.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useContext, useEffect, useState } from "react";
+import { useGetCurrentUser } from "@/Data/Managers/V2/GetCurrentUser";
+import { AuthContext } from "../AuthContext";
+
+/**
+ * Component that implements a authentication provider when jwt authentication is enabled.
```
LukasStordeur, line 6: Might be good to add some more explanation to it? 
Something in the genre of what Bart described?
if the method is jwt you assume that a user is logged in and you show the username by calling the current_user api endpoint

matborowczyk: elaborated a bit more on it, I think now it's clear

src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    await handleErrors(response);
+
+    return response.json();
+  };
+
+  return {
+    /**
+     * Custom hook to fetch the user information from the API once.
+     * @returns The result of the query, including the current user information.
```
LukasStordeur, line 36: @returns UseQueryResult<LoggedUser/CurrentUser, Error> The result of the query, including the current user information.

There's typing available for the queries. It would be good to have a type for the user-data, even if it's just the username.

src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
```diff
@@ -0,0 +1,45 @@
...
+  const { createHeaders, handleErrors } = useFetchHelpers();
+  const baseUrlManager = new PrimaryBaseUrlManager(
+    globalThis.location.origin,
+    globalThis.location.pathname,
+  );
+
+  const baseUrl = baseUrlManager.getBaseUrl(process.env.API_BASEURL);
+
+  const currentUserOrder = async (): Promise<{
+    data: {
```
LukasStordeur, line 20: Why not a type?

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
```diff
@@ -131,3 +135,62 @@ test("GIVEN EnvironmentSelector and environments with identical names WHEN user
 
   expect(selectedEnv).toEqual(envB.id);
 });
+
+const setup = (config: KeycloakAuthConfig | LocalConfig | undefined) => {
```
LukasStordeur, line 139: Why not declare this at the top of the file and reuse the setup for the other tests too?

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
```diff
@@ -131,3 +135,62 @@ test("GIVEN EnvironmentSelector and environments with identical names WHEN user
...
+  server.listen();
+  render(setup({ method: "jwt" }));
+
+  await waitFor(() => {
+    expect(screen.getByText("test_user")).toBeVisible();
+  });
+  server.close();
+});
+
+test("GIVEN EnvironmentSelector WHEN jwt auth is enabled will won't fetched username on load", async () => {
```
LukasStordeur, line 179: The test case has not the jwt auth enabled, it has the database auth enabled.

The text says you'd check whether the user is fetched or not, but you can't exactly test that because you can add as many handlers to your `setupServer()` method, if the call is not triggered it simply won't fire the handler. You can see them as eventlisteners, but for api-calls.

Also, considering you're setting up a database auth, this means the you'd expect the text in the env selector to be something like no user, or null, or whichever the fallback is in that scenario. 

This test is for me a false positive and doesn't add anything to the coverage.

You'd better test a failing path, where the api endpoint returns no user. A case that should result in a failure, since this is not supposed to happen.

LukasStordeur (changes requested):

src/UI/Root/Components/Header/EnvSelector/EnvSelector.tsx
```diff
@@ -44,11 +44,11 @@ export const EnvSelector: React.FC<Props> = ({
           aria-label={toggleText}
           isFullHeight
           onClick={() => setIsOpen(!isOpen)}
-          icon={!authHelper.isDisabled() ? <UserCircleIcon /> : null}
+          icon={!authHelper.getUser() ? <UserCircleIcon /> : null}
```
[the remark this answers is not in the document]
LukasStordeur: I'm not sure I understand your suggestion/comment

src/Data/Auth/AuthProvider.test.tsx
LukasStordeur, line 1: You're build is failing on this file, the test file is empty

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
matborowczyk: changed

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
matborowczyk: added failing path

src/Data/Auth/AuthProvider.test.tsx
matborowczyk: deleted

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
```diff
@@ -1,13 +1,45 @@
...
+import { setupServer } from "msw/node";
+import { FlatEnvironment, RemoteData } from "@/Core";
+import { AuthProvider, KeycloakAuthConfig, LocalConfig } from "@/Data";
+import { AuthTestWrapper, Environment, dependencies } from "@/Test";
 import { DependencyProvider } from "@/UI/Dependency";
+import ErrorBoundary from "@/UI/Utils/ErrorBoundary";
 import { EnvSelectorWithData as EnvironmentSelector } from "./EnvSelectorWithData";
+import { EnvironmentSelectorItem } from "./EnvSelectorWrapper";
+
+const setup = (
```
LukasStordeur, line 18: There are arguments that can have defaults to avoid ending up with so many anonymous arguments down in the tests. And if you organise the order a bit, you'd have minimal arguments to pass to override the variables.

```ts
const setup = (
 // I placed this one first since it's the one that changes the most.  also, avoid passing it as an anonymous function
// it makes it hard to read months later. 
  callback?: (item: EnvironmentSelectorItem) => void,
  // Default it to undefined, this avoids the need to pass it as undefined when calling the method.
// I think you can omit setting the default altogether if the argument is optional... not sure right now, but you get the drift.
  config?: KeycloakAuthConfig | LocalConfig = undefined,
  // these two below are always the same, so you could consider removing them from the arg list and have them declared in your method directly. Or have defaults like these if you feel like they might change or be overridden.
  environments?: FlatEnvironment[] = Environment.filterable,  
  selectedEnv?: FlatEnvironment | undefined = envA, 
) => {
  const queryClient = new QueryClient();
  return (
    <ErrorBoundary>
      <MemoryRouter>
        <QueryClientProvider client={queryClient}>
          <AuthProvider config={config}>
            <AuthTestWrapper dependencies={dependencies}>
              <EnvironmentSelector
                environments={RemoteData.success(environments)}
                onSelectEnvironment={callback || () => {}}
                selectedEnvironment={selectedEnv}
              />
            </AuthTestWrapper>
          </AuthProvider>
        </QueryClientProvider>
      </MemoryRouter>
    </ErrorBoundary>
  );
};
```

matborowczyk: changed

src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    return response.json();
+  };
+
+  return {
+    /**
+     * Custom hook to fetch the user information from the API once.
+     * @returns {UseQueryResult<LoggedUser, Error>} The result of the query, including the current user information.
+     */
+    useOneTime: () =>
```
LukasStordeur, line 39: typing.
This is how I added the correct typing
 useOneTime: (): UseQueryResult<ServiceInstanceModel, Error> =>

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
LukasStordeur: sorry if I'm being annoying, but callback doesn't mean much on what it's a callback for. Can you adjust the naming of the method? onSelectEnvironment or something similar?
 Can you also name the anonymous methods in the places where you use that argument ?

matborowczyk: No worries, you are not! renamed and added names for callback functions

src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
matborowczyk: changed, I had to modify promise above though, because select property influence typing of the UseQueryResult

src/Data/Managers/V2/GetCurrentUser/useGetCurrentUser.ts
```diff
@@ -0,0 +1,44 @@
+import { UseQueryResult, useQuery } from "@tanstack/react-query";
+import { PrimaryBaseUrlManager } from "@/UI";
+import { useFetchHelpers } from "../helpers";
+
+interface LoggedUser {
+  username: string;
+}
+/**
+ * Custom hook for getting the current logged in user from the server.
```
LukasStordeur, line 9: Can we be clearer here, and specify those are React Queries?

matborowczyk: updated all hooks

LukasStordeur (approved):

PR #5854, opened 2024-07-01
Block editing instances with strict_modifier in the composer 

### LINKED ISSUE #5343: Review functionality for deleting instances
Regarding conversation [here](https://github.com/inmanta/web-console/pull/5323). 

In short: Composer can't edit instances that aren't in strict mode, so I disabled editing instances, as well as deleting connection between its embedded instances, and also deleting instances as well. 

Later it was noted that deleting is actually acceptable, but in current state it could be possible for a user to delete, for example, one embedded entity and leave the rest of the instance which would cause in rejected query, as that instance isn't in "strict mode", and that operation basically is attempt to edit it.

I proposed that we could hold with showing option to  delete instance that aren't in "strict mode" and introduce in separate task logic that would prompt user that given part of the instance isn't in strict mode and he can delete all of components that build given instance 

(Note: check if it's possible to add new embedded entity)

matborowczyk:
# Description

Review after merging pull request #5672 - I based that branch on it to avoid unneccesary conflicts

closes #5343 

# Self Check:

- [ ] Changelog entry


matborowczyk:
I'll reopen after the merge

PR #5893, opened 2024-07-23
Issue/5779 add documentation tab

### LINKED ISSUE #5779: [Service Inventory V2] Add Documentation tab
https://internal.inmanta.com/development/frontend/designs/Service-Inventory-2024.html

Implement the Documentation tab.

LukasStordeur:
# Description

Added the Documentation tab to the view, and added the disabling for the Resource tab whenever we are not on the latest tab.

closes #5779 

image

> Note: The title of the sections is provided by the backend, in the case of our local-setup, the attribute with documentation happens to be labeled Documentation, hence the double appearances of the word Documentation in the tab, and in the sections.


src/Slices/ServiceInstanceDetails/Test/Page.test.tsx
```diff
@@ -202,9 +203,108 @@ describe("ServiceInstanceDetailsPage", () => {
...
+    const component = setup();
+    render(component);
+
+    expect(
+      await screen.findByRole("region", { name: "Instance-Details-Success" }),
+    ).toBeInTheDocument();
+
+    // active attribute set
+    // in this version, we should have documentation available
+    expect(screen.getByText("Latest Version")).toBeInTheDocument();
```
matborowczyk, line 227: I know that is a minor difference, but wouldn't `toBeVisible()` better? Theoretically element could be in the document, but not visible. As we are using patternfly I assume that is very unlikely so it's just a suggestion

src/Slices/ServiceInstanceDetails/UI/Tabs/DocumentationTabContent.tsx
```diff
@@ -0,0 +1,169 @@
...
+interface Props {
+  docAttributeDescriptors: DocAttributeDescriptors[];
+  selectedVersion: string;
+}
+
+/**
+ * The DocumentationTabContent Component
+ *
+ * @props {Props} - The component props
+ *  @prop {DocAttributeDescriptors} docAttributeDescriptors -  array of Object containing the needed info for each sections
```
matborowczyk, line 40:
```suggestion
 *  @prop {DocAttributeDescriptors} docAttributeDescriptors -  array of Objects containing the needed info for each sections
```

matborowczyk (approved):

src/Slices/ServiceInstanceDetails/Test/Page.test.tsx
LukasStordeur: I can change it, but since those are not elements are either present or not in the DOM, I opted for this assertion.

LukasStordeur: `toBeVisible` has an additional check for `display:none` compared to `toBeInTheDocument`

PR #5913, opened 2024-08-20
Add navigator to the instance composer

### LINKED ISSUE #5871: [Instance Composer] Update Zooming functionality
### ❗❗❗**Important : this is an ISO 8 feature**

This model is based on the examples available in the JointJS library. 
It has three functionalities: 
- Fullscreen mode => the icon shifts from outwards to inwards, like in most cases where a fullscreen icon is used.
- Fit to Screen, this resets the view and makes everything fit into the view.
- Adjust the zoom level.

Image

matborowczyk:
# Description

Just in case, the cut-of at the full-screen was due to the recording dimensions. 
https://github.com/user-attachments/assets/bf8eb0ac-60b5-4085-a381-6cffbd93bbcd

closes #5871

# Self Check:

- [ ] Changelog entry


src/UI/Components/Diagram/navigator/navigator.ts
```diff
@@ -0,0 +1,266 @@
...
+  setTooltip(tooltip: string): void;
+}
+
+/**
+ * IconButton class
+ *
+ * This class extends the ui.widgets.button class and represents a button with an icon.
+ * It provides methods for setting the icon and the tooltip of the button.
+ *
+
```
LukasStordeur, line 22: remove unneeded newline

src/UI/Components/Diagram/navigator/navigator.ts
```diff
@@ -0,0 +1,266 @@
...
+ * This class is responsible for managing the navigation of the canvas.
+ * It provides methods for zoom to fit, zooming in/out in the canvas and fullscreen toggle.
+ *
+ * @class
+ * @property {ui.Toolbar} toolbar - The toolbar object that contains the navigation controls.
+ * @method constructor - Constructs a new instance of the NavigatorService class.
+ * @param {HTMLElement} element - The HTML element that will contain the navigator.
+ * @param {ui.PaperScroller} scroller - The scroller object that allows and zooming in the canvas.
+ */
+export class NavigatorService {
```
LukasStordeur, line 76: I'd align the naming with the component, so ZoomHandlerService.
Navigator feels like either routing, or being on the wide open sea and reading a map 😆 
(also don't forget to rename the files)

src/UI/Components/Diagram/navigator/navigator.ts
```diff
@@ -0,0 +1,266 @@
...
+          tooltip: "Fit to screen",
+        },
+        {
+          id: "zoomSlider",
+          type: "zoom-slider",
+          min: 0.2 * 100,
+          max: 5 * 100,
+          attrs: {
+            input: {
+              "data-tooltip": "Slide to zoom",
```
LukasStordeur, line 105: can you use the words dict?

src/UI/Components/Diagram/navigator/navigator.ts
```diff
@@ -0,0 +1,266 @@
...
+    fullscreenButton.setIcon(`${fitToScreen}`);
+
+    this.toolbar.on("fit-to-screen:pointerclick", () => this.fitToScreen());
+    this.toolbar.on("fullscreen:pointerclick", () => this.toggleFullscreen());
+
+    document.addEventListener("fullscreenchange", () =>
+      this.updateFullscreenStyling(),
+    );
+
+    document
```
LukasStordeur, line 141: This block here could benefit being split up.
assign the element to a variable and then only add the eventlistener if there's an element.

Also, there's this thing about eventListeners, and anonymous functions, but it might not be applicable in your case.
When you add an anonymous function as eventListener, you can't remove it explicitly, and it will reallocate extra memory on each run because it will redeclare the method. 

Also, you might want to bind your eventHandler method to the context of your class if you want to call it correctly in your addEventListener (It's been a while since I needed that, so I'm not certain)

This is an example on how to bind your functions ;
```ts
class MyClass {
  constructor() {
    this.name = "MyClass Instance";
    this.handleClick = this.handleClick.bind(this); // Bind 'this' to the method
  }

  handleClick(event) {
    console.log(this.name); // 'this' refers to the MyClass instance, not the element
  }

  addEventListenerToElement() {
    const button = document.getElementById('myButton');
    button.addEventListener('click', this.handleClick);
  }
}
```

src/UI/Components/Diagram/navigator/navigator.ts
```diff
@@ -0,0 +1,266 @@
...
+          return;
+        }
+        const slider = event.target as HTMLInputElement;
+
+        this.updateSliderProgressBar(slider);
+      });
+  }
+
+  /**
+   * Fits all of the canvas'es elements to the screen.
```
LukasStordeur, line 154:
```suggestion
   * Fits all of the canvas's elements to the screen.
```

src/UI/Components/Diagram/stencil/stencil.ts
```diff
@@ -163,18 +167,18 @@ export class StencilSidebar {
     serviceInventories: Inventories,
     service: ServiceModel,
   ) {
-    const instanceTab = new InstanceStencilTab(
+    this.instanceTab = new InstanceStencilTab(
```
LukasStordeur, line 170: What lead you to need `this` ?

src/UI/Root/Components/Header/Header.tsx
```diff
@@ -22,17 +22,35 @@ import { DocumentationLinks } from "./Actions/DocumentationLinks";
...
   onNotificationsToggle(): void;
 }
 
+/**
+ * Header component of the application.
+ *
+ * This component is responsible for rendering the header of the application.
+ *
+ * @component
+ * @param {Props} props - The properties that define the behavior of the header.
```
LukasStordeur, line 43: there's an 
`@props `
and `@prop `
that you can use
we might want to make a decision on this to have everything consistent

LukasStordeur (commented):
I can already see you did an effort on the readability, it is appreciated. 

I don't see any tests though, no unit or E2E. 

and you forgot the changelog file

src/UI/Components/Diagram/stencil/stencil.ts
matborowczyk: Added `remove()` function for the cleanup to the useEffect,  and to do so I had to have the reference outside from constructor

src/UI/Components/Diagram/navigator/navigator.ts
matborowczyk: added split and binds, as well as removing events at `remove()`

src/UI/Components/Diagram/zoomHandler/zoomHandler.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const zoomHandler = new ZoomHandlerService(zoom, scroller);
+
+  wrapper.appendChild(zoom);
+  document.body.appendChild(zoom);
+  it("should render zoom handler", () => {
+    expect(screen.getByTestId("zoomHandler")).toBeInTheDocument();
+  });
+
+  it("should fire requestFullscreen9() function when clicking fullscreen button", async () => {
```
LukasStordeur, line 32:
```suggestion
  it("should fire requestFullscreen() function when clicking fullscreen button", async () => {
```

LukasStordeur (approved):
just a typo, otherwise, good to go

PR #5966, opened 2024-09-24
Issue/5859 instance details actions

### LINKED ISSUE #5859: [Service Inventory V2] Implement Instance Actions  component
Image

LukasStordeur:
# Description

Implementation of the different actions that can be performed on an instance, including the expert actions. 
The attributesTab > JSON view will have a Force Update button in expert mode, and an expert action menu is available with the destroy action and force state action.

There is a known issue with the expert modal. When closing the modal, the dropdown of the expert action menu remains open. 

Closes #5859


cypress/e2e/scenario-2.4-new-expert-mode.cy.js
matborowczyk, line 1: suggestion, I would change file names to expert-mode here and old-expert-mode in the file to be removed, but it's just a matter of preference

cypress/e2e/scenario-2.4-new-expert-mode.cy.js
```diff
@@ -0,0 +1,266 @@
...
+      ).as("GetServiceInventory");
+
+      cy.get('[aria-label="Environment card"]')
+        .contains("lsm-frontend")
+        .click();
+      cy.get(".pf-v5-c-nav__item").contains("Service Catalog").click();
+      cy.get("#basic-service").contains("Show inventory").click();
+
+      // Make sure the call to get inventory has been executed
+      cy.wait("@GetServiceInventory");
```
matborowczyk, line 88: I forgot the reason behind awaiting the api call.
Shouldn't awaiting for the` ServiceInventory-<State>` would be enough and more in line with e2e approach to imitate how to user would behave?

src/Data/Managers/V2/DELETE/DeleteInstance/useDeleteInstance.ts
matborowczyk, line 1: I think it would be good to remove old(V1) queries if we have V2 queries unless it would be too big to handle it here and require separate PR. If that's the case then we should keep track on these.

src/Data/Managers/V2/POST/PostExpertStateTransfer/usePostExpertStateTransfer.ts
```diff
@@ -0,0 +1,65 @@
...
+  current_version: ParsedNumber;
+  target_state: string;
+  operation?: string;
+}
+
+/**
+ * React Query to Force update the state of an instance in expert mode.
+ *
+ * @param {string} environment - The Environment where the instance is located
+ * @param {string} instance_id - The hashed id of the instance
```
matborowczyk, line 21: it is hashed? I just don't know and I'm surprised to see that here

src/Slices/ServiceInstanceDetails/Test/AttributesUtils.test.ts
matborowczyk, line 1: I think it's better to have test next to the implementation, especially for the reviews, to avoid jumping around the files.

I would use Test folder for test related utils, mocks, setups etc.

src/Slices/ServiceInstanceDetails/Test/AttributesUtils.test.ts
```diff
@@ -0,0 +1,407 @@
...
+            name: "child2",
+            value: 456,
+            children: [],
+          },
+        ],
+      },
+    ]);
+  });
+
+  it("should format an array of objects correctly", () => {
```
matborowczyk, line 193: I'm not 100% sure but maybe test scenario for nested array of array of objects would be good as well, if we are covering objects, nested objects and deeply nested objects.

src/Slices/ServiceInstanceDetails/UI/Components/ToastAllert.tsx
```diff
@@ -0,0 +1,46 @@
...
+  id,
+  setMessage,
+  variant,
+}) => {
+  return (
+    <AlertGroup isToast={true}>
+      <Alert
+        variant={variant}
+        title={message}
+        data-testid={`${id}-error-message`}
```
matborowczyk, line 36: minor thing - if that's a general alert then I would switch `error` for alert or set conditional value here

src/Slices/ServiceInstanceDetails/Utils/StateUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+  serviceEntity?: ServiceModel,
+): string[] => {
+  if (!serviceEntity) {
+    return [];
+  }
+
+  // filter out the possible transfer objects that have the same source as current state.
+  const possibleStatesTransfers = serviceEntity.lifecycle.transfers;
+
+  // we are making a set because there can be duplicate entries in the possibleStateTransfers array
```
matborowczyk, line 77: Note, shouldn't be that addressed by core, It doesn't make sense to have duplicates values there, but maybe I don't know something

src/Slices/ServiceInstanceDetails/Test/StateTansferUtils.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    expect(result).toEqual([
+      "update_acknowledged",
+      "update_acknowledged_failed",
+    ]);
+  });
+});
+
+describe("getExpertStateTargets", () => {
+  it("should return an empty array if serviceEntity is undefined", () => {
+    const result = getExpertStateTargets(undefined);
```
matborowczyk, line 73: Why there could be undefined serviceEntity?

src/Slices/ServiceInstanceDetails/Test/InstanceActions.test.tsx
```diff
@@ -0,0 +1,480 @@
...
+    ).toBeVisible();
+
+    await act(async () => {
+      const confirmButton = screen.getByRole("button", {
+        name: /yes/i,
+      });
+
+      await userEvent.click(confirmButton);
+    });
+
```
matborowczyk, line 186: There could be assertion that the instance got deleted, currently we only know that dialog got closed and there is no errors

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesEditor.tsx
```diff
@@ -48,8 +77,63 @@ export const AttributesEditor: React.FC<Props> = ({
     setSelectedSet(value);
   };
 
+  /**
+   * useCallback hook onEditorUpdate
```
matborowczyk, line 81: We could make a note when we are using callbacks, it's the performance optimization?

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesEditor.tsx
```diff
@@ -48,8 +77,63 @@ export const AttributesEditor: React.FC<Props> = ({
...
-    setEditorData(JSON.stringify(attributeSets[selectedSet], null, 2));
+    if (isError) {
+      setErrorMessage(error.message);
+    }
+    if (isSuccess) {
+      setIsModalOpen(false);
+    }
+  }, [isSuccess, isError, error]);
+
+  useEffect(() => {
```
matborowczyk, line 135: I think that could be moved to value calculated on render according to new dev docs of react where they encourage to revisit useEffects and not to overuse them - [link](https://react.dev/learn/you-might-not-need-an-effect)

I'm not familiar with the flow and it could be that changing that is not that simple, then it's fine for me

src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
```diff
@@ -0,0 +1,138 @@
...
+    setIsModalOpen(false);
+    setInterfaceBlocked(false);
+    onClose();
+  }, [setIsModalOpen, setInterfaceBlocked, onClose]);
+
+  /**
+   * async method sending out the request to destroy the instance
+   */
+  const onSubmitDestroy = async (): Promise<void> => {
+    mutate("");
```
matborowczyk, line 81: I remember you talking that you couldn't make mutate with no arguments, could you point where you found that info/example? I found a issue from 2022 where someone from RQ team said it should be possible

src/Data/Managers/V2/DELETE/DeleteInstance/useDeleteInstance.ts
LukasStordeur: I kept it separate, considering I didn't know which actions would be kept in the original menu yet. And also to limit the scope of the PR considering it's size.

cypress/e2e/scenario-2.4-new-expert-mode.cy.js
LukasStordeur: Either solutions are fine, this is what we had in the other tests, so I stayed in line, but it's something we can change in a future iteration to align our E2E tests for example

src/Data/Managers/V2/POST/PostExpertStateTransfer/usePostExpertStateTransfer.ts
LukasStordeur: I will change it to UUID, but yes, a UUID is hashed.

src/Slices/ServiceInstanceDetails/Test/AttributesUtils.test.ts
LukasStordeur: If we have a Test folder, then I'd expect all tests to be in that folder. I guess this is a debate on preferences that we can take up in a frontend planning meeting.
To have them open aside eachother, you can always open two windows and keep them side by side.

src/Slices/ServiceInstanceDetails/Test/AttributesUtils.test.ts
LukasStordeur: As you can see in the matchObject, this test covers deeply nested elements, where we have a list of embedded entities, which contain on themselves a extra level of nested elements

src/Slices/ServiceInstanceDetails/Utils/StateUtils.ts
LukasStordeur: No, this is not a Core issue. 
Answered in code

src/Slices/ServiceInstanceDetails/Test/StateTansferUtils.test.ts
LukasStordeur: when the query is pending, it can happen that it is briefly undefined
This is mentioned in the docstrings of the method.

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesEditor.tsx
LukasStordeur: This is because the method is used in a UseEffect dependency, and would cause too many rerenders if not wrapped in a useCallback hook. This is a React practice, and gets flagged by the linting if ommited. I will add this comment in the code, but this is general knowledge about React.

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesEditor.tsx
LukasStordeur: If we keep it on initial calculation, the data will never be updated in the editor when the set is changed, or the version. For this specific case, there's no way around it.

src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DeleteAction.tsx
```diff
@@ -0,0 +1,128 @@
...
+      closeModal();
+    }
+  }, [isError, isSuccess, error, closeModal]);
+
+  return (
+    <>
+      <DropdownItem
+        isDisabled={isDisabled}
+        isDanger
+        key="delete"
```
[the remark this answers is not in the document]
LukasStordeur: Because the key attribute is required on all DropdownItems

src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
LukasStordeur: It will work if empty, but the TS linter will not accept it

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesEditor.tsx
matborowczyk: I don't know the case here, but they mention that if withouth use callback it won't work that means there is some underlying issue, but it couldn't be the case here so I asked

src/Slices/ServiceInstanceDetails/UI/Components/ToastAllert.tsx
matborowczyk, line 1: Name has typo :D

matborowczyk (approved):
It's only that namefile typo but I think it's good to go.

PR #6019, opened 2024-10-29
Issue/5986 header colors

### LINKED ISSUE #5986: Unify colors of the elements
There is a color mismatch and dragged stencil element has no colors

matborowczyk:
# Description

https://github.com/user-attachments/assets/22615d5b-543a-4279-8bd6-5ab160631e7b

closes #5986

# Self Check:

- [ ] Changelog entry


src/UI/Components/Diagram/shapes.ts
```diff
@@ -277,17 +301,42 @@ export class ServiceEntityBlock extends shapes.standard.HeaderedRecord {
...
-    const currentClassName = this.attr(["header", "class"]);
-
-    return this.attr(["header", "class"], `${currentClassName} -${color}`);
+  /**
+   * Sets the tab color based on the entity type.
+   *
+   * @param {EntityType} type - The type of the entity.
+   * @returns {this} updated entity block
+   */
+  setTabColor(type: EntityType): this {
```
LukasStordeur, line 319: Why do you need to return "this" when setting an attribute? 🤔 
I think it would be valuable to have the colors in a type, which reflects the EntityType

src/UI/Components/Diagram/shapes.ts
```diff
@@ -297,13 +346,28 @@ export class ServiceEntityBlock extends shapes.standard.HeaderedRecord {
...
 
-  editColumns(data: Array<ColumnData>, shouldBeCollapsed = true) {
+  /**
+   * Edits the columns of the entity and optionally sets the collapsed state.
+   *
+   * @param {Array<ColumnData>} data - The array of column data to set.
+   * @param {boolean} [shouldBeCollapsed=true] - Flag indicating whether the entity should be collapsed.
+   * @returns {this} The updated entity block.
+   */
+  editColumns(data: Array<ColumnData>, shouldBeCollapsed = true): this {
```
LukasStordeur, line 356: Wouln't update be the better term? You're using it in the return description too. Edit is for me a bit different than updating.

src/UI/Components/Diagram/shapes.ts
matborowczyk: It's the convention that they had on the demo that the actions from the event return updated entity, but I can remove that, as I don't think we are using it

LukasStordeur: It was just a question, maybe just add it in the comment.

src/UI/Components/Diagram/shapes.ts
```diff
@@ -23,17 +25,17 @@ export class ServiceEntityBlock extends shapes.standard.HeaderedRecord {
...
           },
           header: {
-            class: "joint-entityBlock-header",
-            strokeWidth: 1,
+            strokeWidth: 0,
             cursor: "grab",
           },
           headerLabel: {
-            class: "joint-entityBlock-header-label",
+            fontFamily: `"RedHatMono", "Liberation Mono", consolas, "SFMono-Regular", menlo, monaco, "Courier New", monospace`,
```
LukasStordeur, line 36: Do we really need the full track of different fonts?

src/UI/Components/Diagram/shapes.ts
LukasStordeur: The color codes would have more meaning being a Type reflecting the EntityTypes
Also, don't forget to add the comment about "this"

src/UI/Components/Diagram/shapes.ts
```diff
@@ -297,13 +348,28 @@ export class ServiceEntityBlock extends shapes.standard.HeaderedRecord {
     return this;
   }
 
-  editColumns(data: Array<ColumnData>, shouldBeCollapsed = true) {
+  /**
+   * Edits the columns of the entity and optionally sets the collapsed state.
```
LukasStordeur, line 352:
```suggestion
   * Updates the columns of the entity and optionally sets the collapsed state.
```

LukasStordeur (changes requested):

src/UI/Components/Diagram/shapes.ts
matborowczyk: That's the same font track as in patternfly so I would opt for that to avoid some inconsistency just in case

src/UI/Components/Diagram/shapes.ts
matborowczyk: Added to general description of the shape

src/UI/Components/Diagram/shapes.ts
matborowczyk: Just in case as I don't se update here in the code viewer, yet everything is up to date:
Screenshot 2024-11-06 at 13 21 20

src/UI/Components/Diagram/shapes.ts
LukasStordeur: I was only refering to this specific `setTabColor`. I get why in other methods you do : return `this.get(blabla)`. What is odd here in this specific method, is that you returnonly  the setting of the attribute, or the entire scope. 
It's not about just following JointJs examples in general. It must have a reason why we need the full scope in this case.

Can you also assign the Hex codes to an enum so they make more sense? I also wonder if we can't use the JS-tokens from PF (not the css variables, but their JS value counterparts) to avoid needing to update or hunt down wherever we use those color codes.

LukasStordeur (approved):

PR #6020, opened 2024-10-30
add highlighting for loose inter-service relations

### LINKED ISSUE #5998: Add looseElement highlighter to loose related instances

matborowczyk:
# Description

https://github.com/user-attachments/assets/0cb45491-8cd0-4639-a6a2-746c6a2fb845

closes #5998 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/Data/Managers/V2/GETTERS/GetInstanceWithRelations/useGetInstanceWithRelations.ts
+++ b/src/Data/Managers/V2/GETTERS/GetInstanceWithRelations/useGetInstanceWithRelations.ts
@@ -15,6 +15,7 @@ export interface InstanceWithRelations {
   instance: ServiceInstanceModel;
   interServiceRelations: ServiceInstanceModel[];
   coordinates?: string;
+  referenced_by?: boolean;
 }
 
 /**
@@ -29,13 +30,15 @@ interface GetInstanceWithRelationsHook {
  * React Query hook to fetch an instance with its related instances from the API. The related instances are all instances connected with given instance by inter-service relation, both, as a parent and as a child.
  * @param {string} id - The ID of the instance to fetch.
  * @param {string} environment - The environment in which we are looking for instances.
- * @param {ServiceModel} serviceModel - The service Model of the instance (optional as it can be undefined at the init of the component that use the hook)
+ * @param {ServiceModel} [serviceModel] - The service Model of the instance (optional as it can be undefined at the init of the component that use the hook)
+ * @param {boolean} [referenced_by] - A flag indicating if we should fetch instances that reference our main instance - defaults to false.
  * @returns  {GetInstanceWithRelationsHook} An object containing a custom hook to fetch the instance with its related instances.
  */
 export const useGetInstanceWithRelations = (
   instanceId: string,
   environment: string,
   serviceModel?: ServiceModel,
+  referenced_by: boolean = false,
 ): GetInstanceWithRelationsHook => {
   //extracted headers to avoid breaking rules of Hooks
   const { createHeaders, handleErrors } = useFetchHelpers();
@@ -57,7 +60,7 @@ export const useGetInstanceWithRelations = (
   ): Promise<{ data: ServiceInstanceModel }> => {
     //we use this endpoint instead /lsm/v1/service_inventory/{service_entity}/{service_id} because referenced_by property includes only ids, without information about service_entity for given ids
     const response = await fetch(
-      `${baseUrl}/lsm/v1/service_inventory?service_id=${id}&include_deployment_progress=false&exclude_read_only_attributes=false&include_referenced_by=true&include_metadata=true`,
+      `${baseUrl}/lsm/v1/service_inventory?service_id=${id}&include_deployment_progress=false&exclude_read_only_attributes=false&include_referenced_by=${referenced_by}&include_metadata=true`,
       {
         headers,
       },
--- a/src/UI/Components/Diagram/Context/ComposerEditorProvider.tsx
+++ b/src/UI/Components/Diagram/Context/ComposerEditorProvider.tsx
@@ -69,6 +69,7 @@ export const ComposerEditorProvider: React.FC<Props> = ({
     instance,
     environment,
     mainService,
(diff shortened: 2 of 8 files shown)


src/UI/Components/Diagram/Context/ComposerEditorProvider.tsx
```diff
@@ -69,6 +69,7 @@ export const ComposerEditorProvider: React.FC<Props> = ({
     instance,
     environment,
     mainService,
+    !editable, //if editable is true, we don't fetch referenced_by instances as they should be displayed to keep it aligned with the regular instance form
```
matborowczyk, line 72:
```suggestion
    !editable, //if editable is true, we don't fetch referenced_by instances as they should not be displayed to keep it aligned with the regular instance form, they are only displayed in the composer viewer
```

src/UI/Components/Diagram/helpers.ts
```diff
@@ -929,3 +904,52 @@ export function showLinkTools(
 
   linkView.addTools(tools);
 }
+
+/**
+ * Function that remove data about niter-service relation between cells
```
matborowczyk, line 909:
```suggestion
 * Function that remove data about inter-service relation between cells
```

src/UI/Components/Diagram/helpers.ts
```diff
@@ -929,3 +904,52 @@ export function showLinkTools(
...
+  disconnectingCell: ServiceEntityBlock,
+): void => {
+  const elementRelations = elementCell.getRelations();
+
+  // resolve any possible relation connections between cells
+  if (elementRelations && elementRelations.has(String(disconnectingCell.id))) {
+    elementCell.removeRelation(String(disconnectingCell.id));
+
+    //get all connected entities to inter-service relation cell (the current relation that will be removed is included here)
+    const disconnectingCellNeighbors = graph.getNeighbors(disconnectingCell);
```
LukasStordeur, line 929: same remark here about "disconnecting", should be disconnected

src/UI/Components/Diagram/helpers.ts
```diff
@@ -929,3 +904,52 @@ export function showLinkTools(
...
+    const embeddedToID = disconnectingCell.get("embeddedTo");
+
+    //filter out every cell that is either embedded entity of to the disconnectingCell or is core or the disconnectingCell is embedded to it
+    const interServiceRelations = disconnectingCellNeighbors.filter(
+      (cell) =>
+        !(cell.get("embeddedTo") === disconnectingCell.id) ||
+        !(embeddedToID === cell.id),
+    );
+
+    //all left are the inter-service relation connections, if there is more than one then we should highlight the cell
```
LukasStordeur, line 941: the choice of name of the variable, and the comment above make it hard to follow why having more than one element should highlight the cell. I also find it odd you're using `=== 1 ` then just below 🤔 and not `< 1`

LukasStordeur (changes requested):

src/Data/Managers/V2/GETTERS/GetInstanceWithRelations/useGetInstanceWithRelations.ts
```diff
@@ -29,13 +30,15 @@ interface GetInstanceWithRelationsHook {
...
- * @param {ServiceModel} serviceModel - The service Model of the instance (optional as it can be undefined at the init of the component that use the hook)
+ * @param {ServiceModel} [serviceModel] - The service Model of the instance (optional as it can be undefined at the init of the component that use the hook)
+ * @param {boolean} [referenced_by] - A flag indicating if we should fetch instances that reference our main instance - defaults to false.
  * @returns  {GetInstanceWithRelationsHook} An object containing a custom hook to fetch the instance with its related instances.
  */
 export const useGetInstanceWithRelations = (
   instanceId: string,
   environment: string,
   serviceModel?: ServiceModel,
+  referenced_by: boolean = false,
```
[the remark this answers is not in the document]
matborowczyk: 1. Parameters can't have question mark and initializer. So it's being treated as optional, and the order here is fine.
2. I'll rename it but I thought it's clear as that property is going directly to the api endpoint

src/UI/Components/Diagram/helpers.ts
matborowczyk: improved comment, should be, if there is only one cell - as we are including the one that we are trying to disconnect. so 0 should not be possible here, if there is more than 1, which can theoretically happen, then we don't highlight that cell as it's already connected

src/Data/Managers/V2/GETTERS/GetInstanceWithRelations/useGetInstanceWithRelations.ts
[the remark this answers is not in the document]
LukasStordeur: In JavaScript (and TypeScript), arguments with default values are often considered "required" but with an initial default, meaning they provide a sensible value even if not explicitly passed. 

We tend to think of "optional" as "not required at all," but in reality, any argument with a default has a specific intention and should be thought of as a parameter with a baseline, and thus precede the optional arguments in the order of declaration.

src/Data/Managers/V2/GETTERS/GetInstanceWithRelations/useGetInstanceWithRelations.ts
```diff
@@ -29,15 +29,15 @@ interface GetInstanceWithRelationsHook {
  * React Query hook to fetch an instance with its related instances from the API. The related instances are all instances connected with given instance by inter-service relation, both, as a parent and as a child.
  * @param {string} id - The ID of the instance to fetch.
  * @param {string} environment - The environment in which we are looking for instances.
+ *  * @param {boolean} [includesReferencedBy] - A flag indicating if we should fetch instances that reference our main instance - defaults to false.
```
LukasStordeur, line 32: Update docstrings too =) you still flagged it as optional, and have double asterix

LukasStordeur (changes requested):

LukasStordeur (approved):

PR #6023, opened 2024-11-06
Add support for nested status properties

### LINKED ISSUE #6018: Add the ability to display nested properties in the server status view
If there would be  some nested properties like `env a: {connected: true, open: 3}`  then it would be converted to env a: [object Object] (checked on mocked data)
more context:

https://inmanta.slack.com/archives/CKRF0C8R3/p1730126446980999

matborowczyk:
# Description

Fix the issue that nested properties in status list crash the Page.

Screenshot 2024-11-06 at 12 06 15

closes #6018 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/6018-nested-status-properties.yml
+++ b/changelogs/unreleased/6018-nested-status-properties.yml
@@ -1,0 +1,6 @@
+description: Support nested status properties in the Status Page
+issue-nr: 6018
+change-type: patch
+destination-branches: [master]
+sections:
+  minor-improvement: "{{description}}"
--- a/src/Slices/Status/UI/StatusItem.tsx
+++ b/src/Slices/Status/UI/StatusItem.tsx
@@ -10,10 +10,11 @@ import {
   FlexItem,
 } from "@patternfly/react-core";
 import styled from "styled-components";
+import { DetailTuple } from "./StatusList";
 
 interface Props {
   name: string;
-  details: [string, Record<string, string> | string][];
+  details: DetailTuple[];
   icon: React.ReactNode;
   category?: string;
 }
@@ -23,9 +24,9 @@ interface Props {
  *
  * @props {Props} props - The properties for the status item component.
  * @prop {string} name - The name of the status item.
- * @prop {Array<[string, unknown]>} details - The details of the status item, which can include nested objects.
+ * @prop {DetailTuple[]} details - The details of the status item, which can include nested objects.
  * @prop {React.ReactNode} icon - The icon to display for the status item.
- * @prop {string} category - The category of the status item.
+ * @prop {string} [category] - The category of the status item.
  * @returns {React.FC<Props>} The rendered status item component.
  */
 export const StatusItem: React.FC<Props> = ({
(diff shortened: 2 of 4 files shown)


changelogs/unreleased/6018-nested-status-properties.yml
```diff
@@ -0,0 +1,6 @@
+description: Support nested status properties in the Status Page
+issue-nr: 6018
+change-type: patch
+destination-branches: [master]
```
matborowczyk, line 4:
```suggestion
destination-branches: [master, iso7]
```

src/Slices/Status/UI/StatusItem.tsx
```diff
@@ -13,11 +13,21 @@ import styled from "styled-components";
...
   icon: React.ReactNode;
   category?: string;
 }
 
+/**
+ * Renders a status item
+ *
+ * @props {Props} props - The properties for the status item component.
+ * @prop {string} name - The name of the status item.
+ * @prop {Array<[string, unknown]>} details - The details of the status item, which can include nested objects.
```
LukasStordeur, line 26: update docstrings, doesn't match the typing you declared in the props.

src/Slices/Status/UI/StatusItem.tsx
```diff
@@ -13,11 +13,21 @@ import styled from "styled-components";
 
 interface Props {
   name: string;
-  details: [string, string][];
+  details: [string, Record<string, string> | string][];
```
LukasStordeur, line 16: I don't know for you, but I find this is a bit hard to read.
I'd personaly go for something a bit more explicit like this:

```TS
type DetailKey = string;
type DetailValue = Record<string, string> | string;
type DetailTuple = [DetailKey, DetailValue];
```

And then use the DetailTuple type here and in other places where this is needed

src/Slices/Status/UI/StatusItem.tsx
```diff
@@ -59,3 +77,36 @@ const CompactDescriptionList = styled(DescriptionList)`
   --pf-v5-c-description-list--m-compact--RowGap: 0;
   margin-bottom: 16px;
 `;
+
+const IndentedDescriptionListGroup = styled(DescriptionListGroup)`
+  padding-left: 1rem;
```
LukasStordeur, line 82: We should use PF variables for this kind of things.

LukasStordeur (changes requested):
I'm not entirely a fan of the updated view, it makes the list very long. But this is for a later iteration I presume. We could implement collapsibles per section where needed. I'll add it to the discussions tomorrow in the planning meeting.

changelogs/unreleased/6018-nested-status-properties.yml
```diff
@@ -0,0 +1,6 @@
+description: Support nested status properties in the Status Page
+issue-nr: 6018
+change-type: patch
+destination-branches: [master, iso7]
```
LukasStordeur, line 4: Did you test it on iso7? Just for good measure.

matborowczyk: Yes, I run it on iso7 and it worked

src/Slices/Status/UI/StatusItem.tsx
```diff
@@ -52,10 +71,43 @@ const Category = styled.span`
...
+
+/**
+ * Renders sub list for Status value that is a Record instead of string.
+ *
+ * @props {SubListProps} props - The properties for the SubList component.
+ * @prop {string} name - The name of the property.
+ * @prop {Record<string, string>} properties - The sub properties to display in the SubList components.
+ * @returns {React.FC<SubListProps>} The rendered SubList component.
+ */
+const SubList: React.FC<SubListProps> = ({ name, properties }) => {
```
LukasStordeur, line 99: NestedListItem would be better, we have the StatusList component, so it makes it a bit confusing to name this one also a list, when it's actually an item.

LukasStordeur (changes requested):

PR #6024, opened 2024-11-06
Issue/5781 events tab service instance details

### LINKED ISSUE #5781: [Service Inventory V2] Add Events tab
https://internal.inmanta.com/development/frontend/designs/Service-Inventory-2024.html

Add the Events tab

LukasStordeur:
# Description

**Feature:**
The Instance Details has been extended now to display the Events for a selected version in the history.
The user can now consult the events present in the history logs in a dedicated tab in the main section of the page.
The rows can be toggled and will be highlighted in a golden hue when an event is an errored transition.
In the history logs, the instance_version displayed in the events is the selected version minus 1. This has yet to be estimated to be an issue or not, since on the main event page, using the events endpoint, the data seems to be slightly different.

The Compile report link will display a different text depending on the type of report. It has been discussed with other teams, and for now, we can assume that any report containing the "validate/validation" keyword is a validation report, all others are export reports.

closes #5781

image


src/Slices/ServiceInstanceDetails/Test/Page.test.tsx
```diff
@@ -411,6 +411,125 @@ describe("ServiceInstanceDetailsPage", () => {
       ),
     ).not.toBeVisible();
 
+    // Events
```
matborowczyk, line 414: Wouldn't that be better to have the tabs split, just in case if something break then get better overview what

src/Slices/ServiceInstanceDetails/UI/Components/CompileReportLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * CompileReportLink Component
+ *
+ * A plain link button that displays the text "Export" if it's an export compile link or "Validation".
+ * The difference between the two compiles can be spotted in the message present in the details of the event.
+ * A validation compile should always contain a key word validate/validation.
+ *
+ * @prop {Props} props - The props of the CompileReportLink
+ *  @prop {string} compileId - The id of the compile required to construct the link towards the Compile details page.
```
matborowczyk, line 20:
```suggestion
 * @prop {string} compileId - The id of the compile required to construct the link towards the Compile details page.
```

src/Slices/ServiceInstanceDetails/UI/Components/CompileReportLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+ *  @prop {boolean} isExport - Whether the compile link is forwarding to an export report or not. The other case being a validation report.
```
matborowczyk, line 21:
```suggestion
 * @prop {boolean} isExport - Whether the compile link is forwarding to an export report or not. The other case being a validation report.
```

src/Slices/ServiceInstanceDetails/UI/Tabs/EventTabContent.tsx
```diff
@@ -1,10 +1,300 @@
...
+ * 8. when a transfer is received, but not executed, we log an ignored_transition
+ *
+ * The easiest way of looking at it is to consider all event with the same correlation id together.
+ * For the is_error_transition  and ignored_transition the state of the last event in the sequence it most relevant
+ *
+ * @note We are using the events that are pressent in the history logs for the instance details view.
+ * A link button allows the user to navigate the the full events page.
+ *
+ * @prop {Props} props - The EventsTabContent Props
+ *  @prop {string} selectedVersion - The actively selected version on the page.
```
matborowczyk, line 68:
```suggestion
 * @prop {string} selectedVersion - The actively selected version on the page.
```

src/Slices/ServiceInstanceDetails/UI/Components/CompileReportLink.tsx
LukasStordeur: The indent is on purpose, I've been doing it everywhere honestly out of habit to have some sort of hierarchy, but I can adjust it if needed

src/Slices/ServiceInstanceDetails/Test/Page.test.tsx
LukasStordeur: It would slow down the tests considerably to reinstantiate an entire page for each tab, so that's why I did it like this

src/Slices/ServiceInstanceDetails/UI/Tabs/EventTabContent.tsx
```diff
@@ -1,10 +1,300 @@
...
+
+  if (logsQuery.isLoading) {
+    return (
+      <TabContentWrapper id="events">
+        <LoadingView />
+      </TabContentWrapper>
+    );
+  }
+
+  if (!logsQuery.data) {
```
matborowczyk, line 84: Hmm I wonder,  couldn't this view have different order

`if(logsQuery.isSuccess)` condition would have that logic about selected version
I don't see assertion for `logsQuery.isError` here, I assume it's somewhere as this query got initialised somewhere. but if not then instead of ` if (!logsQuery.data)` I would add the one for Error

and at the end we fall back to isLoading as a default return, then we don't need to assert that and then there is no need for ` if (!logsQuery.data)` condition

src/UI/Components/DateWithTimeDiffTooltip/DateWithTimeDiffTooltip.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * DateWithTimeDiffTooltip Component
+ *
+ * Displays a Timestamp with a tooltip showing the time difference between two timestamps in milliseconds.
+ * The timestamp is displayed to the third second fraction.
+ *
+ * @Note Moment will round any further fraction.
+ * We are getting the timestamps to the fourth fraction from the API, but we can't display them with the current Moment library.
+ *
+ * @prop {Props} props - The Props of the DateWitTimeDiffTooltip
+ *  @prop {string} timestamp1 - The first timestamp that needs to be compared with.
```
matborowczyk, line 23:
```suggestion
 * @prop {string} timestamp1 - The first timestamp that needs to be compared with.
```

src/UI/Components/DateWithTimeDiffTooltip/DateWithTimeDiffTooltip.tsx
```diff
@@ -0,0 +1,44 @@
...
+ *  @prop {string} timestamp2 - The second timestamp that needs to be compared against.
```
matborowczyk, line 24:
```suggestion
 * @prop {string} timestamp2 - The second timestamp that needs to be compared against.
```

src/Slices/ServiceInstanceDetails/UI/Components/CompileReportLink.tsx
matborowczyk: Oh okay, no need then I just thought It was off :D I don't mind having it or not, I personally keep them alligned :D

src/Slices/ServiceInstanceDetails/UI/Tabs/EventTabContent.tsx
LukasStordeur: No, because the data is used in the next parts of the code, so I do need the check before using it

Concerning the isError assertion, it is equivalent to !logsQuery data, because the query will fire isError when there is no data at this point. I can add the assertion if you want it

matborowczyk: if logsQuery.isSuccess is true then it will automatically mean that the data is defined, unless interface of the query said different(succesful query can return undefined, but I don't think it's the case here) and then I would change the assertion to the isError, it follows what react query flow and is a bit better if it comes to readability, also you current error handling shows user that there is no data, but no actual error message

LukasStordeur: If there is data, it also automatically means the query is successful. 
Like I said, I can add an additional assertion if you want, or a comment in the code for readability, but the `!logsQuery.data` is required for further use in the code. 

What other error message do you want to display 🤔
You'll have the query error message in the History section, and one in the relevant tabs. 
You'll get:  "There is no events data available for this version."  Which matches what would be displayed on other tabs and  "Error loading Version History" in the History section on the right side of the screen.

matborowczyk (approved):

PR #6027, opened 2024-11-06
Issue/destination fix

matborowczyk:
# Description

Remove iso7 from the desitination in the changelog in task 5949 as it could not be merged due to the missing properties in Th component- also the warning doesn't affect iso7


LukasStordeur (changes requested):
Two changelogs will not pass with the merge tool

LukasStordeur (approved):

PR #6044, opened 2024-11-12
Hide form actions in composer viewer

### LINKED ISSUE #6036: block Instance Composer Viewer from actions

matborowczyk:
# Description

Hide form actions buttons in the instance composer viewer.

Screenshot 2024-11-12 at 09 37 28

closes #6036 

# Self Check:

- [x] Changelog entry


matborowczyk:
> Just wondering why this isn't part of the context instead of a prop you need to pass along ?

It's a variable that doesn't change in the lifecycle of the component, it's used quite shallow if it comes to props, so I don't see there an immediate advantage to pass it to the context

LukasStordeur (approved):

PR #6053, opened 2024-11-15
add identifying attribute to the body of core and inter-service relation entities

### LINKED ISSUE #6031: Show Identifying attribute on root level elements
Root elements don't explicitly have key-attributes to show. Meaning the blocks are empty. 
We should display the `service_identity_attribute_value`, which can be found in the instance data. Like below:

Image

Which property is used for the value can be found in the model `service_identity` and `service_identity_display_name` if available. 
Image

When the instance is in create mode, you will then also know which value to pick in the formstate when this one becomes available.

matborowczyk:
# Description

adds identifying attribute to the body of core and inter-service relation entities

https://github.com/user-attachments/assets/5338b411-131c-431a-ada1-1bc93e878f0c

closes #6031

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/UI/Components/Diagram/init.ts
+++ b/src/UI/Components/Diagram/init.ts
@@ -161,11 +161,12 @@ export function diagramInit(
     },
 
     editEntity: (cellView, serviceModel, attributeValues) => {
-      const keyAttributes = serviceModel.key_attributes || [];
+      let keyAttributes = serviceModel.key_attributes || [];
 
       //service_identity is a unique attribute to Service model, but doesn't exist in the Embedded Entity model
       if ("service_identity" in serviceModel && serviceModel.service_identity) {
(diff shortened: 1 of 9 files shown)


src/UI/Components/Diagram/init.ts
```diff
@@ -161,11 +161,18 @@ export function diagramInit(
     },
 
     editEntity: (cellView, serviceModel, attributeValues) => {
+      let keyAttributes = serviceModel.key_attributes || [];
```
LukasStordeur, line 164: Why do we have the exact same logic twice?

LukasStordeur (approved):

src/UI/Components/Diagram/init.ts
matborowczyk: extracted

PR #6056, opened 2024-11-18
Hide read only embedded entities

### LINKED ISSUE #6034: hide all "r" embedded entities from the view in all modes
https://inmanta.slack.com/archives/C045Y42N49W/p1730916907386389

5. We need to introduce something to be able to hide/ignore certain relation so that the "allocated" block is not visible because the really "break" the flow. We can use annotations for this

- we could indeed just hide all "r" relations

matborowczyk:
# Description

From now on Composer won't display embedded entities with modifier set to "r"

on recording allocated entity is "r"
https://github.com/user-attachments/assets/24086e48-887f-47ba-9b41-01c06fbafeba

closes #6034 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/UI/Components/Diagram/actions.ts
+++ b/src/UI/Components/Diagram/actions.ts
@@ -359,7 +359,7 @@ export function appendEmbeddedEntity(
 
     //iterate through embedded entities to create and connect them
     embeddedEntity.embedded_entities
-      .filter((entity) => entity.modifier !== "r") // filter out read-only embedded entities to de-clutter the view
+      .filter((entity) => entity.modifier !== "r") // filter out read-only embedded entities to de-clutter the view as they can't be edited and can be in multiple places which would result with enormous tree of cells in the graph and the canvas
       .forEach((entity) => {
         const appendedEntity = appendEmbeddedEntity(
           paper,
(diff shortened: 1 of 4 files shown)


src/UI/Components/Diagram/actions.ts
```diff
@@ -358,25 +358,27 @@ export function appendEmbeddedEntity(
...
-        paper,
-        graph,
-        entity,
-        entityInstance[entity.name] as InstanceAttributeModel,
-        instanceAsTable.id as string,
-        embeddedEntity.name,
-        presentedAttr,
-        isBlockedFromEditing,
-      );
+    embeddedEntity.embedded_entities
```
LukasStordeur, line 361: Can you please add to the comments in the code the reason why we are filtering these out.

LukasStordeur (commented):
This is just a small portion of the code, but already, we are iterating through the same thing multiple times, or that's the impression I get. I don't have the overview picture of the code, but I get the impression we are looping over a lot, multiple times in different places, probably could be unified to keep it more optimized at some point

src/UI/Components/Diagram/actions.ts
matborowczyk: added

LukasStordeur: de-cluttering is good but doesn't tell why we are doing this, I vaguely know it's about the alocated data, but I'd like to see an actual explanation on it why they don't need to be displayed.

LukasStordeur (approved):

PR #6069, opened 2024-11-26
Issue/6007 e2e removal

### LINKED ISSUE #6007: add e2e case for removal of inter-service relation
+ check assertion for Fit-to-screen and slider to assert by width, not only by the percentage value

matborowczyk:
# Description

Added e2e for inter-service relations, fixed issue with removing inter-service relations for instance that could have an array of those, added constraint to deploy button, fixed issue with relation removal didn't trigger the update of the cell

closes #6007 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/cypress/e2e/scenario-8-instance-composer.cy.js
+++ b/cypress/e2e/scenario-8-instance-composer.cy.js
@@ -965,6 +965,7 @@ if (Cypress.env("edition") === "iso") {
         .should("have.text", "up", { timeout: 90000 });
 
       //go to details view
+      cy.get('[aria-label="row actions toggle"]').click();
       cy.get("button").contains("Instance Details").click();
 
       //assert that in Active attribute we have only 1 relation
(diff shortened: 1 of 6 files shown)


cypress/e2e/scenario-8-instance-composer.cy.js
```diff
@@ -870,5 +870,135 @@ if (Cypress.env("edition") === "iso") {
...
       });
     });
+
+    it("8.5 composer edit view is able to remove inter-service relation from instance", () => {
+      // Select 'test' environment
+      cy.visit("/console/");
+      cy.get('[aria-label="Environment card"]')
+        .contains("lsm-frontend")
+        .click();
+      cy.get(".pf-v5-c-nav__item").contains("Service Catalog").click();
```
LukasStordeur, line 880: If we can avoid using the class, that would be great :)

LukasStordeur (changes requested):

PR #6081, opened 2024-12-02
Hide left sidebar from Composer Viewer

### LINKED ISSUE #5988: Hide left sidebar in the View mode

matborowczyk:
# Description

Hides Left sidebar from the instance composer viewer. it was done by hiding element, as conditional rendering was stopping lifecycle due to missing ref for sidebar that is required in further steps. It could be resolved but that would equal to some additional conditions, and display: none is straightforward, clean and simplest solution in my opinion

closes #5988 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/UI/Components/Diagram/Canvas.tsx
+++ b/src/UI/Components/Diagram/Canvas.tsx
@@ -156,11 +156,15 @@ export const Canvas: React.FC<Props> = ({ editable }) => {
       <DictModal />
       <CanvasWrapper id="canvas-wrapper" data-testid="Composer-Container">
         <LeftSidebarContainer
-          className="left_sidebar"
+          className={`left_sidebar ${!editable && "view_mode"}`}
           data-testid="left_sidebar"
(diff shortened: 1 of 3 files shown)


src/UI/Components/Diagram/Canvas.tsx
```diff
@@ -156,11 +156,15 @@ export const Canvas: React.FC<Props> = ({ editable }) => {
       <DictModal />
       <CanvasWrapper id="canvas-wrapper" data-testid="Composer-Container">
         <LeftSidebarContainer
-          className="left_sidebar"
+          className={`left_sidebar ${!editable && "view_mode"}`}
```
LukasStordeur, line 159: 🤔 why do we need both assertions? Doesn't !editable imply that you are in view-mode?

matborowczyk: It's not assertion, we append class there

LukasStordeur (approved):

PR #6083, opened 2024-12-03
Issue/6006 split files

### LINKED ISSUE #6006: separate actions and helpers file into smaller files(and it's tests)
https://github.com/inmanta/web-console/pull/5985#discussion_r1812375670

matborowczyk:
# Description

Separate actions and helpers function into smaller files to improve readability of the code

closes #6006 

# Self Check:

- [ ] Changelog entry


src/UI/Components/Diagram/helpers/connetions.test.ts
LukasStordeur, line 1: Typo in file name you wrote connetions instead of connections.

LukasStordeur (commented):
You're not using index files, but those would make the imports cleaner. You'll be able to import from `./helpers` or `@/helpers`, or `./actions` instead of needing to import each function separately when they happen to be spilt in a different file. 

I don't find `misc` a clear definition for the functions that are gathered there. Perhaps they could have been split up further? The test file is already considered too big to render the diff in the PR viewer, which indicates it might still be necessary to consider splitting it.

src/UI/Components/Diagram/helpers/connetions.test.ts
matborowczyk: fixed typo, added index files and renamed the misc file to "data" after a consideration it fit the description, they kind of related, and test files has 576 lines of code, which is still much but it's due to the fact that they manipulate objects. I think there is a chance that on the team  week we could erase some of them(helpers) if we could successfuly move to InstanceForm type of state management

src/UI/Components/Diagram/components/RightSidebar.test.tsx
```diff
@@ -20,7 +20,7 @@ import {
   defaultCanvasContext,
 } from "../Context";
 import { containerModel } from "../Mocks";
-import { addDefaultEntities } from "../actions";
+import { addDefaultEntities } from "../actions/createMode";
```
LukasStordeur, line 23: update import

LukasStordeur (approved):

PR #6110, opened 2025-01-02
6104 rip out old code

### LINKED ISSUE #6104: Remove Old collapsible rows in service inventory

LukasStordeur:
# Description
 
Remove the code sections that are now considered obsolete after the creation of the new Instance details page. 

closes #6104


cypress/e2e/scenario-8-instance-composer.cy.js
```diff
@@ -744,7 +744,7 @@ if (Cypress.env("edition") === "iso") {
       // Expect Canvas to be visible
       cy.get(".canvas").should("be.visible");
 
-      cy.get("main").scrollTo("bottom"); //scroll to the bottom as the error container is clipped in 1500x900 viewport
+      //      cy.get("main").scrollTo("bottom"); //scroll to the bottom as the error container is clipped in 1500x900 viewport
```
LukasStordeur, line 747: I had to comment this out to have the tests pass localy in the ui, no idea if we still need it as is or not. Weird that the changes of this pr affected this though...

matborowczyk: we can remove that as it is no longer required I guess.

It's weird that it was working solution for failing scenario and then it was causing errors :D

matborowczyk (approved):

PR #6140, opened 2025-01-13
Add Link with Discovery resources uri to the Discovered Resources Page

### LINKED ISSUE #5946: [Discovered Resources] - Add Resource Dicsovery uri
Add a Resource Dicsovery uri field into the collapsible that forwards to the Resource details page of that resource

matborowczyk:
# Description

Add Link with Discovery resources uri to the Discovered Resources Page

Screenshot 2025-01-13 at 13 54 58

closes #5946 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/5946-discovery-uri.yml
+++ b/changelogs/unreleased/5946-discovery-uri.yml
@@ -1,0 +1,6 @@
+description: Adds discovery uri to the Discovered Resources Page
+issue-nr: 5946
+change-type: patch
+destination-branches: [master]
+sections:
+  minor-improvement: "{{description}}"
--- a/src/Slices/ResourceDiscovery/UI/Components/ManagedResourceLink.tsx
+++ b/src/Slices/ResourceDiscovery/UI/Components/ManagedResourceLink.tsx
@@ -4,19 +4,20 @@ import { ResourceLink } from "@/UI/Components";
 
 interface Props {
   resourceUri: string | null;
+  type: "managed" | "discovery";
 }
 
 /**
- * @Component that renders a link to the managed resource
+ * @Component that renders a link to the managed or the discovery resource
  * or a "-" if the resourceUri is null or empty.
  *
- * managed_resource_uri comes in format : /api/v2/resource/<rid>
+ * uris comes in format : /api/v2/resource/<rid>
  *
- * @param resourceUri : API URI of the managed resource
+ * @param resourceUri : API URI of the managed/discovery resource
  *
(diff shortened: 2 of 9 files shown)


changelogs/unreleased/5946-discovery-uri.yml
```diff
@@ -0,0 +1,6 @@
+description: Adds discovery uri to the Discovered Resources Page
+issue-nr: 5946
+change-type: patch
+destination-branches: [master]
```
LukasStordeur, line 4: This is a change that is also meant to be done on iso7

src/Slices/ResourceDiscovery/UI/Components/ManagedResourceLink.tsx
```diff
@@ -4,19 +4,20 @@ import { ResourceLink } from "@/UI/Components";
...
- * managed_resource_uri comes in format : /api/v2/resource/<rid>
+ * uris comes in format : /api/v2/resource/<rid>
  *
- * @param resourceUri : API URI of the managed resource
+ * @param resourceUri : API URI of the managed/discovery resource
  *
  * @returns ManagedResourceLink component
  */
-export const ManagedResourceLink: React.FC<Props> = ({ resourceUri }) => {
+export const ManagedResourceLink: React.FC<Props> = ({ resourceUri, type }) => {
```
LukasStordeur, line 20: Right now, the name of the component suggests it only is for one type. So, maybe we should rename it to something more meaningful. `ResourceLink` is perhaps too generic, and we might confuse it with other similar components. But maybe that's fine 🤔

src/Slices/ResourceDiscovery/UI/DiscoveredResourcesRow.tsx
```diff
@@ -40,37 +38,52 @@ export const DiscoveredResourceRow: React.FC<Props> = ({
         </Td>
         <Td
           dataLabel={words("discovered.column.resource_id")}
-          style={{ width: "50vw" }}
           data-testid={words("discovered.column.resource_id")}
         >
-          {row.discovered_resource_id}
+          <Truncate content={row.discovered_resource_id} />
```
LukasStordeur, line 43: This can have unexpected behaviours I've noticed.
You should be able to add the `modifier="truncate"` on the TD element like so:
`<Td modifier="truncate">content</Td>` 
That will respect the width set by the Th element if it's set for example.

LukasStordeur (approved):

PR #6159, opened 2025-01-21
Extract error container into independent component

### LINKED ISSUE #6059: Extract the JSON-editor and Instance Composer Error Message component
Both components that display Error message for those Views have the same UI, and they could be extracted in one component, at first they were separated as the logic to display was quite different but that could be investigated further.

https://github.com/inmanta/web-console/pull/6046#pullrequestreview-2445787380

matborowczyk:
# Description

Extract error container into independent component, visually it looks the same as it was

closes #6059 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/UI/Components/Diagram/components/ErrorsContainer.tsx
+++ b/src/UI/Components/Diagram/components/ErrorsContainer.tsx
@@ -1,7 +1,7 @@
 import React, { useContext } from "react";
 import { Content } from "@patternfly/react-core";
 import { words } from "@/UI/words";
-import Validation from "../../Validation/Validation";
+import { ErrorMessageContainer } from "../../ErrorMessageContainer";
 import { CanvasContext } from "../Context";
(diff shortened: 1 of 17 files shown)


src/UI/Components/Diagram/components/ErrorsContainer.tsx
```diff
@@ -13,7 +13,7 @@ import { RelationCounterForCell } from "../interfaces";
  *
  * @returns {React.FC | null} The rendered validation component or null if no missing relations are found or the canvas is not edited/dirty.
  */
-export const Validation: React.FC = () => {
+export const ErrorsContainer: React.FC = () => {
```
LukasStordeur, line 16: Maybe ErrorMessageCollapsible/ErrorMessageContainer would be a bit more explicit

LukasStordeur (approved):

PR #6172, opened 2025-01-24
Replace v1 queries related to services for React-Query 

### LINKED ISSUE #5975: Replace endpoint service inventory
Replace the catalog call on the service inventory to the updated service inventory call to display the summary

matborowczyk:
# Description

Replace v1 queries related to services for React-Query 

closes #5975 

# Self Check:

- [ ] Changelog entry


src/Data/Managers/V2/POST/UpdateCatalog/useUpdateCatalog.ts
```diff
@@ -0,0 +1,53 @@
...
+import { PrimaryBaseUrlManager } from "@/UI";
+import { useFetchHelpers } from "../../helpers";
+
+/**
+ * React Query hook for updating environment catalog.
+ *
+ * @param {string} environment  - The environment to use for creating headers.
+ * @returns {UseMutationResult<void, Error, void, unknown>}- The mutation object from `useMutation` hook.
+ */
+export const useUpdateCatalog = (
```
LukasStordeur, line 15: The name might be a bit confusing, perhaps we could go for something closer to the endpoint name, like useExportCatalog. I'd avoid the update keyword since it's a post call.

LukasStordeur (approved):
Aside the name of the hook, I think everything is fine. Just noticed you updated the Service Config, but not the one for the instance details yet, so I think this doesn't close the initial issue with the refresh on the instance details page of instances which have a config.

PR #6186, opened 2025-02-05
Buildmaster pipeline update

LukasStordeur:
# Description

- Local-auth tests were failing because the dropdown menu didn't close after an option was selected. 
- update logo to switch correctly
- update colors in charts to match progress bar

https://github.com/user-attachments/assets/1557e5a3-9dcf-4760-a824-f87aee837c86


src/UI/Root/Components/Header/EnvSelector/EnvSelector.tsx
```diff
@@ -42,6 +42,7 @@ export const EnvSelector: React.FC<Props> = ({
         position: "end",
       }}
       onOpenChange={(open: boolean) => setIsOpen(open)}
+      onClick={() => setIsOpen(!isOpen)}
```
matborowczyk, line 45: Kind of optional change, but assignment of that function would be better

matborowczyk (approved):
All good, only that one comment

PR #6188, opened 2025-02-06
Add all relevant queries for instance and service

matborowczyk:
Sub PR to simplify the the review process for removal of V1 queries related to service and service instance as I missed one service query previously.

Added one Temporary component, as pagination changed with V2, and other pages are still using the v1 queries


src/UI/Components/OldPaginationWidget/Provider.tsx
```diff
@@ -0,0 +1,88 @@
...
+  setCurrentPage: (currentPage: CurrentPage) => void;
+  variant?: "top" | "bottom";
+}
+
+/**
+ * Wrapper for Pagination for V1 queries as we are basing our-selfs on links served by backend - This component is Temporarily added to maintain pagination functionality for V1 queries until all will be removed
+ *
+ * Note: Parameters responsible for pagination on endpoint doesn't allow to pass numerical range for displayed results, or any other way to navigate through pages other than previous/next page
+ */
+export const Provider: React.FC<Props> = ({
```
LukasStordeur, line 29: Maybe we can avoid renaming it in the index file, and add a extra note about his being the old one?

LukasStordeur (changes requested):

src/Data/Managers/V2/ServiceInstance/PatchAttributes/helpers.test.ts
```diff
@@ -0,0 +1,77 @@
+import { Field } from "@/Test";
+import { getBodyV1, getBodyV2 } from "./helpers";
+
+const currentAttributes = { attr1: "some value", attr2: "", attr3: null };
+
+test("GIVEN getBodyV1 WHEN not setting optional attributes THEN generates correct body ", () => {
```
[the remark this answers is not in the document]
matborowczyk: it depends on service models, I assume there could be clients that still have some old models, and it's not easy to get rid of it, but that question to core/solutions

src/UI/Components/OldPaginationWidget/Provider.tsx
matborowczyk: removed the renmaing, there is note in the docstrings about it being used in V1 and should be removed when all v1 will be gone

LukasStordeur (approved):

PR #6219, opened 2025-03-05
add new, adequate display name for the embedded instances

### LINKED ISSUE #5591: [Composer] names of the entity is not matching
Currently the name of the relation are being used as titles, it should the actual name of the entity.

matborowczyk:
# Description

Add the new display name for embedded entities according to the new property

Screenshot 2025-03-05 at 15 33 20

closes #5591 

# Self Check:

- [ ] Changelog entry


src/UI/Components/Diagram/Mocks/services.ts
```diff
@@ -825,6 +825,7 @@ export const testService: ServiceModel = {
   embedded_entities: [
     {
       name: "peerings",
+      type: "peerings",
```
LukasStordeur, line 828: isn't the type supposed to be singular instead of plural? I know this is mock data, but wouldn't hurt to keep it right.

src/UI/Components/Diagram/shapes.ts
```diff
@@ -254,13 +255,13 @@ export class ServiceEntityBlock extends shapes.standard.HeaderedRecord {
       },
     );
 
-    this.set("entityName", name);
-    this.attr(["headerLabel", "data-testid"], "header-" + name);
+    this.set("entityName", name); //name is still assigned to the entityName attribute, not to disturb the existing logic
```
LukasStordeur, line 258: hmm, might be more confusing later on tho. No?

matborowczyk: I thought about this, and I think it's better to keep it as it is, because the data that matters in case of the logic is the name. we want to display the type but it is actually not relevant. I can add more to the comment if you want.

src/UI/Components/Diagram/Mocks/services.ts
matborowczyk: My bad, changed

src/UI/Components/Diagram/shapes.ts
LukasStordeur: yeah that's okay, a bit more context never hurts

LukasStordeur (approved):
If you add that bit of context to the entityName comment, it's okay

PR #6226, opened 2025-03-07
Enable compression for HTTPS connections and improve response handling

LukasStordeur:
Enable compression for HTTPS connections and improve response handling.

Cloudflare has compression enabled by default, which doesn't provide a content-lenght in the API responses like we used to assert whether there is a response or not.

### DIFF
--- a/src/Slices/CreateInstance/UI/CreateInstance.tsx
+++ b/src/Slices/CreateInstance/UI/CreateInstance.tsx
@@ -47,20 +47,11 @@ export const CreateInstance: React.FC<Props> = ({ serviceEntity }) => {
       setIsDirty(true);
       setErrorMessage(error.message);
     },
-    onSuccess: (response) => {
-      if (!response || !response.data) {
-        console.warn(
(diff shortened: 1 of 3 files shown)


matborowczyk:
I I'm not big fan of `<Response | undefined> ` especially if we lose Typing assertion from react-query.

src/Slices/CreateInstance/UI/CreateInstance.tsx
```diff
@@ -47,11 +47,20 @@ export const CreateInstance: React.FC<Props> = ({ serviceEntity }) => {
       setIsDirty(true);
       setErrorMessage(error.message);
     },
-    onSuccess: ({ data }) => {
+    onSuccess: (response) => {
+      if (!response || !response.data) {
```
matborowczyk, line 51: this should not be like that I guess

LukasStordeur: oh my bad, was from debugging

matborowczyk (approved):

PR #6228, opened 2025-03-11
upgrade queries related to compilations

### LINKED ISSUE #6223: continuation on react-query replacement

matborowczyk:
# Description

Upgraded compilation related queries, remove one duplicate, upgraded test for the affected views

closes #6223 

# Self Check:

- [ ] Changelog entry


src/Data/Managers/V2/helpers/useQueries.ts
```diff
@@ -441,3 +441,25 @@ export const useDeleteWithoutEnv = (options?: { message?: string }) => {
     }
   };
 };
+
+export const useHead = () => {
```
LukasStordeur, line 445: docstrings

src/Slices/CompileReports/UI/Page.tsx
```diff
@@ -35,55 +33,67 @@ export const Page: React.FC = () => {
...
+            <PaginationWidget
+              data={data}
+              pageSize={pageSize}
+              setPageSize={setPageSize}
+              setCurrentPage={setCurrentPage}
+            />
+          }
+          afterRecompile={() => refetch()}
+        />
+        {data.data.length <= 0 ? (
```
LukasStordeur, line 77: don't we typically do the select method on the GET queries to avoid data.data?

src/UI/Root/Components/Sidebar/Navigation/Navigation.test.tsx
```diff
@@ -11,6 +16,7 @@ import {
   QueryResolverImpl,
   QueryManagerResolverImpl,
 } from "@/Data";
+import * as queryModule from "@/Data/Managers/V2/Compilation/GetCompilerStatus/useGetCompilerStatus";
```
LukasStordeur, line 19: It's the first time I see us using this naming convention, is there a specific reason?

src/UI/Root/Components/Sidebar/Navigation/Navigation.test.tsx
```diff
@@ -171,24 +180,27 @@ test("GIVEN Navigation WHEN on 'Service Catalog' THEN 'Service Catalog' is highl
...
-    method: "HEAD",
-    url: "/api/v1/notify/env",
-  });
+  expect(Indication).toBeNull();
+});
 
-  await act(async () => {
-    await apiHelper.resolve(200);
+test("GIVEN Navigation WHEN Compilation Reports are pending THEN 'Compile Reports' Indication is visible", async () => {
+  jest.spyOn(queryModule, "useGetCompilerStatus").mockReturnValue({
```
LukasStordeur, line 193: Why the spy instead of the msw implementation?

src/UI/Root/Components/Sidebar/Navigation/NavigationItem.tsx
```diff
@@ -113,9 +102,12 @@ const CompileReportItem: React.FC<Label & Url> = ({ label, url }) => {
         aria-label="Sidebar-Navigation-Item"
       >
         {label}
-        {data.kind === "Success" && data.value === true && (
+        {isSuccess && data === 200 && (
```
LukasStordeur, line 105: why is data 200? and not the statuscode? Feels strange to me to assign the status code to the data object

src/Data/Managers/Helpers/QueryManager/ContinuousWithEnv.test.tsx
```diff
@@ -6,6 +6,7 @@ import { StoreProvider } from "easy-peasy";
 import { Either, RemoteData } from "@/Core";
 import { QueryManagerResolverImpl, QueryResolverImpl } from "@/Data";
 import { getStoreInstance } from "@/Data/Store";
+import { drawerQuery } from "@/Slices/Notification/Core/Query";
```
LukasStordeur, line 9: drawerQuery feels like the wrong name. I know this is out of scope of this specific task, but just wanted to voice the observation.

LukasStordeur (changes requested):

src/Slices/CompileReports/UI/Page.tsx
matborowczyk: queries with pagination have format of {data:[], links:{}, metadata:{}}

src/UI/Root/Components/Sidebar/Navigation/Navigation.test.tsx
matborowczyk: because there is no trigger in that specific scenario, I opted for shortest approach

src/UI/Root/Components/Sidebar/Navigation/NavigationItem.tsx
matborowczyk: there is no status code in the react-query itself, there is FetchStatus but that's not the same, it's the status of the query, not status code of the request itself. the boolean value that was in the v1 request was the same assertion as it is here

src/UI/Root/Components/Sidebar/Navigation/Navigation.test.tsx
matborowczyk: nope, Do you want me to change it, I chose it spontaneously as it is imported as a whoel module, and it's query so it came out naturally

src/Data/Managers/Helpers/QueryManager/ContinuousWithEnv.test.tsx
matborowczyk: I think it will be removed when we move notification queries to v2

src/UI/Root/Components/Sidebar/Navigation/Navigation.test.tsx
LukasStordeur: I think it's better to not rename the import :D

src/UI/Root/Components/Sidebar/Navigation/NavigationItem.tsx
LukasStordeur: No, what I mean is, this useGetCompilerStatus query is normally a HEAD call, which doesn't return a response body. For me, it doesn't make sense that data should equal 200.

src/UI/Root/Components/Sidebar/Navigation/Navigation.test.tsx
matborowczyk: Oh, that's what you meant. we need to import like this to be able to spy on that module, this convention is already existing in the code base. I think it's the best approach from what I researched

PR #6239, opened 2025-03-19
Security alerts iso7

matborowczyk:
fixing conflicts for security alerts


changelogs/unreleased/buildmaster-security-prismjs.yml
```diff
@@ -0,0 +1,3 @@
+description: Fix security alert regarding the prismjs and @babel/runtime package
+change-type: patch
+destination-branches: [master, iso8, iso7]
```
LukasStordeur, line 3: you'll have to remove master and iso8. Also not sure if we need to fix them still on iso7 🤔 if it doesn't cause issues ,fine, but if it makes some tests fail, we might not want to spend time fixing them there

PR #6244, opened 2025-03-26
Fix version conflict - Instance details page

### LINKED ISSUE #6233: Resources error fix for service instance details view
when update to the instance occur the resoruces tab doesn't update version immediately and there is some split second when error of version mismatch happen occur.

proposed solution is to assert both on error and before if the version is correct, and do a retry for this specific view

https://files.slack.com/files-tmb/T02KH0H20-F08HH3D8C3X-75368df606/image_720.png

matborowczyk:
Fix version conflict error behaviour in resource tab on instance details page

https://github.com/user-attachments/assets/f75330b0-2475-4692-b7fa-bd728e2700d8

closes #6233 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/Slices/ServiceInstanceDetails/UI/Tabs/ResourcesTabContent.tsx
+++ b/src/Slices/ServiceInstanceDetails/UI/Tabs/ResourcesTabContent.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from "react";
+import React, { useContext } from "react";
 import { Content, TabContent, TabContentBody } from "@patternfly/react-core";
 import { useGetInstanceResources } from "@/Data/Managers/V2/ServiceInstance";
 import { words } from "@/UI";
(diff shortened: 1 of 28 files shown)


src/Slices/ServiceInstanceDetails/UI/Tabs/ResourcesTabContent.tsx
```diff
@@ -43,14 +42,21 @@ export const ResourcesTabContent: React.FC = () => {
...
-        <ErrorView
-          message={error.message}
-          ariaLabel="Error_view-Resources-content"
-        />
-      </TabContent>
-    );
+    // If the error is because of the version, we don't want to show the error view, and fall back to the loading view as the process of updating version is still ongoing
+    if (
+      !error.message.includes(
+        "Request conflicts with the current state of the resource: The given current version",
```
LukasStordeur, line 48: Is this the only way? Doesn't seem like a sustainable solution.

matborowczyk: Added status code to error handling, and updated all queries to have that, to keep them aligned

LukasStordeur (approved):

PR #6245, opened 2025-03-27
Fix JSON Editor toggling

### LINKED ISSUE #6177: Switching from Form to JSON-Editor after updating integer values coverts integers to string and creates errors
When updating integer values(optional or required) and changing to JSON-Editor it will converts those updates values to string which will trigger error in the json editor validation

matborowczyk:
# Description

Fix switching between JSON editor and Input Form converting values to string when it should not.
Converting to string will now only happen if the dict is invalid

closes #6177 

https://github.com/user-attachments/assets/0206934b-e10d-4ed8-8e25-e538e150670a

# Self Check:

- [ ] Changelog entry


src/UI/Components/ServiceInstanceForm/Helpers/createFormState.spec.ts
LukasStordeur, line 1: Descriptions of the tests are the same.
Aren't the createEditFormState and createDuplicateFormState basically the same? I don't mind asserting them too, but might feel a bit redundant. 
It does also happen that we get from the backend the dict as either an object, or stringified object (with escaped characters if my memory serves me well.)

src/UI/Components/ServiceInstanceForm/Helpers/createFormState.spec.ts
```diff
@@ -20,7 +88,21 @@ test("GIVEN createFormState WHEN passed a dict field with a default value THEN c
   const formState = createFormState(fields);
 
   expect(formState).toEqual({
-    [Field.dictionary.name]: "{}",
+    [Field.dictionary.name]: {},
+  });
+});
+
+test("Given creteFormState WHEN passed nested fields state THEN creates formState correctly", () => {
```
LukasStordeur, line 95:
```suggestion
test("Given createFormState WHEN passed nested fields state THEN creates formState correctly", () => {
```

src/UI/Components/ServiceInstanceForm/Helpers/createFormState.ts
```diff
@@ -93,9 +101,22 @@ export const createEditFormState = (
         case "Textarea":
         case "TextList":
         case "Text": {
-          acc[curr.name] = curr.type.includes("dict")
-            ? stringifyDict(originalAttributes?.[curr.name])
-            : cloneDeep(originalAttributes?.[curr.name]);
+          if (curr.type === "int?" || curr.type === "float?") {
```
LukasStordeur, line 104: I'm confused, if we have a Text field, or TextList, how is it we have the possibility that we are dealing with integers or floats? is it purely to cover the empty state `""`?

src/UI/Components/ServiceInstanceForm/Helpers/createFormState.ts
```diff
@@ -93,9 +101,22 @@ export const createEditFormState = (
...
+            const value = originalAttributes?.[curr.name];
+
+            //empty string assertion is for converting input form to JSON Editor
```
LukasStordeur, line 107: This is not clear to me. 
The JSON editor is relying on the JSON-Schema. 
We cannot assign null like that there when the field is empty. 
I feel like I'm missing the actual problem.

src/UI/Components/ServiceInstanceForm/Helpers/createFormState.ts
```diff
@@ -173,9 +194,22 @@ export const createDuplicateFormState = (
         case "Textarea":
         case "TextList":
         case "Text": {
-          acc[curr.name] = curr.type.includes("dict")
-            ? stringifyDict(originalAttributes?.[curr.name])
-            : cloneDeep(originalAttributes?.[curr.name]);
+          if (curr.type === "int?" || curr.type === "float?") {
```
LukasStordeur, line 197: This feels redundant, and not optimized. 
The formState is a complex function, adding even more complexity will make it hard to read.

LukasStordeur (commented):
Hmm, Not entirely certain we are approaching the problem correctly.

src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx
```diff
@@ -223,9 +223,18 @@ export const FieldInput: React.FC<Props> = ({
             !isNew
           }
           type={field.inputType}
-          handleInputChange={(value, _event) =>
-            getUpdate(makePath(path, field.name), value)
-          }
+          handleInputChange={(value, _event) => {
```
LukasStordeur, line 226: We are doing this manipulation in other places too, checking if something is parseable to JSON. I'd suggest the following and reuse it where we are doing this try catch :

```ts
const handleInputChange = (value, _event) => {
  if (field.type.includes("dict")) {
    getUpdate(makePath(path, field.name), tryParseJSON(value));
  } else {
    getUpdate(makePath(path, field.name), value);
  }
};

const tryParseJSON = (value) => {
  try {
    return JSON.parse(value);
  } catch (_error) {
    // If the value is not a valid JSON string, return the original value
    return value;
  }
};
```

matborowczyk:
I addressed all remarks and according to our call.

Improved test cases to better reflect the issue, unified test as code added was extracted and reused so there is no room for only on formstate creator to success/fail in this scenario

I checked other inputs and no other place is parsingJSON or work with the input in similar fashion where that parsing could be implemented but reused it in conversion.

PR #6283, opened 2025-04-11
Update callback Queries

### LINKED ISSUE #6282: Update Callbacks queries

matborowczyk:
# Description

Update callback Queries

closes #6282 

# Self Check:

- [ ] Changelog entry


src/Slices/ServiceDetails/UI/Tabs/Callbacks/CreateCallbackForm.tsx
```diff
@@ -8,49 +8,71 @@ import {
...
+   * This function gather the input data and calls the mutation function to create a callback.
+   * It resets the error state before attempting to create the callback.
+   */
+  const onCreate = () => {
+    setError(null);
+    mutate({
+      callback_url: url || "",
+      callback_id: id || undefined,
+      service_entity,
+      minimal_log_level: logLevel === null ? undefined : (logLevel as LogLevelString),
```
LukasStordeur, line 73: Why do we need the casting here?

matborowczyk: because select box convert logLevel from `LogLevelString` to `string` and to keep typings align I opted for what was already there

LukasStordeur: kay

LukasStordeur (approved):

cypress/e2e/scenario-2.1-basic-service.cy.js
```diff
@@ -193,7 +192,10 @@ if (Cypress.env("edition") === "iso") {
         .within(() => {
           cy.get('[data-label="version"]').trigger("click"); //it's done to avoid flake where the tooltip comes in a way and click ins't triggered
         });
-      cy.get('[data-testid="selected-version"]').should("have.text", "Version: 2");
+      cy.get('[data-testid="selected-version"]', { timeout: 30000 }).should(
```
LukasStordeur, line 195: I'm not convinced increasing the timeout will resolve your issue

matborowczyk: I would like to give it a chance, this part of the code is tied to the continuous query, so I think default timeout of 10s could be not enough. 

2 on 3 passed after 4/4 failures, so I would run it few more times to assert it. 

8th run was good but I did not remove the command to merge unit test, that's why it shows it as failed run.

If result won't be consistent I will give it more thought at monday as I don't have other flow for this scenario in my head at the moment

I won't merge it until you will be back at Tuesday, so we can discuss it if you will be in doubt

PR #6293, opened 2025-04-15
Fix Encoding issues in the Resource Page

### LINKED ISSUE #6292: fix encoding url parameters

matborowczyk:
# Description

Fix Encoding issues in the Resource Page as well as make sure that all potential values that could encounter similar issue is addressed.

Previously encoding was done at the lower level due to the structure of the Query Manager, currently due to more granular approach I did not see a easy way to create global middle step to address all parameters, but I'm open for any suggestions

closes #6292 

# Self Check:

- [ ] Changelog entry


src/Data/Managers/V2/helpers/utils.ts
```diff
@@ -0,0 +1,11 @@
+export function urlEncodeParams<Type extends object>(params: Type) {
```
LukasStordeur, line 1: docstrings. 
File name could be the same as the method considering it only contains that.

LukasStordeur (commented):
hmm, for now, I guess that will do. Perhaps there's some sort of settings we can use for the fetch to always encode it correctly.

LukasStordeur (approved):

PR #6306, opened 2025-04-23
Merge Environment, Project and Server Status Queries

matborowczyk:
Final PR for Merging Environment, Project and Server Status Queries

closes #6235 

# Self Check:

- [ ] Changelog entry


src/Slices/Settings/UI/Tabs/Configuration/Provider.tsx
```diff
@@ -54,25 +70,29 @@ export const Provider: React.FC<Props> = ({ settings: { settings, definition } }
...
-  const { commandResolver } = useContext(DependencyContext);
-  const updateSetting = commandResolver.useGetTrigger<"UpdateEnvironmentSetting">({
-    kind: "UpdateEnvironmentSetting",
+  const updateSetting = useUpdateEnvironmentSetting({
+    onError: (error) => setErrorMessage(error.message),
   });
-  const resetSetting = commandResolver.useGetTrigger<"ResetEnvironmentSetting">({
-    kind: "ResetEnvironmentSetting",
+  const resetSetting = useResetEnvironmentSetting({
+    onError: (error) => setErrorMessage(error.message),
```
LukasStordeur, line 77: does the error get cleared on success? Or does it stay on the screen?

src/Test/Inject/dependencies.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import { FileFetcher, FlatEnvironment } from "@/Core";
+import { AuthContextInterface, PrimaryArchiveHelper, defaultAuthContext } from "@/Data";
+import { MockEnvironmentHandler, MockFeatureManager, MockFileManager } from "@/Test/Mock";
+import { PrimaryRouteManager, useEnvironmentModifierImpl, DependencyProvider } from "@/UI";
+import { UrlManagerImpl } from "@/UI/Utils";
+import { EnvironmentDetails } from "../Data";
+
+export const MockedDependencyProvider: React.FC<
```
LukasStordeur, line 9: docstrings

src/Test/Mock/MockEnvironmentHandler.ts
```diff
@@ -1,16 +1,20 @@
 import { EnvironmentHandler, FlatEnvironment } from "@/Core";
 
-export function MockEnvironmentHandler(environment: string): EnvironmentHandler {
+export function MockEnvironmentHandler(environment: FlatEnvironment): EnvironmentHandler {
```
LukasStordeur, line 3: docstrings

src/Test/Utils/base-setup.tsx
```diff
@@ -1,71 +1,31 @@
...
+import { StaticScheduler, DeferredApiHelper, MockedDependencyProvider } from "@/Test";
 import { BlockingModal } from "@/UI/Components";
-import { DependencyProvider } from "@/UI/Dependency";
-import { PrimaryRouteManager } from "@/UI/Routing";
+import * as envModifier from "@/UI/Dependency/EnvironmentModifier";
 import { TestMemoryRouter } from "@/UI/Routing/TestMemoryRouter";
 import { testClient } from "./react-query-setup";
 
 export function baseSetup(Page: React.ReactNode, halted: boolean = false) {
+  jest.spyOn(envModifier, "useEnvironmentModifierImpl").mockReturnValue({
```
LukasStordeur, line 12: this will be removed at some point after the refactoring?

src/UI/Components/InlineEditable/EditableField.tsx
```diff
@@ -45,29 +48,26 @@ export const EditableField: React.FC<Props> = ({
...
-
-    if (Maybe.isSome(error)) {
-      setSubmitError(error.value);
-    }
+    onSubmit(value);
   };
   const onEditClick = () => {
     setEditable(true);
-    setSubmitError("");
+    setError(null);
```
LukasStordeur, line 63: does this mean that whenever the user clicks on the field, the error gets cleared?

src/UI/Dependency/Dummy/DummyFeatureManager.ts
```diff
@@ -40,4 +40,7 @@ export class DummyFeatureManager implements FeatureManager {
   getLicenseInformation(): StatusLicense {
     throw new Error("Method not implemented.");
   }
+  setServerStatus(_serverStatus: ServerStatus): void {
```
LukasStordeur, line 43: why is serverStatus part of the featureManager?

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
```diff
@@ -1,196 +1,283 @@
...
-                    onSelectEnvironment={onSelectEnvironment}
-                    selectedEnvironment={Environment.filterable[0]}
-                  />
-                </AuthTestWrapper>
-              </AuthProvider>
-            </StoreProvider>
-          </DependencyProvider>
+        <QueryClientProvider client={testClient}>
+          <StoreProvider store={store}>
+            <AuthProvider config={config}>
```
LukasStordeur, line 51: is this meant to be outside the AuthTestWrapper? (this is out of scope I know)

src/UI/Root/Components/Header/EnvSelector/Provider.tsx
```diff
@@ -1,21 +1,27 @@
...
+  const { environmentHandler, routeManager, featureManager } = useContext(DependencyContext);
   const location = useLocation();
   const navigate = useNavigate();
   const selected = environmentHandler.useSelected();
-  const [data] = queryResolver.useOneTime<"GetEnvironments">({
-    kind: "GetEnvironments",
-    details: false,
-  });
+
+  const envDetails = true;
```
LukasStordeur, line 22: boolean variables should reflect boolean aspect

src/UI/Root/Root.test.tsx
```diff
@@ -1,138 +1,97 @@
...
+  MockedDependencyProvider,
   Project,
   ServerStatus,
-  StaticScheduler,
 } from "@/Test";
-import { DependencyProvider, EnvironmentHandlerImpl } from "@/UI/Dependency";
 import { TestMemoryRouter } from "../Routing/TestMemoryRouter";
 import { Root } from "./Root";
 
+jest.spyOn(defaultAuthContext, "getToken").mockReturnValue("mocked_token");
```
LukasStordeur, line 20: Hmmm, it didn't require mocking before, is this an unwanted side-effect that needs to be checked out?

LukasStordeur (changes requested):
refetch interval needs to be a const, I believe we already have one or discussed it in previous PR's

LukasStordeur (commented):
There are also two new errors/warnings in the logs from the unit-tests

LukasStordeur:
The code-coverage is also going down quite a lot, this should be adressed in a second stage to not lose too much.
(left is master, right is this PR.)
image

src/Data/Managers/V2/Environment/GetEnvironmentDetails/useGetEnvironmentDetails.ts
```diff
@@ -0,0 +1,41 @@
...
+        select: (data) => data.data,
+      }),
+
+    useContinuous: (id: string): UseQueryResult<Environment, Error> =>
+      useQuery({
+        queryKey: ["get_environment_details-continuous", id],
+        queryFn: () => get(`/api/v2/environment/${id}?details=true`),
+        retry: false,
+        select: (data) => data.data,
+        refetchInterval: 5000,
```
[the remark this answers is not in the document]
matborowczyk: I had to miss those

src/UI/Components/InlineEditable/EditableField.tsx
matborowczyk: Yes, as it was so far

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
matborowczyk: Hmm, I took a look on it and its wrapper on MockedDependencyProvider and it's in correct place

src/UI/Dependency/Dummy/DummyFeatureManager.ts
matborowczyk: all the features are from the server status, previously it was accessed through store, which is not longer the case so we need to set it there

matborowczyk:
I mentioned that in previous PR but it could get missed, I am ware of warning I will address those with dryRun thing, there is also re-rendering errors that are due to issue with envHandler and modifier. 

I'm addressing drop in the coverage with removal of unused code

src/Test/Utils/base-setup.tsx
LukasStordeur: mocking it like this, could hide some bugs over time, so if we can remove it in a later stage, I think it would be safer

src/UI/Components/InlineEditable/EditableField.tsx
LukasStordeur: Can you make a ticket to review this behaviour? I'm not entirely certain this is UX friendly

src/UI/Dependency/Dummy/DummyFeatureManager.ts
LukasStordeur: hmmm, why do we need to set the entire serverStatus in the featureManager? It doesn't make sense from a logical point of view, to have the full serverStatus under a featureManager. It's not coherent or predictable to be here.

LukasStordeur:
> I mentioned that in previous PR but it could get missed, I am ware of warning I will address those with dryRun thing, there is also re-rendering errors that are due to issue with envHandler and modifier.
> 
> I'm addressing drop in the coverage with removal of unused code

Okay, I thought it safe to mention there were new ones bessides the dryRun queries :D

src/Test/Utils/base-setup.tsx
matborowczyk: but how does that differ compared to mocking like we used to? 

I'm totally open to change that but I guess that if we  have modifier tested then here is no room for hidden bugs as we use an actual envModifier with only one function mocked.

src/UI/Components/InlineEditable/EditableField.tsx
matborowczyk: https://github.com/inmanta/web-console/issues/6330

src/UI/Dependency/Dummy/DummyFeatureManager.ts
matborowczyk: how else we want to access features that are stored in the serverStatus? I can extract those that are used, but so far  the logic of the app was that app fetched server status, then at first time when some data from FeatureManager is accessed manager tries to get servers status, if there is none then it crashes the app with error. 

I think we can refactor FeatureManager to be hook on it's own to fetch the status and store it's features but that's beyond the scope and I can add it to the queue

src/Test/Utils/base-setup.tsx
LukasStordeur: I'm probably missing some pieces, but for me, the "new" implementation should only require mocking external libraries like jointJs, or monaco, and the api responses with MSW to build the page for the test setup. 
The page/test suit we have, that use the full page as context, are more integration tests than unit tests. And yes, the Modifier might be covered in a unit case, but it should work in a full integration too. This is why I'm making this comment, but we can review this in a later stage.

matborowczyk: okay, now I see your point of view more clearly. Makes totally sense.  I here firstly went with the flow as it was that some of these managers/handlers are mocked. then that parameter has to go and we need to add extra call for the environment to access halting setting. 
fine by me,I added that comment to the task about

src/UI/Dependency/Dummy/DummyFeatureManager.ts
matborowczyk: https://github.com/inmanta/web-console/issues/6331

src/Data/Managers/V2/Miscellaneous/GetVersionFileInfo/useGetVersionFileInfo.ts
LukasStordeur, line 1: I'm going to make the same remark as before, but I'm not a fan of the word Miscellaneous. It doesn't mean anything. We can afford specific folders for these queries too and keep the overview readable and coherent

src/Injector.tsx
```diff
@@ -69,7 +67,7 @@ export const Injector: React.FC<React.PropsWithChildren<Props>> = ({ store, chil
   const commandResolver = new CommandResolverImpl(new CommandManagerResolverImpl(store, apiHelper));
   const urlManager = new UrlManagerImpl(featureManager, baseUrl);
   const fileFetcher = new FileFetcherImpl(apiHelper);
-  const environmentModifier = EnvironmentModifierImpl();
+  const environmentModifier = useEnvironmentModifierImpl();
```
LukasStordeur, line 70: We are keeping the name EnvironmentModifierImpl? Is there nothing less cryptic?

src/Slices/Dashboard/UI/Page.tsx
```diff
@@ -1,39 +1,46 @@
...
 export const Page: React.FC = () => {
-  const { queryResolver, environmentHandler } = useContext(DependencyContext);
+  const { environmentHandler } = useContext(DependencyContext);
 
-  const [envData, retry] = queryResolver.useOneTime<"GetEnvironmentDetails">({
-    kind: "GetEnvironmentDetails",
-    details: true,
-    id: environmentHandler.useId(),
-  });
+  const envName = environmentHandler.useSelected()?.name || ""; //useSelected() theoretically could return undefined but as we use useId() the view would throw an error in that case
```
LukasStordeur, line 19: sounds like somethings not typed correctly to me

src/Slices/Settings/UI/Tabs/Configuration/Container.tsx
```diff
@@ -8,28 +8,20 @@ import { InputRow } from "./Components";
 interface Props {
   infos: EnvironmentSettings.InputInfo[];
   errorMessage: string;
+  showUpdateBanner: boolean;
```
LukasStordeur, line 11: Can you move it down to group it with setShowUpdateBanner?

src/Slices/Settings/UI/Tabs/Configuration/Provider.tsx
```diff
@@ -53,33 +70,66 @@ export const Provider: React.FC<Props> = ({ settings: { settings, definition } }
...
-  const resetSetting = commandResolver.useGetTrigger<"ResetEnvironmentSetting">({
-    kind: "ResetEnvironmentSetting",
+
+  const resetSetting = useResetEnvironmentSetting({
+    onError: (error) => setErrorMessage(error.message),
   });
+
   const handleReset = (id: string) => {
     dispatch({ type: "reset", payload: id });
+    if (errorMessage) {
```
LukasStordeur, line 96: is the if-case really necessary?  i would have added it the same way you did for the onError, to keep it consistent

src/Slices/Settings/UI/Tabs/Environment/Tab.tsx
```diff
@@ -1,39 +1,60 @@
...
+        ariaLabel="EnvironmentDetails-Error"
+      />
+    );
+  }
+  if (projectsData.isError) {
+    return (
+      <ErrorView
+        message={projectsData.error.message}
+        retry={projectsData.refetch}
+        ariaLabel="EnvironmentDetails-Error"
```
LukasStordeur, line 39: same aria-label as the other one?

src/UI/Components/UpdateBanner/UpdateBanner.tsx
```diff
@@ -1,27 +1,26 @@
...
 import { DependencyContext } from "@/UI/Dependency";
 import { words } from "@/UI/words";
 
-interface Props {
-  apiHelper: ApiHelper;
-}
-
-export const UpdateBanner: React.FunctionComponent<Props> = (props) => {
+/**
+ * Component for displaying an update banner when a new version is available.
```
LukasStordeur, line 8: would be nice to be complete here and specify this is about the app version. For someone new to this codebase, it might quickly become confusing with all the different version-terms available.

src/Slices/Settings/UI/Tabs/Environment/Tab.tsx
matborowczyk: from the user perspective it's the same view, the same component, so I thought there is no need in changing that. the same approach is done in Service Instance detail view, home view and few other

src/Injector.tsx
matborowczyk: it will be addressed in the graphQL iteration

matborowczyk:
Screenshot 2025-05-07 at 15 10 06
oss test succeed

matborowczyk:
Screenshot 2025-05-08 at 09 31 32
keycloak as well

PR #6310, opened 2025-04-24
Issue/6231 environment queries part3

matborowczyk:
# Description

Part 3 of the env update, all remaining logic implementation, it doesn't include tsc fixes for test files it assures that app works as it was and E2E test passes

# Self Check:

- [ ] Changelog entry


matborowczyk:
in the provider there is minor change, as previously projectNames were merged to the environment on the stateHelper level, now we assign it in where it's needed. to minimise the already big influence across app I opted for that

Added docstrings where there were missing, or at least tried to add them everywhere possible, at the top level

Removed one interface which was essentially duplication of FlatEnvironment, I will create ticket to investigate interfaces there as there is room to improvement in the logic related Envs. merging Handler with modifier, cleaning up providers, and initialziators, etc. but it has to be discussed

src/Data/Managers/V2/Environment/UpdateEnvironmentSetting/useUpdateEnvironmentSetting.ts
```diff
@@ -30,13 +30,19 @@ export const useUpdateEnvironmentSetting = (
     mutationFn: ({ id, value }) => post(`/api/v2/environment_settings/${id}`, { value }),
     mutationKey: ["update_environment_setting"],
     onSuccess: () => {
-      client.invalidateQueries({
+      client.refetchQueries({
```
LukasStordeur, line 33: Do we need to have it repeated like that? Can't we pass an array to this method?

src/Test/Data/Environment.ts
```diff
@@ -7,6 +18,7 @@ export const a: EnvironmentModel = {
   repo_branch: "",
   repo_url: "",
   halted: false,
+  settings: {},
```
LukasStordeur, line 21: I'd at least add one case that has settings

src/Slices/Settings/UI/Tabs/Environment/Components/ErrorHandlingContext.tsx
LukasStordeur, line 1: I'm not entirely followin why we'd need this. 

You want to prevent prop-drilling on input fields, but by doing that, you remove some of the scopedness of these inputfields, and tying them explicitly to the environment settings. I don't know if that's an issue, but I can see that the input components affected are located in the /Component folder, so it means they should be considered global/shareable over multiple views.

Why do you want to share the same error message?
What would be the prop-drilling you are talking about? Addidng a "error-message" prop to an input field is quite normal, and I'd personally wouldn't consider it prop-drilling. I can only find two references that use this context. 

If the reason is related to React Query itself, I think there's a misunderstanding somewhere as to why we'd need this strategy, while ReactQuery itself acts as a stateManager too.

src/Data/Managers/V2/Environment/UpdateEnvironmentSetting/useUpdateEnvironmentSetting.ts
matborowczyk: unfortunately putting more queries would result with no refetches at all as queryKey matcher would look for a query that have all keys in the array. I think I said it few weeks/moths back that I spent way too much time finding why something was refetching as intended :D Learnt hard way :D 
I think what we could do is to add one, more general key to the queries to reset/refetch them at once.

src/Slices/Settings/UI/Tabs/Environment/Components/ErrorHandlingContext.tsx
matborowczyk: Now as I read I think reinitialising query could be more fitted there, I didn't think of that approach. but that will further localise those components.

Previously V1 query was passed down and resolution of error path was done in the input. react query doesn't work like that, and I could either create a context that would be reached in the component or pass down error and function to clear the error from the env settings component 

I did not want to rebuild all of the components and leaned towards something that was pretty fast and left room for this object to be global, even though they are tied only to Environments Settings.

src/Data/Managers/V2/Environment/UpdateEnvironmentSetting/useUpdateEnvironmentSetting.ts
LukasStordeur:
```
await queryClient.refetchQueries({ type: 'active' })
await queryClient.refetchQueries({ stale: true })
```
Perhaps in our case, we can use one of either? 

I also think the way we define the querykeys might not be ideal. Food for thoughts, but using the factory system with some sort hierarchy, and cutting down the keys in smaller key-words instead of concatenating it into one string would be better.` "environement", "get", "continuous" "one-time"` should probably be separated from the core key. You'd have a few strings in your array, but it would allow to invalidate/refetch more efficiently and with less repetition. You could then do the partial key match like suggested in the docs.

```
// refetch all active queries partially matching a query key:
await queryClient.refetchQueries({ queryKey: ['environment'], type: 'active' })
```

src/Slices/Settings/UI/Tabs/Environment/Components/ErrorHandlingContext.tsx
LukasStordeur: This workaround is for me a red flag. There's something wrong somewhere, and this is a temporary bandaid on top of it, that adds complexity, and goes against the principles of encapsulation and separation of concern. The input fields shouldn't care about React Query logic. They should be stupid enough to just accept an attribute telling them to show an error message if there is one.

src/Data/Managers/V2/Environment/UpdateEnvironmentSetting/useUpdateEnvironmentSetting.ts
```diff
@@ -30,15 +30,9 @@ export const useUpdateEnvironmentSetting = (
...
-        queryKey: ["get_environment_settings-one_time"],
-      });
-      client.invalidateQueries({
-        queryKey: ["get_environment_details-one_time"],
-      });
-      client.invalidateQueries({
-        queryKey: ["get_environment_details-continuous"],
-      });
+      client.refetchQueries({ type: "active" });
+      client.refetchQueries({ stale: true });
```
LukasStordeur, line 34: I don't think you need both to be honest, I used either as examples

src/UI/Root/Components/PageFrame/SettingsProvider.tsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import { useGetEnvironmentSettings } from "@/Data/Managers/V2/Environment";
+
+export const SettingsProvider: React.FC = () => {
```
LukasStordeur, line 4: 🤔 how is this beneficial compared to using the ReactQuery directly where we need it?

src/Test/Data/Environment.ts
```diff
@@ -1,4 +1,35 @@
-import { EnvironmentExpertOnly, EnvironmentModel } from "@/Core";
+import { EnvironmentExpertOnly, EnvironmentModel, FlatEnvironment } from "@/Core";
+
+export const env: FlatEnvironment = {
```
LukasStordeur, line 3: Is it right, that you will add a test later for this then? Or where is this env imported :D

PR #6316, opened 2025-04-25
6131 login page

### LINKED ISSUE #6131: Login page UI adjustments
With the PF migration, the login page isn't in line anymore with the suggested  PF version. 
Also, the routing has some issues, This page is at /login instead of /console/login so that means a refresh kills the app

LukasStordeur:
# Description

- Repair the redirection issues, and cover trailing slashes to prevent not found pages. 
- Updated the Login page to use the Patternfly component instead; This means we lose the "show/hide" functionality on the password field. 
- Updated the button in the userManagement modal to not use the Content element inside the button, as Content elements by default have black text, and that the button doesn't plain text to be wrapped.

closes #6131 

https://github.com/user-attachments/assets/639dc5b6-c02d-4680-b5ea-a7c656eedf0f

image

image

image


src/Data/Auth/Providers/DatabaseAuthProvider.tsx
```diff
@@ -10,16 +11,23 @@ import { AuthContext } from "../AuthContext";
 export const DatabaseAuthProvider: React.FC<React.PropsWithChildren> = ({ children }) => {
   const [user, setUser] = useState<string | null>(null);
   const navigate = useNavigate();
+  const baseUrlManager = new PrimaryBaseUrlManager(
```
matborowczyk, line 14: It's optional as I don't know how much work that would add but maybe we could ad `navigateTo("Login", undefined);` there, I don't remember if the path is there but that would align with what we have in other places.

src/Slices/Login/UI/LoginForm.tsx
```diff
@@ -32,18 +18,19 @@ interface Props {
...
  * @returns {React.FC<Props>} The rendered component.
  */
-export const LoginForm: React.FC<Props> = ({
-  submitButtonText,
-  submitButtonLabel = "login-button",
-}) => {
+export const LoginPageComponent: React.FC<Props> = ({ submitButtonText }) => {
   const { authHelper } = useContext(DependencyContext);
   const navigate = useNavigate();
+  const baseUrlManager = new PrimaryBaseUrlManager(
```
matborowczyk, line 24: `navigateTo("Home", undefined);` could work here

matborowczyk (changes requested):

src/Data/Auth/Providers/DatabaseAuthProvider.tsx
LukasStordeur: Was my first thought too, and I tried it, but to work properly, the login page needs to be outside the RouteOutlet. It also means we can't add it to the PageDictionary/PrimaryPageManager directly. The RouteOutlet makes use of the Initializer, which expects the environement endpoint to be called, resulting the login page to also require that endpoint to return something else than a 401. You end up in a loop. To change this behaviour, would require a lot of work compared to the simpler approach of keeping the Login Route outside the dictionary of pages.

This also responds to the navigateTo useage, which requires the page to be part of the PrimaryPageManager :)

matborowczyk (approved):

PR #6325, opened 2025-04-29
Final partial PR - test fixes and changes, lint, tsc fixes

matborowczyk:
Final partial PR for the Env Queries

I'm aware that there are few new warning in the logs, which will be addressed afterwards in the separate PR as I will tackle dryRun warnings

As I said on the call There are also bad setState errors, even though functionality of the app remains, that will be also addressed later on, either with GraphQL task as it will once again address Environment calls.

Those re-renders/bas setState is due the sub-optimal flow as we now doubling the state. and manouver from react-query to those environmentHandler/Modifier. I think we could further extract it to separate component to have layer that will fetch environment related data and provide it down the app, but that should be discussed

Beside re-rendering there is an issues of Expert-mode and one e2e test case failed which I will address tomorrow, as my mind is fried for today.

I will here also address drop in coverage, probably noticable, it's due that some queries got introduced recently without dropping old one, here is the same. all removal will be done also in separate PR

Screenshot 2025-04-29 at 15 36 12


src/UI/Components/InlineEditable/EditableTextField.tsx
```diff
@@ -53,5 +53,7 @@ const EditView: EditViewComponent = ({ value, onChange, onSubmit, label }) => (
 );
 
 const StaticView: StaticViewComponent = ({ value, ...props }) => (
-  <InlineValue {...props}>{value}</InlineValue>
+  <InlineValue role="textbox" {...props}>
```
LukasStordeur, line 56: why are we setting the role here?

LukasStordeur (commented):
I presume the e2e will be fixed in the next one?

src/UI/Components/InlineEditable/EditableTextField.tsx
matborowczyk: axe assertion in Env Settings scenario was complaining that this div has aria-label but no role assigned

src/Test/Data/Environment.ts
```diff
@@ -3,11 +3,11 @@ import { EnvironmentExpertOnly, EnvironmentModel, FlatEnvironment } from "@/Core
 export const env: FlatEnvironment = {
   id: "env",
   name: "environment_name_env",
-  project_id: "project_id_env",
+  project_id: "project_id_b",
```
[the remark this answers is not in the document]
matborowczyk: I think it should be fine, this is this new env that I added that you asked about usage in tests. I changed id to fit scenario, I didn't want add project just for that one tests, and in practice multiple environment could have one project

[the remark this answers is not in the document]
LukasStordeur: okay

src/UI/Components/InlineEditable/EditableTextField.tsx
LukasStordeur: yeah, but role textbox is for an input element, this is a div. It shouldn't have that role.

matborowczyk: I could not find better fitting role for this component, and it's a preview of value for input component but I understand your remark. I change the tag of the component

matborowczyk: I looked for patternfly Content component but this is also div without role, paragraph won't work  and I think this is closest role that could be there. I can create ticket to rework those components to be proper inputs with read-only view instead of what it currently is. what do you think about it?

LukasStordeur: Why are we passing an aria-label to a div to begin with?

LukasStordeur: sure make a ticket

matborowczyk: https://github.com/inmanta/web-console/issues/6328

LukasStordeur (approved):

PR #6327, opened 2025-04-30
Issue/6231 env queries e2e fix

matborowczyk:
E2E fix for env queries update

all errors were due to the fact of envSettings not being set even though environment have the settings on it's own

This is temporary solution as envModifier and Handler have to be rework with GraphQL task as some of the queries won't be used anymore or we can initialization of the app do in one go.

EnvSettings Provider won't work as it use to work in old state manager as it was causing re-renders as now there are multiple sources of through that dynamically updates the state beside banner flickering when toggling expertMode.

Generally re-rendering that now is happening was not noticable in the usage of the app it for sure but negativaly impacts performance and is not desired behaviour.

the new design for env handler and modifier has to be disscussed


src/UI/Dependency/EnvironmentModifier.tsx
```diff
@@ -36,10 +36,14 @@ export function useEnvironmentModifierImpl(): EnvironmentModifier {
   }
 
   function useSetting(settingName: keyof EnvironmentSettings.DefinitionMap): boolean {
```
LukasStordeur, line 38: Can you add some comments about the problems we are facing now? To keep them in the code.

matborowczyk: added mention in the docstrings for this function, created task to resolve warning and errors where I mentioned re-rendering and mentioned that issue in the GraphQL task  and linked this task to the docstrings

src/Test/Inject/dependencies.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import { FileFetcher, FlatEnvironment } from "@/Core";
+import { AuthContextInterface, PrimaryArchiveHelper, defaultAuthContext } from "@/Data";
+import { MockEnvironmentHandler, MockFeatureManager, MockFileManager } from "@/Test/Mock";
+import { PrimaryRouteManager, useEnvironmentModifierImpl, DependencyProvider } from "@/UI";
+import { UrlManagerImpl } from "@/UI/Utils";
+import { EnvironmentDetails } from "../Data";
+
+export const MockedDependencyProvider: React.FC<
```
LukasStordeur, line 9: docstrings

src/Test/Mock/MockEnvironmentHandler.ts
```diff
@@ -1,16 +1,20 @@
 import { EnvironmentHandler, FlatEnvironment } from "@/Core";
 
-export function MockEnvironmentHandler(environment: string): EnvironmentHandler {
+export function MockEnvironmentHandler(environment: FlatEnvironment): EnvironmentHandler {
```
LukasStordeur, line 3: docstrings

src/Test/Utils/base-setup.tsx
```diff
@@ -1,71 +1,31 @@
...
+import { StaticScheduler, DeferredApiHelper, MockedDependencyProvider } from "@/Test";
 import { BlockingModal } from "@/UI/Components";
-import { DependencyProvider } from "@/UI/Dependency";
-import { PrimaryRouteManager } from "@/UI/Routing";
+import * as envModifier from "@/UI/Dependency/EnvironmentModifier";
 import { TestMemoryRouter } from "@/UI/Routing/TestMemoryRouter";
 import { testClient } from "./react-query-setup";
 
 export function baseSetup(Page: React.ReactNode, halted: boolean = false) {
+  jest.spyOn(envModifier, "useEnvironmentModifierImpl").mockReturnValue({
```
LukasStordeur, line 12: will we remove the need for this after the refactoring is done?

src/UI/Components/InlineEditable/EditableField.tsx
```diff
@@ -60,6 +60,7 @@ export const EditableField: React.FC<Props> = ({
   };
   const onEditClick = () => {
     setEditable(true);
+    setError(null);
```
LukasStordeur, line 63: So if I click again on the input field, the error is gone? 🤔 Doesn't sound like a correct UX

src/UI/Dependency/DependencyResolver.tsx
```diff
@@ -9,8 +9,11 @@ interface Props {
 export const DependencyResolver: React.FC<Props> = ({ environment }) => {
   const { fileFetcher, environmentModifier } = useContext(DependencyContext);
 
-  fileFetcher.setEnvironment(environment.id);
-  environmentModifier.setEnvironment(environment);
+  useEffect(() => {
+    fileFetcher.setEnvironment(environment.id);
+    environmentModifier.setEnvironment(environment);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [environment]);
```
LukasStordeur, line 16: why are we not including the fileFetcher? Or the environmentModifier?

src/UI/Root/Components/Sidebar/EnvironmentControls/EnvironmentControls.test.tsx
```diff
@@ -1,33 +1,16 @@
...
-  DeferredApiHelper,
-  dependencies,
-  DynamicCommandManagerResolverImpl,
-  DynamicQueryManagerResolverImpl,
-  EnvironmentDetails,
-  MockEnvironmentHandler,
-  StaticScheduler,
-} from "@/Test";
-import { DependencyProvider } from "@/UI/Dependency";
+import { HttpResponse } from "msw";
```
LukasStordeur, line 7: group imports

LukasStordeur (changes requested):

PR #6337, opened 2025-05-08
Enable authentication on documentation links

### LINKED ISSUE #6335: add auth token in doc url

LukasStordeur:
# Description

Enable passing a token in the url for the authentication links when a token is available. 

closes #6335


src/UI/Root/Components/Header/Actions/DocumentationLinks.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  getGeneralAPILink(): string {
+    return API_LINK;
+  }
+  getLSMAPILink(): string {
+    return LSM_API_LINK;
+  }
+  getApiUrl(): string {
+    return API_URL;
+  }
+  setEnvironment(): void {
```
matborowczyk, line 27: I think that's not required in the UrlManager

matborowczyk (changes requested):
I merged my PR,  I think you would have to change `DependencyProvider ` to `MockedDependencyProvider` and you could use `jest.spyOn(defaultAuthContext, "getToken").mockReturnValue("mocked_token");` instead mocking whole AuthHelper, but that's up to preference

LukasStordeur:
> I merged my PR, I think you would have to change `DependencyProvider ` to `MockedDependencyProvider`

Why? These are unit tests and only require the two dependencies to be available. On top of it, the `MockedDependencyProvider` doesn't allow passing overrides for specific dependencies like the `urlManager`. The tests work fine with just `DependencyProvider `

matborowczyk (approved):

PR #6343, opened 2025-05-09
Fix 403 re-direct and env-settings query

### LINKED ISSUE #6342: 403 should not redirect to login page
The only response that should trigger the redirect, is a 401.
403 should display the error message to the user returned by the api. 

This used to be implemented in the baseApiHelper

matborowczyk:
# Description

It turns out that useFetchHelper was reaching to wrong dependency for being able reach auth at the Injector level

I fixed also redirect issue, as I was already at this files

closes #6342 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/6342-enable-fixes.yml
+++ b/changelogs/unreleased/6342-enable-fixes.yml
@@ -1,0 +1,5 @@
+description: redirect to login when only when 401 is returned, fix auth issue in environment settings query
+issue-nr: 6342
+change-type: patch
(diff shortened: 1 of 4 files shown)


matborowczyk:
> do we still need the baseApiHelper? I thought it was part of the old QueryManagement?

There is file Fetcher which is independent dependency in the provider, I will create a task to update and remove apiHelper

LukasStordeur (approved):

PR #6346, opened 2025-05-09
6061 logviewer compile report

### LINKED ISSUE #6061: Upgrade Codehighlighter to PF Log-viewer
Move to https://www.patternfly.org/extensions/log-viewer/react-demos
For the Log viewers and expandable code highlighters.

LukasStordeur:
# Description

Implementation of the Patternfly LogViewer component in the Compile details view. This replaces the collapsible sections that previously displayed each stage of the compile. 

A mock of the component has been added to the unit-tests. 

closes #6061 

https://github.com/user-attachments/assets/2359a5fd-bc84-4984-9c67-fd46ae318c95


src/UI/Components/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,191 @@
...
+  logs: LogViewerData[];
+}
+
+/**
+ * LogViewer component displays logs with a dropdown to select log parts and a download button.
+ * Uses PatternFly's LogViewer for rendering logs.
+ * @param logs Array of log objects, each with data, id, name, duration, and failed status.
+ */
+export const LogViewerComponent: React.FC<LogViewerProps> = ({ logs }) => {
+  const [selectedLogId, setSelectedLogId] = useState(logs[0]?.id || "");
```
matborowczyk, line 40: shouldn't we provide empty view if there is no logs?

matborowczyk (changes requested):

src/UI/Components/LogViewer/LogViewer.tsx
LukasStordeur: The safe operator is there because we are dealing with an array, but the check is already there at a higher level to not display the logViewer if we don't have reports.

src/UI/Components/LogViewer/LogViewer.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+    // Select second log
+    const option = screen.getByRole("option", { name: /Failed Log/i });
+    await userEvent.click(option);
+
+    // Check if new log content is displayed
+    const logViewer = screen.getByTestId("log-viewer");
+    expect(logViewer).toHaveTextContent("Error line 1");
+    expect(logViewer).toHaveTextContent("Error line 2");
+    expect(screen.getByText(/Duration: 200 ms/)).toBeInTheDocument();
```
matborowczyk, line 61: Duration is different value in the mock vs assertion, also here is 200 ms and above is 1s.

PR #6353, opened 2025-05-12
6347 - fail on warnings and errors in test logs

### LINKED ISSUE #6347: improve build checks
Improve the build check to be more strict

LukasStordeur:
# Description

Implement a tracker in Jest configuration for warnings and error messages. If any are present, they will be printed at the end of the tests. 
The failure is on the full test suite that ran. This is to avoid masking the output of individual test cases. 

closes #6347 

preview of what the failure would look like when msw catches an unhandled http call:

image


matborowczyk (approved):
That's really neat!
Only one minor thing, the formatting function could be extracted  as it's being reused there but it's not exported so I'm also fine with this staying as is.

PR #6360, opened 2025-05-13
Issue/6350 graphql notifications

### LINKED ISSUE #6350: GraphQL notification query

matborowczyk:
# Description

- Implementation of the graphQL query for the notifications 
- fetching notifications through graphQL compared to the REST query at the moment have limited filtering
- structure differs fairly, there are tools that automatically convert queries into type safe interfaces but I think that's beyond the scope and requires further disscussion

- briefly mentioned issue with queryName is due to the client extracting query Name and setting it as additional variable in the request, we can discuss it in the later stage whether we want to support it on the backend or not, especially as variables aren't supported either

closes #6350 

# Self Check:

- [ ] Changelog entry


src/Data/Managers/V2/Notification/GetNotificationsQL/useGetNotificationQL.tsx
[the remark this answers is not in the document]
matborowczyk: modified response

matborowczyk:
> Which is the example you used as a reference to go for graphql-request instead of the suggested implementation on React Query pages?

IT's basically useQuery implementation minus [GraphQL Code Generator](https://graphql-code-generator.com/)

src/Data/Managers/V2/Notification/GetNotificationsQL/useGetNotificationQL.tsx
```diff
@@ -0,0 +1,99 @@
...
+  };
+  errors: string[] | null;
+  extensions: Record<string, unknown>;
+}
+
+/**
+ * Return Signature of the useGetInstance React Query
+ */
+interface GetNotifications {
+  useContinuous: () => UseQueryResult<NotificationQueryResponse, CustomError>;
```
LukasStordeur, line 36: So we have now, a `NotificationQLResponse, NotificationQueryResponse, PartialNotification, Notification, NotificationGraphQLMock,` and lingering `NotificationQL` in the docstrings.

Is there a reason why we can't output to `Notification`? This would prevent the need for additional types/interfaces. This is for me one big mish mash, of old and new mixed together.

src/Slices/Notification/UI/Badge/Badge.test.tsx
```diff
@@ -93,4 +104,27 @@ describe("Badge", () => {
     expect(badge).toBeDisabled();
     expect(badge).not.toHaveAttribute("data-variant");
   });
+  test("Given Badge WHEN request fails THEN error is shown", async () => {
```
LukasStordeur, line 107: I expect the new formatting rules to insert a newline between functions here 🤔

src/Data/Managers/V2/Notification/GetNotificationsQL/useGetNotificationQL.tsx
```diff
@@ -0,0 +1,99 @@
...
+  extensions: Record<string, unknown>;
+}
+
+/**
+ * React Query hook for fetching notifications using GraphQL.
+ *
+ * @param {Props} props - The props object containing environment ID, cleared status, and order by.
+ * @returns GetNotifications A query result containing notifications data or an error.
+ */
+export const useGetNotificationQL = ({ envID, cleared, orderBy }: Props): GetNotifications => {
```
LukasStordeur, line 63: I wouldn't use QL as an identifier for the name of the query. This is the limited notification list we need for the app-shell. It would make sense to me to use a name that would reflect that instead. Especially considering you're already including extensions in the response, making it more than about Notifications.

Also be careful about plural/singular wording. `Notifications != Notification`

LukasStordeur (commented):
I find the typing not consistent. 

There's too much confusion possible between the query to fetch the notifications for the full notification page which has paging/filtering, and this one that is part of the app-shell (or whichever you want to call it). 

I also find it weird to add a new interface for the mocks.

src/Data/Managers/V2/Notification/GetNotificationsQL/useGetNotificationQL.tsx
matborowczyk: We can't output just Notification as that's not what we fetch, I think what we could do is to await the fetch in Query function, and if there is an errors then throw it manually, but I don't think that's the correct approach. The query. 

I simplified a bit, removed `NotificationGraphQLMock` `NotificationQLResponse` `NotificationQueryResponse` and `NotificationQL`

LukasStordeur: Since you're pulling in extensions too, (see other comment I left) I think calling this "notifications" is not exactly right.

matborowczyk: Okay, hmmmm, I lack the idea at the moment, because QL was about GraphQL which covered different response Extended notifications? THat doesn't sound good especially  as we are trimming them down

PR #6365, opened 2025-05-14
Replace File Fetcher

### LINKED ISSUE #6344: Update FileFetcher
FileFetcher is last dependency that use BaseApiHelper, we need to upgrade it to be able remove apiHelper and it's interfaces for good

matborowczyk:
# Description

- Updating File Fetcher and removing old implementation with all mocks
- removing unused components that I missed with env queries update
- remove RemoteData as it's no longer used
- remove ApiHelper with all mocks as it's no longer used 

Just in case This split second flicker is loading state that is almost immediately replaced with the data
https://github.com/user-attachments/assets/5625e083-6d45-4d43-9bf3-e9f1ef29bdef

closes #6344 

# Self Check:

- [ ] Changelog entry


src/Data/Managers/V2/Server/GetFile/useGetFile.ts
```diff
@@ -0,0 +1,41 @@
...
+      if (typeof data.content !== "undefined") {
+        // decode base64 string
+        try {
+          return window.atob(data.content);
+        } catch (_e) {
+          return data.content;
+        }
+      }
+
+      return "No data";
```
LukasStordeur, line 38: words dict, and "No data" is maybe not so user friendly

LukasStordeur (approved):

PR #6369, opened 2025-05-16
Issue/6367 server status fix

### LINKED ISSUE #6367: Server Status error fix

matborowczyk:
# Description

- Added loading state for server status fetch
- picked only relevant attributes from server status for feature Manager
- removed unused fetching hooks
- removed cleaning cache and moved environementId used for calls up to be able to trigger refetch of queries in a way that is designed by react query and that also improved UX for switching env - IF you don't want it at this PR it's all packed in one commit co I can revert it and cherry pick into separate branch.

- I didn't manage to get the error even when throttling internet to 3G, but I will run few times e2e to make sure that it won't appear. 

closes #6367 

# Self Check:

- [ ] Changelog entry


matborowczyk:
> * changelog file missing
> * And is this supposed to get rid of the Injector errors in the console too?

Yup, I will add that.
No, that will be handled in separate PR as discussed to not to mix tasks.

changelogs/unreleased/6367-server-status-fix.yml
```diff
@@ -0,0 +1,5 @@
+description: Fix server status throwing an error
+issue-nr: 6367
+change-type: patch
+destination-branches: [master, iso8]
+
```
LukasStordeur, line 5: Once in a while, it doesn't hurt to add a section to show the description :)

src/Data/Common/PrimaryFeatureManager.ts
```diff
@@ -41,12 +36,12 @@ export const PrimaryFeatureManager = (
     return commitHash;
   }
 
-  function get(): ServerStatus {
-    if (!serverStatus.data) {
+  function get(): Pick<ServerStatus, "features" | "extensions" | "version" | "edition" | "slices"> {
+    if (!features) {
       throw new Error("ServerStatus has not yet been set.");
```
LukasStordeur, line 41: Shouldn't we update the error message now?

src/Data/Managers/V2/helpers/useQueries.ts
```diff
@@ -14,16 +13,14 @@ import { useFetchHelpers } from "../helpers";
  *
  * @template Response - The type of the response data.
  */
-export const useGet = (options?: { message?: string }) => {
+export const useGet = (envId: string, options?: { message?: string }) => {
```
LukasStordeur, line 16: Strikes me that this one is using `envId`, while all others are using `env`

changelogs/unreleased/6367-server-status-fix.yml
matborowczyk: I opted not to add description to issues that did not went into user facing production, I don't know if that has sense, but why user should read about issue resolved that was was introducent in the iteration, from his perspercive there was no issue in the first place :) THat's why I recently don't add descriptions :)

LukasStordeur: let's rediscuss that on the planning meeting then

matborowczyk:
6/6 e2e runs without server status issue, 4 passed in a row, first two failed due to the 2.1.3 flake

LukasStordeur (approved):

PR #6370, opened 2025-05-16
adjust codeEditor based on feedback

### LINKED ISSUE #6366: Replace PF copy button with custom button
Patternfly's copy button on the CodeEditor only works on HTTPS, they use navigator.clipboard method, which is only supported on Secure origins. We can override the button and inject our own instead.

LukasStordeur:
# Description

- Replaced the copy button provided by default for the  codeEditor with a customControl that uses a copy method that is supported by both http and https.
- Update the default stage on the compile reports to be the failed one or the last one.
- Update the background color of all codeEditors to be the same as the app background.

closes #6366

### DIFF
--- a/src/UI/Components/LogViewer/LogViewer.tsx
+++ b/src/UI/Components/LogViewer/LogViewer.tsx
@@ -29,15 +29,24 @@ export interface LogViewerData {
 
 interface LogViewerProps {
   logs: LogViewerData[];
+  defaultSelected?: string;
 }
 
 /**
  * LogViewer component displays logs with a dropdown to select log parts and a download button.
  * Uses PatternFly's LogViewer for rendering logs.
- * @param logs Array of log objects, each with data, id, name, duration, and failed status.
+ *
+ * @prop {LogViewerData[]} logs - Array of log objects, each with data, id, name, duration, and failed status.
+ * @prop {string} defaultSelected - The id of the log to select by default.
+ *
+ * @note If the defaultSelected log is not found in the logs array, the last log will be selected.
+ *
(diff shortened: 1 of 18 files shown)


src/UI/Components/LogViewer/LogViewer.tsx
```diff
@@ -29,15 +29,24 @@ export interface LogViewerData {
...
+ *
+ * @note If the defaultSelected log is not found in the logs array, the last log will be selected.
+ *
+ * @returns {React.FC<LogViewerProps>} LogViewer component
  */
-export const LogViewerComponent: React.FC<LogViewerProps> = ({ logs }) => {
-  const [selectedLogId, setSelectedLogId] = useState(logs[0]?.id || "");
+export const LogViewerComponent: React.FC<LogViewerProps> = ({ logs, defaultSelected }) => {
+  const [selectedLogId, setSelectedLogId] = useState(
+    defaultSelected || logs[logs.length - 1]?.id || ""
```
matborowczyk, line 48: Hmmm, in views that have lists we often have separate view when the array is empty, `<EmptyView />` I think. Maybe that's something that could be thought about. not strong requirement per se, but I think it would be cleaner

matborowczyk (approved):
I approve as comment left is more something to take into consideration than hard req

src/UI/Components/LogViewer/LogViewer.tsx
LukasStordeur: you commented the same last time. We are dealing with an array, so we need the safe operator here to target a certain index. The check whether there are logs or not, is already there. We don't show the logViewer if there are no log reports.

PR #6381, opened 2025-05-19
Fix bad state for Envs

### LINKED ISSUE #6380: Fix bad state in env handler/modifier

matborowczyk:
# Description

- Fix bad state caused by setting envs through provider, now they are fetched through injector
- merged EnvModifier with EnvHandler, to simplify access to environments and also remove any re-renders when setting env between those two hooks
- removed need for mocking envs through jest.mock

https://github.com/user-attachments/assets/a5cc22e6-be0a-40c4-8288-17b1559d89f8

Screenshot 2025-05-20 at 14 14 33

closes #6380

# Self Check:

- [ ] Changelog entry


src/UI/Root/Root.test.tsx
```diff
@@ -66,6 +66,9 @@ describe("Root", () => {
 
     render(component);
 
+    // Wait for the env selector toggle to be visible and then check for accessibility violations after everything is rendered
```
LukasStordeur, line 69: what lead to this first check?

src/Test/Mock/MockEnvironmentHandler.ts
```diff
@@ -23,20 +23,40 @@ export function MockEnvironmentHandler(environment: FlatEnvironment): Environmen
...
+  function useIsServerCompileEnabled(): boolean {
+    return Boolean(environment.settings.server_compile);
+  }
+
+  function useIsExpertModeEnabled(): boolean {
+    return Boolean(environment.settings.enable_lsm_expert_mode);
+  }
+
+  function setAllEnvironments(): void {
+    return;
```
LukasStordeur, line 47: in case of void functions, you don't need to explicitly return. you can leave the brackets empty.
`  const myMethod = (): void => {}` or  ` function myMethod = (): void => {}` will have the same result.
This comment applies to the MockFeatureManager.ts too

src/UI/Dependency/Dummy/DummyFeatureManager.ts
```diff
@@ -1,4 +1,4 @@
-import { FeatureManager, JsonParserId, StatusLicense } from "@/Core";
+import { FeatureManager, JsonParserId, ServerStatus, StatusLicense } from "@/Core";
 
 export class DummyFeatureManager implements FeatureManager {
```
LukasStordeur, line 3: what's the difference between the MockFeatureManager, and DummyFeatureManager?

src/UI/Routing/RouteOutlet.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useContext } from "react";
 import { Outlet } from "react-router";
 import { DependencyContext } from "@/UI/Dependency";
-import { Initializer } from "../Root/Components/Initializer";
+import { Initializer } from "@/UI/Root/Components/Initializer/Initializer";
```
LukasStordeur, line 4: missing index file in the Initializer folder?

src/UI/Dependency/Dummy/DummyFeatureManager.ts
matborowczyk: Dummies in our app are  to indicate that given hook/provider is not set up and they will throw error, I assume it's a way to faster indicate that for test setup something is missing in the flow.

MockFeature Manager is, well, Mocked data

src/Test/Mock/MockEnvironmentHandler.ts
matborowczyk: That will trigger `Unexpected empty method`

src/Data/Common/PrimaryFeatureManager.ts
```diff
@@ -18,9 +18,12 @@ import { words } from "@/UI";
 export const PrimaryFeatureManager = (
   jsonParserId: JsonParserId = "Native",
   commitHash: string = "",
-  appVersion: string = "",
-  features?: Pick<ServerStatus, "features" | "extensions" | "version" | "edition" | "slices"> | null
+  appVersion: string = ""
 ): FeatureManager => {
+  const [features, setFeatures] = useState<
+    Pick<ServerStatus, "features" | "extensions" | "version" | "edition" | "slices"> | undefined
```
LukasStordeur, line 24: we need the undefined in the typing because we are setting it to undefined ourselves? If that's the case, I'd opt for null, as null shows it's intentionally set to null.

LukasStordeur (approved):

PR #6410, opened 2025-06-02
Fix e2e

matborowczyk:
# Description

- Standalone PR broken e2e on iso & oss
- according to the core "Setting deployed due to known good status" log won't appear in the logs ever again
- continuous queries in Initializer were causing refresh of the Instance Forms due to updating envs - those continuous queries were not necessary to updating halted status for environment nor other use-case

closes _Add ticket reference here_

# Self Check:

- [ ] Changelog entry


matborowczyk:
Screenshot 2025-06-03 at 07 50 30
 from the top it's oss and iso e2e

PR #6416, opened 2025-06-05
Implement fetching notifications for Notification Badge through GraphQL

### LINKED ISSUE #6350: GraphQL notification query

matborowczyk:
# Description

Implementation of the graphQL query for the notifications

fetching notifications through graphQL compared to the REST query at the moment have limited filtering

structure differs fairly, there are tools that automatically convert queries into type safe interfaces but I think that's beyond the scope and requires further disscussion

briefly mentioned issue with queryName is due to the client extracting query Name and setting it as additional variable in the request, we can discuss it in the later stage whether we want to support it on the backend or not, especially as variables aren't supported either

closes #6350 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/6350-graphql-notifications.yml
+++ b/changelogs/unreleased/6350-graphql-notifications.yml
@@ -1,0 +1,5 @@
+description: Implement GraphQL query for notifications
+issue-nr: 6350
+change-type: patch
+destination-branches: [master, iso8]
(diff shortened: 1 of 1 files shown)


LukasStordeur (approved):
Update the query with the QueryFactory after it got merged, and it's good to go

PR #6425, opened 2025-06-06
Implement GraphQL query for environment handler and expert mode

### LINKED ISSUE #6352: Change refetching in the Expert Modifier
currently when environment is updated we need to manually refetch  envsettings, to avoid inconsistencies and to make updates available the fastest. with graphQL we would have everything in the one query, so updates could trigger only graphQL call

matborowczyk:
# Description

- Implement GraphQL query for environment handler and expert mode
- to fully remove environment Settings from env handler we need two more settings added to environments in grahpQL `isProtected` and `isCompileEnabled` which would show settings `protected_environment` and `server_compile`
- there could be use for that query for envSelector if we would also include to environments projectName

I due to new interface I had to include some small modifications to the Env Selector, but as they were outside of the scope I did not do anything more I can create separate task for that or we could ask for that projectName to be included so we could revisit this component 

closes #6352 

# Self Check:

- [ ] Changelog entry


src/Data/Queries/Slices/Environment/GetEnvironmentPreview/useGetEnvironmentPreview.tsx
```diff
@@ -0,0 +1,90 @@
+import { useQuery, UseQueryResult } from "@tanstack/react-query";
+import { gql } from "graphql-request";
+import { Environment } from "@/Core/Domain";
+import { CustomError, useGraphQLRequest } from "@/Data/Queries";
+import { KeyFactory, SliceKeys } from "@/Data/Queries/Helpers/KeyFactory";
+
+/**
+ * Partial Environment type that represents the Environment type but with only the id, name, halted, and extra isExpertMode property.
```
LukasStordeur, line 8: Update docstrings :)

LukasStordeur (approved):

PR #6437, opened 2025-06-11
Change isCompiling REST  query for graphQL

### LINKED ISSUE #6436: add isCompiling graphQL query

matborowczyk:
# Description

- Change isCompiling REST query for graphQL query added to environmentHandler
- fix duplicate/edit form issue with updating environmentHandler
- revert refreshing on switching tabs - with fix on edit forms this will now can be safely enabled back to default
- add cache clearing after 5 seconds, i opted for 5 seconds instead of 10 as it doesn't affect active/stale views, it will clear cache for queries that will be inactive for 5 seconds

closes #6436 

https://github.com/user-attachments/assets/357cccf9-3e37-4cca-bd63-d2da25099895

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/UI/Dependency/EnvironmentHandler.ts
+++ b/src/UI/Dependency/EnvironmentHandler.ts
@@ -83,7 +83,7 @@ export function EnvironmentHandlerImpl(
 
     return env.halted;
   }
-  
+
   function useIsCompiling(): boolean {
     if (env === null) return false;
 


matborowczyk:
> I'm confused, wasn't `isCompiling` to be part of the other graphQL query with the `environmentPreview`?
> 
> If not, okay, why did you call it CompilingPreview, while there's only one property in there. Will there be a Compiling interface? What would be the full interface then?

it could be but as we have environment as part of the dependency context it's hard to do continuous query without that form refreshing issue, because we are updating environments from the initialiser. I'll take one more look on that

matborowczyk:
> > it could be but as we have environment as part of the dependency context it's hard to do continuous query without that form refreshing issue, because we are updating environments from the initialiser. I'll take one more look on that
> 
> If that's the case, what's the point of using graphQL for this instead of the regular endpoint?

having definitely clearer view at what we are fetching instead to asserting against query status code, but I get your point. I will check continuous query through Dependency Context and extracting envHandler to be standalone

matborowczyk:
https://github.com/user-attachments/assets/93bb4251-e535-4008-aaaf-58438fdaa9ad

LukasStordeur (approved):

PR #6445, opened 2025-06-17
Add functionality to change password for users

### LINKED ISSUE #6439: Add ability to change the password of users
The user view currently does not allow to change a password while there is an API endpoint for it.

https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/protocol/methods_v2.py#L1554

matborowczyk:
# Description

- Add functionality to change password for users

https://github.com/user-attachments/assets/2eb8d87e-c755-4e6e-b34c-58c08c9f025c

closes #6439 

# Self Check:

- [ ] Changelog entry


LukasStordeur (changes requested):
I'm missing an e2e test for the password change :)

LukasStordeur:
ALso just noticed you use the label "Actions" in the headers, but we never use a label for the last column when it contains action buttons.

LukasStordeur (approved):
Perhaps just use the smaller size buttons for in the table? Might make it look more balanced

matborowczyk:
Screenshot 2025-06-17 at 16 08 10

PR #6460, opened 2025-06-19
Create functionality to manage user roles

### LINKED ISSUE #6440: Add ability to assign roles to a user
The user management should support adding roles to users. There are two API endpoints:

- To list all the roles
- To assign the roles

https://inmanta.slack.com/archives/C045Y42N49W/p1749212135942279

matborowczyk:
# Description

- On the video the empty view is not aligned correctly, it is fixed on the screenshot, I didn't want to record it once again :) 
- error messages for actions, removal and addition of roles are showed through alerts

Screenshot 2025-06-19 at 17 28 32

Screenshot 2025-06-19 at 17 30 02

https://github.com/user-attachments/assets/6fdbc2cb-c14e-4895-a48b-0abf482cec13

closes #6440 

# Self Check:

- [ ] Changelog entry


src/Slices/UserManagement/UI/Components/RolesMainColumn/RolesMainColumn.tsx
```diff
@@ -0,0 +1,52 @@
...
+   */
+  const onRetry = (): void => {
+    roles.refetch();
+    setAlert("", AlertVariant.danger, ""); //
+  };
+
+  if (roles.isSuccess) {
+    return (
+      <Td dataLabel={words("userManagement.roles")} aria-label="roles-success">
+        {[...new Set(roles.data.map((role) => role.name))].join(", ")}
```
LukasStordeur, line 32: This should be a link/button that collapses/expands the row if clicked

src/Slices/UserManagement/UI/Components/RolesMainColumn/RolesMainColumn.tsx
```diff
@@ -0,0 +1,52 @@
...
+  roles: UseQueryResult<UserRoleInfo[], Error>;
+  setAlert: (title: string, variant: AlertVariant, message: string) => void;
+}
+
+/**
+ * A functional component that renders the top column of the role table.
+ * @param {UseQueryResult<UserRoleInfo[], Error>} roles - The roles of the user.
+ * @returns {React.FC<Props>} The rendered role top column.
+ */
+export const RolesMainColumn = ({ roles, setAlert }: Props) => {
```
LukasStordeur, line 18: What's a main column ?

matborowczyk: renamed to RolesToggleCell

LukasStordeur:
It's almost sad to see a new endpoint, where you can't add/delete multiple roles at once. Especially if this is just an array of strings. Disabling the select while the request is pending, and you have a few roles to assign, is a drag.

LukasStordeur:
Could you also show the UI now that it is a button/link that can open/close the rows?

src/Slices/UserManagement/UI/Page.test.tsx
```diff
@@ -153,21 +330,90 @@ describe("UserManagementPage", () => {
...
+    expect(successView).toBeInTheDocument();
+
     await act(async () => {
       const results = await axe(document.body);
 
       expect(results).toHaveNoViolations();
     });
+  });
+
+  it("should display error alert view when users roles API call fails", async () => {
```
LukasStordeur, line 348:
```suggestion
  it("should display error alert view when updating roles fails", async () => {
```

LukasStordeur (commented):
In the error handling, do we uncheck the selected value that triggered the error response and assert this in the tests?

matborowczyk:
> In the error handling, do we uncheck the selected value that triggered the error response and assert this in the tests?

We don't check it until it's success, or to better word it, we get selected roles from the endpoint

src/Slices/UserManagement/UI/Page.test.tsx
matborowczyk: this suggestion is not adequate to what that test checks

LukasStordeur: the original isn't better imo, it makes me think you need a call both for the user, and the roles to populate the roles column 🤔But based on the content of the test, it might be the case, and if that is, then why isn't it present in the getUser query?

matborowczyk:
added e2e test scenarios, I will provide screenshot from the pipeline as soon as it will end :) 
Updated the label for no roles to get the span instead of div, and to align the look, what do you think?

Screenshot 2025-06-30 at 11 01 59

PR #6474, opened 2025-07-02
ISO 7 - Dryrun change : update e2e

LukasStordeur:
# Description

This is the iso 7 PR for https://github.com/inmanta/web-console/pull/6471

From now on, the LifecycleTransfer resource never reports a change in the dryrun. Thus, the lines related to asserting the next_version would be bumped, can be removed.

### DIFF
--- a/cypress/e2e/scenario-4-desired-state.cy.js
+++ b/cypress/e2e/scenario-4-desired-state.cy.js
@@ -485,16 +485,6 @@ describe("Scenario 4 Desired State", () => {
       }
     });
 
-    // await the end of the dry-run and expect to find two rows with expandable content.
-    cy.get(".pf-v6-c-card__expandable-content", { timeout: 20000 }).should(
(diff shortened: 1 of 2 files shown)


cypress/e2e/scenario-4-desired-state.cy.js
```diff
@@ -554,18 +518,12 @@ describe("Scenario 4 Desired State", () => {
       }
     });
     // expect the view to still contain the diff of the last dry-run comparison
-    cy.get(".pf-v5-c-card__expandable-content", { timeout: 20000 }).should(
+    cy.get(".pf-v6-c-card__expandable-content", { timeout: 20000 }).should(
```
bartv, line 521: Is this correct for iso7?

LukasStordeur: Right, I missed that and was confused for a moment why the test wasn't passing 😅

PR #6491, opened 2025-08-05
Align build date format iso8

LukasStordeur:
# Description

Align the buildDate format to ISO format in the version.json file.

### DIFF
--- a/changelogs/unreleased/align-buildDate.yml
+++ b/changelogs/unreleased/align-buildDate.yml
@@ -1,0 +1,3 @@
+description: Fix the date format of the build date.
+change-type: patch
+destination-branches: [iso8]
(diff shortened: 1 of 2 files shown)


cypress/e2e/scenario-6-resources.cy.js
jptrindade, line 1: Why were the messages number decreased? Is it due to the changes in the Jenkinsfile?

LukasStordeur: It's due to a change in core, we don't get the same amount of logs as we did before. I might rethink the assertions later, but for now, it's blocking the pipeline :D

jptrindade (approved):

PR #6533, opened 2025-08-26
Prevent stalling when environment is missing or invalid.

LukasStordeur:
# Description

## Issue
The app crashed whenever the url contained an invalid environment id, or if the id was missing anywhere in the query-hooks.

The error was being thrown during React's rendering phase from within a custom hook (environmentHandler.useId() or environmentHandler.useName()). While Error Boundaries should catch rendering errors, there are specific cases where they might not:
- Context hook errors - Errors thrown from within context hooks during the initial render might bypass error boundaries in certain scenarios
- Hook execution timing - The error occurs during hook execution which might happen before the Error Boundary is fully set up

## Solution
I removed the throws and handled the case where the id was required and would block the UI. The app now redirects to the home page whenever it cannot find the environment id. 
I also had the behavior setting the status-down, a missing environment would result in showing the icon red, but the statuspage doesn't provide information about missing environments, so it wasn't relevant.

https://github.com/user-attachments/assets/f9b8bb3e-71b6-4906-b4fc-57a6ba145708

### DIFF
--- a/changelogs/unreleased/prevent-stalling-env-missing.yml
+++ b/changelogs/unreleased/prevent-stalling-env-missing.yml
@@ -1,0 +1,6 @@
+change-type: patch
+description: Prevent stalling when environment is missing or invalid.
+destination-branches:
+- master
+- iso8
+sections: {}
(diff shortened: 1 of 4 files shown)


Hugo-Inmanta (approved):

changelogs/unreleased/prevent-stalling-env-missing.yml
jptrindade, line 1: iso7 was not affected?

PR #6542, opened 2025-08-29
6512 composer improvements

LukasStordeur:
# Resolved
- Improved some namings here and there
- Added tests for methods I extracted
- Reused the ErrorView Component instead of a local declared one.
- Items that were already on the canvas on initial load are now correctly disabled

# In next iteration
I'll restructure the ServiceEntityBlock class to remove the wrapping functions that instantiate each a slightly different type of Block/Instance.


src/UI/Components/Diagram/actions/editMode.ts
```diff
@@ -169,51 +169,77 @@ export function appendInstance(
...
+  paper,
+  instanceEntityBlock,
+  serviceModel,
+  instance,
+  isBlockedFromEditing = false,
+}: AddEmbeddedEntitiesParams): ServiceEntityBlock[] {
   const { embedded_entities } = serviceModel;
-  //iterate through embedded entities to create and connect them
-  //we are basing iteration on service Model, if there is no value in the instance and if the value has modifier set to "r", skip that entity - "r" entities are read-only, they can't be edited and can be in multiple places which would result with enormous tree of cells in the graph and the canvas which would discourage user from using the Instance Composer
+  const attributes = instance.candidate_attributes || instance.active_attributes || undefined;
```
jptrindade, line 207: Isn't  `|| undefined` redundant?

src/UI/Components/Diagram/helpers/relations.ts
```diff
@@ -1,47 +1,46 @@
...
-    return extractRelation(instance.active_attributes);
-  } else {
+  if (!attributes) {
     return [];
   }
+
+  return relationKeys
+    .map((key) => attributes[key])
+    .filter(Boolean)
+    .map(String);
```
jptrindade, line 36: what does this actually do? Specifically the filter and the map

src/UI/Components/Diagram/helpers/relations.ts
```diff
@@ -1,47 +1,46 @@
...
 };
 
 /**
  * Finds the inter-service relations entity types for the given service model or embedded entity.
  *
+ * TODO: This recursive function should be adjusted to work with levels, and also cover x-level of nested inter-service relations.
+ *
```
jptrindade, line 43: Is there a ticket for this? If so can you add the ticket number in the comment?

src/UI/Components/Diagram/stencil/LeftSidebar.ts
```diff
@@ -103,6 +83,47 @@ export class StencilSidebar {
...
+  public toggleTab = (clickedElement: EventTarget) => {
+    if (clickedElement instanceof HTMLElement) {
+      // The clickedElement can also be the entire tabContainer, which we don't want to react on.
+      clickedElement.innerText === "Inventory" &&
+        this.toggleTabVisibility(this.inventoryTab, this.instanceTab, "inventory-tab");
+
+      clickedElement.innerText === "New" &&
+        this.toggleTabVisibility(this.instanceTab, this.inventoryTab, "new-tab");
+    }
+  };
```
jptrindade, line 125: Not sure how it is done in the frontend, but couldn't we use an enum to store the possible configurations of this `clickedElement.innerText` instead of hard coding `Inventory/New`?

src/UI/Components/Diagram/helpers/relations.ts
Hugo-Inmanta: I could be wrong, but if this behaves the same as it does in python, this takes a list of keys (`relationKeys`) as input and for each of these key, we look at the associated value in a map (`attributes[key]`), if it is truthy, we cast it to string and add it to the output list, if it is falsy, we drop it.

- filter will apply a function on each element and only keep those that pass the test
- map will apply a given function on each element and replace them with the result

this looks like a bug compared to previous implementation. I think this now only keeps the truthy attributes (and drops all falsy ones e.g. "false" ) whereas before we would drop only the undefined ones.

src/UI/Components/Diagram/stencil/LeftSidebar.ts
LukasStordeur: Yes, we do use them too, but I admit I didn't yet focus on that part of the refactoring :D 
I'll update

src/UI/Components/Diagram/helpers/relations.ts
LukasStordeur: Correct. 
But the bug was the previous implementation, where empty values, or null values would be retained, which wouldn't be valid ids.
But looking for implementations, I realize the method is not used anymore 😅 Should have started with that before making a change to it...

Issue #6562, opened 2025-09-09
Add Discovered Details page


LukasStordeur:
Add a details page for the discovered resources to match the behaviour of the resource page.

Issue #6565, opened 2025-09-10
Add filtering to discovered resources page


LukasStordeur:
Add filtering to discovered resources page.
Add sorting back.
Use the agent/type/value from api instead of splitting it from the full resource id on the Frontend.

PR #6566, opened 2025-09-10
6562 Discovered Resource Details page

LukasStordeur:
# Description

Add a page to display the details of a discovered resource, matching the behaviour on the Resource page. 
The links forwarding to the Managed/discovery resource are now in a kebab menu at the end of the row. When no links ara available, it is disabled to keep the space it uses in the row, preventing the "show details" button to not be aligned properly in rows that don't have links.

**Note:** 
> - Sorting is temporarily disabled, until the API supports filtering by agent/value/type.
> - E2E tests cannot be implemented, the local-setup doesn't have discovered resources.
> - The name of the branch is not accurate, I was innitially supposed to start with the filtering, but had to switch to working on the page instead.

closes: 
- #6562 

https://github.com/user-attachments/assets/e944cc56-09c0-48c1-adbe-d5ee32435481

### DIFF
--- a/src/UI/words.tsx
+++ b/src/UI/words.tsx
@@ -576,8 +576,6 @@ const dict = {
   "discovered.column.managed_resource": "Managed resource",
   "discovered.column.discovery_resource": "Discovery resource",
   "discovered.column.show_details": "Show Details",
-  "discovered.column.disabled_actions":
-    "This resource has no managed or discovery resource associated.",
   "discovered_resources.title": "Discovered Resources",
   "discovered_resources.values": "values",
   "discovered_resources.show_resource.managed": "Show managed resource",


src/Data/Parsers/ResourceIdSplitter.test.ts
```diff
@@ -0,0 +1,265 @@
...
+    it("should parse resource ID with incomplete brackets (missing closing bracket)", () => {
+      const resourceId = "resource::type[agent,key=value";
+      const result = splitResourceId(resourceId);
+
+      expect(result).toEqual({
+        type: "resource::type",
+        agent: "agent",
+        value: "value",
+      });
+    });
```
Hugo-Inmanta, line 129: EDIT: ignore this comment 
> 
> Is there a reason why we want this behaviour ? I don't have a good feel for how realistic it is/what the flow is like, but I feel this could lead to errors e.g.:
> - full resource id is `resource::type[agent,key=value_untruncated]`
> - for some reason it gets truncated to `resource::type[agent,key=value`
> - `splitResourceId` will silently parse `value=value` instead of `value=value_untruncated`
>

Hugo-Inmanta: Ignore above, just realized this parsing was a temporary work around https://github.com/inmanta/web-console/issues/6565

src/UI/words.tsx
```diff
@@ -568,8 +570,14 @@ const dict = {
...
   /** Discovered Resources related text */
   "discovered.column.resource_id": "Resource Id",
+  "discovered.column.agent": "Agent",
+  "discovered.column.value": "Value",
+  "discovered.column.type": "Type",
   "discovered.column.managed_resource": "Managed resource",
   "discovered.column.discovery_resource": "Discovery resource",
+  "discovered.column.show_details": "Show Details",
+  "discovered.column.disabled_actions":
+    "This resource has no managed or discovery resource associated.",
```
Hugo-Inmanta, line 580: This one is not used afaics ?

LukasStordeur: Good catch, I had it initially, but wasn't needed since the Kebab menu from patternfly is not able to display disabled tooltips

Issue #6502, opened 2025-08-12
Port exisiting filters to Side pannel widget


LukasStordeur:
The existing filters can already be ported to the new sidepannel widget. They will get replaced later with the new graphQL endpoint capacities.

PR #6615, opened 2025-11-14
6502 - Move exisiting Filters into a Collapsible Sidepanel

LukasStordeur:
# Description

This is the first stage to move towards the new state filters.

- The exisiting filters are now moved into a sidepanel, with a tab for the different sorts of filters.
- The sidepanel is resizable
- The requires column has been removed.

https://github.com/user-attachments/assets/c62e1ebf-0473-4974-8f9f-b2b94b8e770c

close: 
- #6502

### DIFF
--- a/src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/utils.ts
+++ b/src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/utils.ts
@@ -4,7 +4,7 @@ import { toggleValueInList } from "@/Core";
  * invertFilter.
  *
  * Returns the opposite representation of a status filter entry by toggling the leading `!`.
- * 
+ *
  * @param {string} selection - The current status filter toggle value.
  * @returns {string} The inverted selection.
  */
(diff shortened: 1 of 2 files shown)


Hugo-Inmanta (approved):
As usual, I did a best effort review, but I must admit most of this is dark magic to me :D 

One minor comment regarding the video: I don't know whether it is even possible or how easy it would be to implement but I find it slightly off-putting that the whole page disappears when applying a new filter. Would it be possible e.g. to keep displaying the previous results in the background with a 'loading' animation on top of it ?

LukasStordeur:
> As usual, I did a best effort review, but I must admit most of this is dark magic to me :D
> 
> One minor comment regarding the video: I don't know whether it is even possible or how easy it would be to implement but I find it slightly off-putting that the whole page disappears when applying a new filter. Would it be possible e.g. to keep displaying the previous results in the background with a 'loading' animation on top of it ?

It's something I'd like to fix aswell, but currently, the entire page reloads on updating the url, which contains the keys for the filters. I'll tackle it when updating the queries. It's also a behaviour that is in the current version of the page.

PR #6676, opened 2026-02-12
Resolve security alerts

LukasStordeur:
# Description

There were a few outstanding security alerts. 
And the next Lodash dependency update was causing issues on jenkins. I replaced the methods with local reusable ones + tests.


Hugo-Inmanta (changes requested):

src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx
```diff
@@ -606,6 +605,37 @@ const DictListFieldInput: React.FC<DictListProps> = ({
   );
 };
 
+function getValue<T>(obj: unknown, path: string | (string | number)[], defaultValue: T): T;
+function getValue<T = unknown>(
```
jptrindade: I think this should have a more specific name

Issue #6655, opened 2026-01-16
Add support to run compile from a clean slate


arnaudsjs:
[This PR](https://github.com/inmanta/inmanta-core/pull/9903) adds a `reinstall` parameter to the `notify_change` endpoints. It allows requesting a new compile that first performs a clean checkout of the project and creates a new compiler venv. 

**Task:**
Implement an additional option "Cleanup and Recompile" in the Recompile button dropdown. 
Implement a toaster message when the action is performed in similar fashion as for the Recomile and Update project & recompile.
This feature is available for both OSS and ISO versions. 
ISO9 & ISO10

<img width="425" height="136" alt="Image" src="https://github.com/user-attachments/assets/42741564-deaf-43ef-936a-2f516132f723" />

PR #6730, opened 2026-03-06
feat: new button is added where it is possible to run a cleanup and recompile

AronH99:
# Description

Added support to run compile from a clean slate, it is now possible to run a cleanup and recompile for the compile reports.
If u navigate to the compile reports a new button is added under the recompile dropdown which will execute this functionality.

https://github.com/inmanta/web-console/issues/6655

### DIFF
--- a/cypress/e2e/scenario-5-compile-reports.cy.js
+++ b/cypress/e2e/scenario-5-compile-reports.cy.js
@@ -128,30 +128,28 @@ describe("5 Compile reports", () => {
     cy.get('[aria-label="Toggle"]').click();
 
     // Get the current row count before triggering the compile
+    cy.get("tbody tr").its("length").as("initialRowCount");
+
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/6655-run-compile-from-clean-slate.yml
```diff
@@ -0,0 +1,6 @@
+description: Added support to run compile from a clean slate, it is now possible to run a cleanup and recompile
+issue-nr: 6655
+change-type: feat
```
LukasStordeur: patch

LukasStordeur (changes requested):
Can you move "Triggered by the web-console" text to the words.ts file, and add some nuance to it depending on which compile was triggered. (That means some tests will be impacted aswell.)

Issue #6711, opened 2026-02-26
[REQUEST] - GraphiQL page


LukasStordeur:
### Summary

Implement a page containing the GraphiQL editor. This would allow us to have a more interactive documentation of our graphql endpoint.

PR #6751, opened 2026-03-17
6711 graphiql page

LukasStordeur:
# Description

A GraphiQL editor has been implemented for testing GraphQL queries against the Console's GraphQL API. 
The editor is accessible via the "Developer Tools" icon-menu in the header, 
and provides features such as syntax highlighting, autocompletion, and query history.

I also opted for a dropdown menu in the header to avoid adding yet another icon that would look similar to others. We might consider adding the service-catalog rest api link there too when an environment is active/selected.

> Note: GraphiQL monaco editor ships with unresolved .map imports. These do not break the dev/builds, but were crowding the terminal. These warnings are not something we can fix. I supressed these specific warnings. 

closes:
- #6711 

https://github.com/user-attachments/assets/cdcffe8c-8563-4b37-a707-3f429727e41e

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -179,9 +179,7 @@ const stripBrokenSourcemapsPlugin = () => ({
   transform(code: string, id: string) {
     if (
       id.includes("node_modules") &&
-      (id.includes("marked") ||
-        id.includes("graphiql") ||
-        id.includes("@graphiql")) &&
+      (id.includes("marked") || id.includes("graphiql") || id.includes("@graphiql")) &&
       code.includes("sourceMappingURL")
     ) {
       return {
--- a/src/Data/Queries/Slices/GraphiQL/PostGraphQL/usePostGraphQL.ts
+++ b/src/Data/Queries/Slices/GraphiQL/PostGraphQL/usePostGraphQL.ts
@@ -23,9 +23,9 @@ export const usePostGraphQL = (env?: string): UseMutationResult<unknown, Error, 
     () =>
       new PrimaryBaseUrlManager(
         globalThis.location.origin,
-        globalThis.location.pathname,
+        globalThis.location.pathname
       ).getBaseUrl(import.meta.env.VITE_API_BASEURL),
-    [],
+    []
   );
 
   return useMutation({


src/Slices/GraphiQL/UI/Page.tsx
```diff
@@ -0,0 +1,130 @@
...
+    [mutateAsync]
+  );
+
+  // GraphiQL's internal introspect() is a no-op when a schema prop is provided
+  // (shouldIntrospect is set to false). Intercept the sidebar button click via
+  // event delegation and invalidate the React Query cache instead.
+  const handleRefetchSchema = useCallback(
+    (e: React.MouseEvent) => {
+      const target = e.target as HTMLElement;
+      if (target.closest('[aria-label^="Re-fetch GraphQL schema"]')) {
```
LukasStordeur: This event delegation relies on an internal aria-label of the GraphiQL library (`"Re-fetch GraphQL schema"`). If a future GraphiQL upgrade renames or rewrites the button, this condition will silently evaluate to `false` and the refetch button will appear to work (no error) but do nothing.

Two suggestions:
1. Add a comment pinning this to the exact `graphiql` version in `package.json` so it's immediately visible as "version-sensitive".
2. Consider adding a small integration smoke-test (or at least a comment in `Page.test.tsx`) that verifies the re-fetch path, so a breakage gets caught rather than going unnoticed.

vite.config.ts
LukasStordeur: The `monaco-graphql` alias hard-codes a deeply-nested `node_modules` path:

```ts
replacement: resolve(__dirname, "./node_modules/@graphiql/react/node_modules/monaco-graphql"),
```

This path exists today because Yarn didn't hoist `monaco-graphql` to the root. If the lockfile is ever updated and Yarn does hoist it, the alias will point to a path that no longer exists and the build will silently fall back to the hoisted copy — reintroducing the duplicate Monaco language registrations this is meant to prevent.

A few suggestions:
- Add a comment that explains the exact symptom this guards against ("duplicate `graphql` language registered in Monaco → console error + broken autocomplete") so future developers understand *why* the alias must stay.
- Add a `require.resolve` guard or a startup assertion that confirms the target path exists, so a lockfile bump fails loudly rather than silently.

PR #6784, opened 2026-03-27
Feat/6690 graphql resources

LukasStordeur:
# Description

This part handles the replacement of the REST-api call with the gql. 
I had to bridge some functionality to keep everything working.


src/Data/Queries/Slices/Agents/GetAgents/getUrl.ts
```diff
@@ -1,33 +1,54 @@
...
-              process_name: filter.process_name,
-              status: filter.status,
-            },
+            filter: { name: filter.name, process_name: filter.process_name, status: filter.status },
           },
           { allowDots: true, arrayFormat: "repeat" }
         )}`
       : "";
   const sortParam = sort ? `&sort=${Sort.serialize(sort)}` : "";
+  const cursorParam = cursor
```
LukasStordeur: Nested ternaries are in general not such a good practice towards readability. Just an FYI for the future :D you don't perse need to change it now.

src/Data/Queries/Slices/Resource/GetResources/useGetResources.ts
```diff
@@ -37,42 +48,135 @@ interface GetResources {
...
-export interface GetResourcesParams {
+interface GetResourcesParams {
   pageSize: PageSize.PageSize;
-  filter: Filter;
+  filter: Resource.Filter;
   sort?: Sort.Type<Resource.SortKey>;
   currentPage: CurrentPage;
 }
 
+const GET_RESOURCES_QUERY = gql`
```
LukasStordeur, line 58: Suggestion for future, if you like the idea.
Since those gql queries can be long, we can maybe keep them in a dedicated file, one for each hook.

LukasStordeur (commented):
If the changelog is updated, this will be good to go. Make sure to pull master first in here too

src/Data/Queries/Slices/Resource/GetResources/useGetResources.ts
AronH99: yes we can do that for sure

AronH99: It also depends on the use case because graphql allows you to query any fields, few or many. 
If u want to reuse certain queries but you don't need all of the fields because u want to query less information then managing this might prove difficult.
We used to work with apollo client and the caching system on that one made this into a real issue/nightmare.
But i believe it might be easier with tanstack to manage. 
Anyway i think bundling them still gives a good overview of the needs of the console.

AronH99 (approved):

PR #6801, opened 2026-04-03
Fix Dependabot authentication for JointJS npm registry

bartv:
## Summary
- Add `https://npm.jointjs.com` to `npmRegistries` in `.yarnrc.yml` so Dependabot can properly inject authentication tokens when checking for dependency updates.
- Dependabot's Yarn Berry integration matches and replaces tokens via `npmRegistries`, not `npmScopes`. The existing `npmScopes.joint` entry alone was insufficient for Dependabot, even though it works for regular Yarn operations.

## Test plan
- [ ] Verify Dependabot can successfully check for `@joint/*` package updates after merge
- [ ] Verify `yarn install` still works locally and in Jenkins CI

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -8,6 +8,9 @@ npmRegistries:
   "https://npm.pkg.github.com":
     npmAlwaysAuth: true
     npmAuthToken: "${IRT_GIT_TOKEN:-}"
+  "https://npm.jointjs.com":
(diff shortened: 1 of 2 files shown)


bartv:
Closing — further investigation shows that adding the registry to `npmRegistries` won't help since Dependabot replaces all env vars with dummy credentials regardless. See the Slack thread for details.

Issue #6504, opened 2025-08-12
Resource Filters: Update Table content


LukasStordeur:
- Remove collapsible sections
- Add a column for the compound states
- Update the "no resource found" text with "The given combination of filters didn't match any exisiting resources, please edit your filter values." 
- Color the filter counter red when no resources match given filters.

PR #6803, opened 2026-04-10
fix: updated table view of resources and more

AronH99:
# Description

So i did alot of things:

1. Normally all of the requirements from the ticket should be done (see video).

2. I adjusted the types to fit the needs i came across but i am not 100% sure if the naming is very clear, it's kind of hard to come up with good naming for these kind of things.
Unless i made a mistake this seemed to me the most logical conclusion.

3. Separated the helpers for the query to a different file and did adjustments on them for the filtering part, now it is possible to combine filtering of all compound states but only 1 status per state if that makes sense.

4. Did some cleanup along the way and corrected a mistake i made in my previous pr.

5. Adjusted the ux like u said to work with opening/closing the filter and menu so that only one of them would be open at all times.

6. Struggled a lot with Patternfly and it's anti customization but i tried as much as possible to follow their core design without deviation too much from the design u made, let me know what u think or would like to see changed but it seems like Patternfly really is hard to customise which is a curse and a blessing at the same time i guess.
I feel like i did the best i could because i encountered a lot of niche issues which was kind of frustrating to work around with to be honest.
Now the only thing i would love to improve a little is whenever the DeployingDot appears and disappears then the table cells shift a little in width which is annoying.
I tried a lot of things and a lot of different implementations but couldn't seem to make this better.
It's not the end of the world but still not 100% fine-tuned I guess.
Would propose maybe to keep the dot permanently added but then only blinking whenever it is deploying, that way we don't have a shifting layout sometimes and it should still be clear what's deploying and what isn't?
Had also some issues with the popover component from Patternfly which is why I currently use the onClick implementation i got going right now (see video) which should make clear that people can click on it to display more information.

7. The totalCount in the topleft corner of the page, i left a comment somewhere with your name on it, should it reflect the orphan resources as well or just the non orphan ones? So total the api returns or what the summary returns?
Because i feel like that can get confusing in the UI honestly.
I return the total of the summary for now.

8. Talking about orphans, I asked around in the standup, the core team and sollutions team, they recommended me to the current solution I have concerning the implementation of the orphaned resources.
They get grey/blackish icons now and limited data to show but still have sameish cell as the non orphaned resources.
If you or Bart had anything else in mind I am happy to learn more about it then :)

This kind of sums it up, next week I will start working on the filtering part of the resources to try and finish that entirely.

https://github.com/inmanta/web-console/issues/6504

https://github.com/user-attachments/assets/2b7c7984-e5a8-4c23-bc0f-8f72e00af979

Screenshot 2026-04-13 at 11 16 23


src/Core/Domain/Resource/Resource.ts
```diff
@@ -59,6 +61,14 @@ export interface CompoundState {
   lastHandlerRun: Record<LastHandlerRunState, number>;
 }
 
+/** Possible uppercased values grouped by state. */
+export type CompoundStateUpper = {
```
LukasStordeur: This looks like a red flag to me, and shows we have a structural issue with our approach. 

1. Why do we need this? Is it truly for typing validation, or is it a runtime display check to know what label we need to show? Is it because the innitial typing above didn't cover our needs? In which case, is this the right way? Or should we adjust the previous typing to reflect what we need?
2. I don't know for you, but this sort of Typing is overly complicated, and hard to read. 
3. What is the difference between this one, and the one below naming wise? Hard to tell unless you dissect the typing itself.

src/Core/Domain/Resource/Resource.ts
```diff
@@ -78,12 +88,18 @@ export interface Row {
   type: string;
   agent: string;
   value: string;
-  numberOfDependencies: ParsedNumber;
-  deployState: Status;
+  status: {
+    isDeploying: boolean;
+    isOrphan: boolean;
+    lastHandlerRunAt?: string;
+  } & CompoundStateUpper;
```
LukasStordeur: I was under the impression that the old endpoints wouldn't return the Compound properties. Why are we adding it to the deprecated Status enum? Is it to bridge old and new? Is this temporary?

src/Core/Domain/Resource/Resource.ts
```diff
@@ -78,12 +88,18 @@ export interface Row {
...
+  requiresLength: number;
   id: string;
 }
 
-export type RowFromVersion = Omit<Row, "deployState">;
+export type RowFromVersion = Omit<Row, "status" | "requiresLength"> & {
```
LukasStordeur: I'm not exactly fond of the & union operator in typing. It's more often a code smell than anything else, where you're trying to glue up some pieces that weren't properly designed. It makes things harder to read too.

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
```diff
@@ -0,0 +1,92 @@
...
+} = {
+  blocked: Resource.BlockedState,
+  compliance: Resource.ComplianceState,
+  lastHandlerRun: Resource.LastHandlerRunState,
+};
+
+/** Mapping of CompoundStateType values to their corresponding fields in CompoundState.
+ * This is used to determine which field to update when a status filter is toggled.
+ */
+export const STATE_FIELD_MAP: Record<Resource.CompoundStateType, keyof Resource.CompoundState> =
```
LukasStordeur: This makes me dizzy x) 
First of all, you're doing super duper complex logic to map, loop, iterate, to I have no idea what at first glance. To then cast it to tell Typescript you don't need a chekc here, because you know what you're doing. What's the point of all that typing you're doing, if you're not going to use it?

src/Slices/Resource/UI/ResourcesPage/Components/ResourceTableControls.tsx
```diff
@@ -30,41 +31,59 @@ interface Props {
...
+        {summaryWidget}
+        <ToolbarItem variant="pagination">{paginationWidget}</ToolbarItem>
+        <ToolbarItem>
+          <Button
+            onClick={onToggleFilters}
+            variant="link"
+            icon={
+              <FilterIcon
+                {...(noResourcesFound && {
+                  style: { color: "var(--pf-t--color--red--60)" },
```
LukasStordeur: for standard icon colors, you can do this to not override the css tokens Patternfly is using. 
```TS
          <Icon status="danger">
            <FilterIcon />
          </Icon>
```
But in this case, I believe you'd want the whole button to be in danger state perhaps? Wouldn't that reduce the complexity?

src/Slices/Resource/UI/ResourcesPage/Page.tsx
```diff
@@ -59,7 +66,10 @@ export const Page: React.FC = () => {
     }, 0);
   }, [filterWithDefaults]);
 
-  const onCloseFilterWidget = useCallback(() => setIsDrawerExpanded(false), []);
+  const onCloseFilterWidget = useCallback(() => {
+    if (!isSidebarOpen) onSidebarToggle();
```
LukasStordeur: blocked statement please.

And I remember I mentionned this could wait for a later stage but you might have forgotten :P (mostly because it's not that straightforward and has more to it than it seems.)
From the info you provided in the PR description, it's not exactly what we want. It shouldn't be "at all times" or toggling back an forth whenever the filter widget closes and opens. I'd leave it out for now

src/Slices/Resource/UI/ResourcesPage/Page.tsx
```diff
@@ -100,8 +110,8 @@ export const Page: React.FC = () => {
               {words("inventory.tabs.resources")}
             </Content>
             <Label icon={<CubesIcon />} variant="outline" color="blue">
-              {/* TODO: Replace later on with real data */}
-              {mockCompoundResourceData.totalCount}
+              {/* TODO: ask lucas if this should be metadata total or resourceSummary total */}
```
LukasStordeur: Considering it's part of the Summary concept, I'd take that one. We'll see based on feedback if there's a need for the other one instead.

src/Slices/Resource/UI/ResourcesPage/ResourcesTable.tsx
```diff
@@ -65,15 +65,8 @@ export const ResourcesTable: React.FC<Props> = ({
...
-      {rows.map((row, index) => (
-        <ResourceTableRow
-          row={row}
-          key={row.id}
-          index={index}
-          isExpanded={isExpanded(row.id)}
-          onToggle={onExpansion(row.id)}
-          numberOfColumns={tablePresenter.getNumberOfColumns()}
-        />
+      {rows.map((row, i) => (
```
LukasStordeur: `index` was fine ;)

src/Slices/Resource/UI/ResourcesPage/ResourcesTable.tsx
```diff
@@ -46,17 +42,21 @@ export const ResourcesTable: React.FC<Props> = ({
...
         {displayName}
       </Th>
     );
   });
 
   return (
-    <Table {...props} variant={TableVariant.compact} isStickyHeader>
-      <Thead>
+    <Table {...props} isStickyHeader>
+      <Thead className="resources-table-head">
```
LukasStordeur: What does the class do?

src/Slices/Resource/UI/ResourcesPage/ResourcesTablePresenter.ts
```diff
@@ -1,67 +1,43 @@
...
+  return resources.map((resource) => ({
+    type: resource.resourceType,
+    value: resource.resourceIdValue,
+    agent: resource.agent,
+    status: {
+      blocked: resource.state?.blocked,
+      compliance: resource.state?.compliance,
+      lastHandlerRun: resource.state?.lastHandlerRun,
+      lastHandlerRunAt: resource.state?.lastHandlerRunAt
+        ? moment.utc(resource.state.lastHandlerRunAt).format("DD MMMM YYYY, HH:mm")
```
LukasStordeur: You might want to use the DateWithTimeDiffTooltip in the rows instead of formatting it here :)

src/Slices/Resource/UI/ResourcesPage/ResourcesTablePresenter.ts
LukasStordeur, line 1: Classes aren't always bad, what was your motivation to move to a different implementation?

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,192 @@
...
 
-interface Props {
-  row: Resource.Row;
-  isExpanded: boolean;
-  onToggle: () => void;
-  numberOfColumns: number;
-  index: number;
-}
+/** Orphans are not actively a part of the latest intent anymore so limited information is displayed for them. */
+const PopoverContent = ({ row }: { row: Resource.Row }) => {
```
LukasStordeur, line 22: Probably would be nice to display in the popover that the resource is orphaned.

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,192 @@
...
+        </FlexItem>
+        <Content component="p">
+          {words("resources.popover.lastDeployed")(row.status.lastHandlerRunAt)}
+        </Content>
+      </Flex>
+    );
+  }
+
+  return (
+    <Flex direction={{ default: "column" }} gap={{ default: "gapMd" }}>
```
LukasStordeur, line 53: You probably can optimize this to be less repetitive.

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,192 @@
...
+      <Flex alignItems={{ default: "alignItemsCenter" }} gap={{ default: "gapSm" }}>
+        <FlexItem style={{ display: "inline-flex" }}>
+          {statusGroupIcons["blocked"]({ state: row.status.blocked })}
+        </FlexItem>
+        <Content component="p">
+          {capitalize(row.status.blocked.toLowerCase().replace(/_/g, " "))}
```
LukasStordeur: Using a one time Enum/mapping for this would be more lightweight than applying a lowercase+replace on each single row, on overy single view update

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,192 @@
...
+        <FlexItem style={{ display: "inline-flex" }}>
+          {statusGroupIcons["lastHandlerRun"]({ state: row.status.lastHandlerRun })}
+        </FlexItem>
+        <Content component="p">
+          {capitalize(row.status.lastHandlerRun.toLowerCase().replace(/_/g, " "))}
+        </Content>
+      </Flex>
+      <Flex alignItems={{ default: "alignItemsCenter" }} gap={{ default: "gapSm" }}>
+        <FlexItem style={{ display: "inline-flex" }}>
+          <ClockIcon
```
LukasStordeur:
```
      <Icon size="lg">
        <ClockIcon />
      </Icon>
```

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,192 @@
...
+          />
+        </FlexItem>
+        <Content component="p">
+          {row.requiresLength}{" "}
+          {row.requiresLength === 1
+            ? words("resources.popover.requirement")
+            : words("resources.popover.requirements")}
+        </Content>
+      </Flex>
+      {row.status.isDeploying && (
```
LukasStordeur, line 99: This type of conditional statement is the reason why you get the glitch in your rows. Either have a placeholder with the same width as the dot, or make the dot transparent when it's not deploying.

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,192 @@
...
       </Tr>
-    )}
-  </Tbody>
-);
+    </Tbody>
+  );
+};
+
+const StyledTD = styled(Td)`
+  /* Removes padding from cell and moves it to the Flex */
```
LukasStordeur, line 195: Why? The idea is not to fight the design system, but work along their guidelines. 
If you're trying to match the figma frames, they're only a suggestion on how it should look. If you can't get the exact same look, that's totally fine :) 
Using overrides likes these will bite us later, when we have a new version of patternly that gets released.

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,192 @@
...
+                }}
+              >
+                <DeployingDot />
+              </FlexItem>
+            )}
+            <FlexItem style={{ display: "inline-flex" }}>
+              {statusGroupIcons["blocked"]({
+                state: row.status.blocked,
+                // We override color because whenever orphan we just show gray icons
+                overrideColor: row.status.isOrphan ? "var(--pf-t--color--gray--60)" : undefined,
```
LukasStordeur: I think it would be nicer to simply highlight the entire row in a grayish hue. I don't know by heart, but I think PF rows support the `striped` attribute on each row. 
As for the icons, they can fall back to a disabled color.

src/UI/Components/CompoundResourceStatus/CompoundResourceStatus.tsx
```diff
@@ -13,30 +14,43 @@ const isCompoundStatusEntry = (
...
+
   const onClick = (id: Resource.CompoundStateType) => {
-    return updateFilter((filter) => ({ ...filter, status: [id] }));
+    // This makes sure we can only filter 1 stateType for every compound state
+    // It also filters out the stateType whenever u click on it again
+    return updateFilter((filter) => {
+      const group = STATE_FIELD_MAP[id];
+      const isActive = filter.status?.includes(id);
+      const withoutGroup = (filter.status ?? []).filter(
+        (s) => STATE_FIELD_MAP[s as Resource.CompoundStateType] !== group
```
LukasStordeur: Try to avoid casting, if you need it for code we design ourselves, there is often something wrong to begin with, or a type guard method missing

src/UI/Components/CompoundResourceStatus/CompoundResourceStatus.tsx
```diff
@@ -13,30 +14,43 @@ const isCompoundStatusEntry = (
...
-  totalCount,
+  resourceSummary: { totalCount, blocked, compliance, lastHandlerRun },
   updateFilter,
 }: CompoundResourceProps) => {
+  const compoundState: Resource.CompoundState = { blocked, compliance, lastHandlerRun };
+
   const onClick = (id: Resource.CompoundStateType) => {
-    return updateFilter((filter) => ({ ...filter, status: [id] }));
+    // This makes sure we can only filter 1 stateType for every compound state
+    // It also filters out the stateType whenever u click on it again
```
LukasStordeur: Why only one stateType? I'm also not entirely sure it's intuitive that clicking again would remove the filtering. I'd personnaly expect clicking on any of the statebar sections, would simply add that specific chip to the filtering

src/UI/Components/CompoundResourceStatus/config.tsx
```diff
@@ -43,12 +43,31 @@ export const statusPriority: Record<Resource.CompoundStateType, number> = {
...
   height: t_global_icon_size_font_xl.value,
-  color: "var(--pf-t--color--gray--50)",
 };
 
+/** We normalize the state toLowerCase since the resourceSummary has lowerCase
+ * and the resources themselves return upperCase states.
+ */
+const resolveColor = (
+  state?: Resource.CompoundStateType | Resource.CompoundStateUpperType,
+  overrideColor?: string
```
LukasStordeur: overriding should be used carefully, and not be considered as a standard implementation, especially when making something new. A default, or disabled variation can be better.

src/UI/Components/DeployingDot/DeployingDot.css
```diff
@@ -0,0 +1,18 @@
+.deploying-dot {
```
LukasStordeur: I'm pretty sure you could reuse the compile dot for this. You can rename it for a more generic use.

`CompileReportsIndication`

LukasStordeur (changes requested):
I saw you had various conversations about states ect, but I can't find anything about it in the code as additional context. It would be a nice thing to add imo, instead of pinning it in slack somewhere ;)

LukasStordeur:
What issue did you face with the popover element?

src/Slices/Resource/UI/ResourcesPage/ResourcesTablePresenter.ts
AronH99: I'd like to reuse a helper or a component but i don't think the on u mention is very fitting, it's already in a popover and to then have another tooltip popover in a popover.. seems a little much.
If there is any other function or component i can use i'd like to or otherwise i can make myself one which only converts ISO strings?
Let me know

src/Slices/Resource/UI/ResourcesPage/ResourcesTablePresenter.ts
AronH99: Honestly i very much dislike classes. 
Overall functional programming is:
1. Easier to read
2. Less boilerplate (in this case) => got way less lines of code with just removing it.
3. Easier to test and implement quickly
5. Better for three shaking, u now just import whatever u need from this TablePresenter.

I also believe classes aren't really useful if the class itself hasn't got a state of it's own to manage which was the case here.

src/Slices/Resource/UI/ResourcesPage/ResourcesTablePresenter.ts
LukasStordeur: The thing is, I suspect they will want to have the relative time like we use everywhere else. (x-days/x-minutes/x-hours ago) and in that case, it requires a tooltip to show the actual date. (I know it's not in the figma design)

If need be, you can split the DateWithTimeDiffTooltip component in two parts (and rename it), one with tooltip, and one with the date between brackets. So we can keep the exact same format accross the app.

src/Slices/Resource/UI/ResourcesPage/ResourcesTablePresenter.ts
LukasStordeur: I do prefer functional programming too, but I also like and prefer consistency and scoping tasks. This TablePresenter is a concept that is used in all tables so far, and if we decide to change it, it should be agreed on first, and scoped :)

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
AronH99: Didnt do the striped suggestion since it still had to be clear to the user that it can click on the icons, your disabled/default refactor suggestion has been implemented tho

src/Slices/Resource/UI/ResourcesPage/ResourcesTablePresenter.ts
AronH99: I see, well u want me to remake it into a class then? 
I feel like it will honestly just take more work unnecessarily here...
But i get where you are coming from.
Your decision of course

LukasStordeur: I'm not opposed moving away from the class implementation. But I'd prefer it in a fully separated task so we have better control over how we want to make the shift. Right now, you mimicked one on one the class, and lose the class blueprint that imposes a certain structure. Now I agree not all methods are relevant all the time. I think this needs a bit more thinking than being done on the fly in the middle of a different refactor

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
AronH99: So in this case i need the custom styling for the reason we discussed about the hover and rendering issues.. (also the popover hover was inconsistent so i opted for the onClick variant)
I adjusted it a little as we discussed in our meeting to make everything very consistent without any ui layout shift happening.
It should be minimal styling changes for optimal layout now.
Also using the correct tokens so hopefully they get supported in future versions as well so it should be kind of future proof or no?

src/Slices/Resource/UI/ResourcesPage/ResourcesTablePresenter.ts
AronH99: https://github.com/inmanta/web-console/issues/6809

made a ticket for it and added a comment to the ResourcesTablePresenter file.

AronH99:
So just to be clear again, I implemented the popover click variant (with its necessary styling and hover) because the hover from patternfly was kind of glitchy with sometimes showing me two popovers at the same time.
As a bonus I also think the current implementation with the onClick is better UI/UX wise.

Also not all tests regarding the resource page are passing but i will address that in my last pr since probably a lot more things will  change by then.

Also:
Screenshot 2026-04-15 at 09 24 33

src/Core/Domain/Resource/Resource.ts
```diff
@@ -4,60 +4,117 @@ import { CurrentPage } from "@/Data/Common/UrlState/useUrlStateWithCurrentPage";
...
+    requiresLength: number;
+    state: ResourceState;
+  };
 }
 
-export interface FromVersion extends Omit<Resource, "status"> {
-  resource_version_id: string;
-}
+/** Unwrapped resource node — convenience alias for use after unwrapping the API response. */
+export type FlatResource = Resource["node"];
```
LukasStordeur, line 98: Why isn't Resource the default, and RawResource for example the raw data incoming from the api? I'd find it more logical

src/Slices/Resource/UI/ResourcesPage/Page.tsx
```diff
@@ -78,16 +85,16 @@ export const Page: React.FC = () => {
...
+  const resources = data?.resources;
+  const resourceSummary = data?.resourceSummary;
+  const rows = createRows(resources);
+
   return (
     <>
       <PageSection
         hasBodyWrapper={false}
         style={{
-          position: "sticky",
```
LukasStordeur, line 86: Why remove it?

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
```diff
@@ -0,0 +1,144 @@
...
+      filter.isOrphan = true;
+      continue;
+    }
+
+    if (s === "!orphaned") {
+      filter.isOrphan = false;
+      continue;
+    }
+
+    if (!isCompoundStateKey(s)) continue;
```
LukasStordeur, line 65: if it's not a compounStateKey, you continue... This one doesn't make sense to me

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    const { field, value } = STATUS_FIELD_MAP[s];
+
+    switch (field) {
```
LukasStordeur: can't you do:
`[field].push(value)` instead of a switch case?

src/Slices/Resource/UI/ResourcesPage/Components/CompileReportsIndication.tsx
```diff
@@ -7,17 +7,19 @@ const pendingAnimation = keyframes`
  100% { opacity: .2}
 `;
 
-export const CompileReportsIndication = styled.span`
+export const CompileReportsIndication = styled.span<{
```
LukasStordeur, line 10: Since it's now used for more than the CompileReports, I'd rename it to something more generic.

src/Slices/Resource/UI/ResourcesPage/Page.tsx
```diff
@@ -100,8 +107,7 @@ export const Page: React.FC = () => {
               {words("inventory.tabs.resources")}
             </Content>
             <Label icon={<CubesIcon />} variant="outline" color="blue">
-              {/* TODO: Replace later on with real data */}
-              {mockCompoundResourceData.totalCount}
+              {resourceSummary?.totalCount ?? 0}
```
LukasStordeur, line 110: You'll add later the isDeploying count / totalCount, when there are deploying resources, with a small spinner?

src/Core/Domain/Resource/Resource.ts
```diff
@@ -4,60 +4,117 @@ import { CurrentPage } from "@/Data/Common/UrlState/useUrlStateWithCurrentPage";
...
 
-export interface IdDetails {
-  resource_type: string;
+/** Table row shape used for the resource view. */
+export interface Row {
```
LukasStordeur, line 101: I know you kept how it was before. But to me Row lives in the UI layer, not in the Domain layer. So I'd add this type to a more localized place related to the table.

src/UI/Components/CompoundResourceStatus/config.tsx
```diff
@@ -1,54 +1,104 @@
...
+  variant?: IconVariant
+): string | undefined => {
+  if (variant === "default") {
+    return "var(--pf-t--color--gray--50)";
+  }
+  if (variant === "disabled") {
+    return "var(--pf-t--color--gray--60)";
+  }
+
+  return state ? colorConfig[state.toLowerCase() as Resource.CompoundStateKey] : undefined;
```
LukasStordeur, line 79: You created a typeGuard for the keys, you can use it here. `isCompoundStateKey()`

src/UI/Components/CompoundResourceStatus/config.tsx
```diff
@@ -1,54 +1,104 @@
...
+ */
+type IconVariant = "default" | "disabled";
+
+/** Resolves the correct colors for every compound state Icon. */
+const resolveColor = (
+  state?: Resource.CompoundState,
+  variant?: IconVariant
+): string | undefined => {
+  if (variant === "default") {
+    return "var(--pf-t--color--gray--50)";
```
LukasStordeur, line 73: https://www.patternfly.org/patterns/status-and-severity/#severity-icons

I'd pick among those tokens for the colors.

src/UI/Components/CompoundResourceStatus/config.tsx
```diff
@@ -1,54 +1,104 @@
...
-  color: "var(--pf-t--color--gray--50)",
+/** default is for the legendbar icons and disabled is for the orphaned resources.
+ * both implementations can be found on the resource page.
+ */
+type IconVariant = "default" | "disabled";
+
+/** Resolves the correct colors for every compound state Icon. */
+const resolveColor = (
+  state?: Resource.CompoundState,
+  variant?: IconVariant
```
LukasStordeur, line 70: It makes sense to assign default, so you don't have to explictly pass "default"
```suggestion
  variant: IconVariant = "default"
```

src/UI/Components/LegendBar/LegendBar.tsx
```diff
@@ -25,7 +25,7 @@ export const LegendBar: React.FC<Props> = ({ items, total, label, ...props }) =>
         {items.length <= 0 ? (
           <LoneItem key="none" label={label || ""} />
         ) : (
-          items.map((item) => <Item key={item.id} {...item} height="20px" />)
```
LukasStordeur, line 28: You're removing the height, but I believe this legendBar is used in other places. We don't want to affect those.

src/UI/Components/CompoundResourceStatus/config.tsx
```diff
@@ -1,54 +1,104 @@
...
+): string | undefined => {
+  if (variant === "default") {
+    return "var(--pf-t--color--gray--50)";
+  }
+  if (variant === "disabled") {
```
LukasStordeur, line 75: perhaps using "orphan" might be more explicit?

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,218 @@
...
+  text: React.ReactNode;
+}> = ({ icon, text }) => (
+  <Flex alignItems={{ default: "alignItemsCenter" }} gap={{ default: "gapSm" }}>
+    <FlexItem style={{ display: "inline-flex" }}>{icon}</FlexItem>
+    <Content component="p">{text}</Content>
+  </Flex>
+);
+
+/** Orphans are not actively a part of the latest intent anymore so limited information is displayed for them. */
+const PopoverContent = ({ row }: { row: Resource.Row }) => {
```
LukasStordeur, line 22: Wouldn't it be better to extract this into its own file? 
It would make it easier to follow along what is what, because now, you have two sorts of Rows, one for the table, and one that is in fact not a real row, but an item in your popover dialog.
It would also allow to rename this to a more coherent component name, such as StateInfo or ResourceStateInfo, whichever you like best.

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
LukasStordeur: Why don't you place the icons in a button component instead of customizing the cell to act like a button? (Yes, I'm still convinced you don't need all that customization)

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,218 @@
...
+              <FlexItem style={{ display: "inline-flex" }}>
+                {statusGroupIcons["blocked"]({
+                  state: row.status.blocked,
+                  variant: row.status.isOrphan ? "disabled" : undefined,
+                })}
+              </FlexItem>
+              <FlexItem style={{ display: "inline-flex" }}>
+                {statusGroupIcons["compliance"]({
+                  state: row.status.compliance,
+                  variant: row.status.isOrphan ? "disabled" : undefined,
```
LukasStordeur, line 165: is the variant undefined supported? Or is it supposed to be default?

src/Slices/Resource/UI/ResourcesPage/Page.tsx
```diff
@@ -78,16 +85,16 @@ export const Page: React.FC = () => {
     return <LoadingView ariaLabel="ResourcesPage-Loading" />;
   }
 
+  const resources = data?.resources;
+  const resourceSummary = data?.resourceSummary;
+  const rows = createRows(resources);
```
LukasStordeur, line 90: createRows doesn't support undefined, which can happen if data is not defined in line 88. This will throw.

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
```diff
@@ -0,0 +1,144 @@
+import { Resource, Sort } from "@/Core/Domain";
+import { Handlers } from "@/Core/Domain/Pagination/Pagination";
+import { CurrentPage } from "@/Data/Common";
+
+/** This is used to map the statusses of all compound states correctly */
+const STATUS_FIELD_MAP = {
```
LukasStordeur, line 6: This still bothers me and seems redundant. 

You're using it to type check, while you have the CompoundState union type for this. 

If you want to refine to which type of state a value belongs, I'd use your typing to refine it, not redeclare the value/pairs in this manner. 

I need to think a bit more about this

LukasStordeur (changes requested):
I know there's the page test you cannot fix because you're waiting on the graphQL update (what's the status of that by the way?) 
But there are plenty of new additions that can be scoped without requiring the graphQL endpoint. 
I'm also not yet seeing additional information that was shared about the intrisic designs of the compound state :)

src/Slices/Resource/UI/ResourcesPage/Page.tsx
AronH99: i don't get what you are asking here?

src/Slices/Resource/UI/ResourcesPage/Page.tsx
AronH99: Yes the section being sticky served no purpose?

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
AronH99: Continue skips the entire iteration of the for loop, meaning if it's not one of those keys we have specified in the mapping we dont even want to move one

src/Slices/Resource/UI/ResourcesPage/Page.tsx
LukasStordeur: Yes, it did serve a purpose :)

src/Slices/Resource/UI/ResourcesPage/Page.tsx
LukasStordeur: That the badge should also display the amount that is currently deploying whenever there are deploying resources :)

src/UI/Components/LegendBar/LegendBar.tsx
AronH99: no this was the misstake i was talking about i made earlier, it was the other way around, i applied it for everything in my previous pr, with this fix it is now rectified so that it isn't applied everywhere

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
AronH99: well no it is supposed to be undefined, i'll change the naming of default because i see that it might be confusing..
its just variant can be 2 things because i need two variants for these icons..
otherwise it follows the color config

src/Slices/Resource/UI/ResourcesPage/Page.tsx
AronH99: it normally gets caught by:

```
  if (isError) {
    return <ErrorView message={error.message} ariaLabel="ResourcesPage-Error" retry={refetch} />;
  }
  if (!isSuccess) {
    return <LoadingView ariaLabel="ResourcesPage-Loading" />;
  }
```
  
 because otherwise i would think the data we have would be valid...
 hence why i probably don't get any ts issues
 or am i missing something here?

src/Slices/Resource/UI/ResourcesPage/Page.tsx
AronH99: ResourceSummary total count does take into account the deploying resources?
because it will return with all kinds of compound state resources and add them all together?

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
AronH99: Do you have any suggestions because honestly i can do 5 more iterations but i'd think none of them ever will be perfect, to me this is quite an elegant solution which should indicate any ts issues whenever we would make adjustments/api changes arise..
But i wouldn't know any better way to do this without rewriting my ts file again?

src/Slices/Resource/UI/ResourcesPage/Page.tsx
LukasStordeur: Yes, but you have a counter for isDeploying, no? If there are resources deploying, it should be displayed in the badge:
`${isDeployingCount}/${totalCount}` (this is just to illustrate what I mean, but you'll need to add the spinner and conditional ect)

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
LukasStordeur: I still don't see what the usecase is for the undefined?

src/Slices/Resource/UI/ResourcesPage/Page.tsx
LukasStordeur: const resources = data?.resources; => This says, data could potentially be undefined. Otherwise, you wouldn't need the ?-operator.

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
LukasStordeur: Make a ticket, I'll come back to it later.

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
AronH99: `{statusGroupIcons["blocked"]({
   state: row.status.blocked,
   variant: row.status.isOrphan ? "disabled" : undefined,
})}`
              
if we don't pass undefined the default would take over...
so then the state colors won't come through on the icons

src/Slices/Resource/UI/ResourcesPage/Page.tsx
AronH99: Bad habit of mine, i can remove the ?-operator because the hook already returns [] in case there really is no data available

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
LukasStordeur: > if we don't pass undefined the default would take over

This doesn't make sense, logically speaking. Passing undefined to prevent the default to take over. That sounds like an implementation flaw.

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
LukasStordeur, line 1: Ok, did some more thinking about this. 

Since typeGuards are something we'll need more, I'd make a factory for those.
```ts
function createTypeGuard<const T extends readonly string[]>(values: T) {
  const set = new Set<string>(values); // iterating sets are usually more performant.

  return (value: string): value is T[number] => {
    return set.has(value);
  };
}
```

This is just for one of them, so you get the gist of it.
```ts
export const COMPLIANCE_VALUES = [
  "COMPLIANT",
  "NON_COMPLIANT",
  "HAS_UPDATE",
  "UNDEFINED",
] as const;

export type Compliance = typeof COMPLIANCE_VALUES[number]; 
// this results in 
// type Compliance =
//  | "COMPLIANT"
//  | "NON_COMPLIANT"
//  | "HAS_UPDATE"
//  | "UNDEFINED";
```

Then you can have dedicated typeguards where needed for those:

```ts
export const isLastHandlerRun = createTypeGuard(LAST_HANDLER_RUN_VALUES);
export const isCompliance = createTypeGuard(COMPLIANCE_VALUES);
export const isBlocked = createTypeGuard(BLOCKED_VALUES);

export function isCompoundState(value: string): value is CompoundState {
  return (
    isLastHandlerRun(value) ||
    isCompliance(value) ||
    isBlocked(value)
  );
}
``` 
That should get rid of the need of the extra mapping object you needed.

AronH99: From the typeguard perspective this might be a good idea but it still doesn't solve the core of the issue which is mapping the states to the correct filters?
=> can't really remove the STATUS_FIELD_MAP entirely can I?

AronH99: I'll leave this open for now and commit what i have already so you can review whatever is necessary.

AronH99:
So i normally addressed all of your comments, went through the unit tests they should all be passing right now. 
Also made the pagination of the resource table work in both directions.
And as u can see on the screenshot i reused the deploy dot since that made most sense

Screenshot 2026-04-17 at 09 29 10

src/Core/Domain/Resource/Resource.ts
```diff
@@ -4,60 +4,144 @@ import { CurrentPage } from "@/Data/Common/UrlState/useUrlStateWithCurrentPage";
...
-/** Flat record of all compound state types mapped to counts. */
-export type StateRecord = Record<CompoundStateType, number>;
-
-/** Status counts grouped by state. */
-export interface CompoundState {
-  blocked: Record<BlockedState, number>;
-  compliance: Record<ComplianceState, number>;
-  lastHandlerRun: Record<LastHandlerRunState, number>;
-}
+/** Unwrapped resource node — convenience alias for use after unwrapping the API response. */
```
LukasStordeur: This is the default now, not the convenience alias, right?

src/Data/Common/usePaginatedTable.ts
```diff
@@ -37,11 +37,11 @@ export function usePaginatedTable<TFilter = undefined, TSort extends string = st
     route,
   });
 
-  //when sorting is triggered, reset the current page
+  //when sorting or filtering is triggered, reset the current page
   useEffect(() => {
     setCurrentPage({ kind: "CurrentPage", value: "" });
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [sort.order, sort.name]);
+  }, [sort.name, sort.order, JSON.stringify(filter)]);
```
LukasStordeur, line 46: Why JSON.stringify? 
It has a few pitfalls, and not always worth it concerning performance. 
The resulted string comparison doesn't care about order of keys/properties/values. 
Was the reference comparisson not sufficient?

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
LukasStordeur: For the typing you can consider this for the next steps and simplify your mapping to one single source of truth ( not for this pr). 

```ts
export const COMPLIANCE = {
  compliant: "COMPLIANT",
  non_compliant: "NON_COMPLIANT",
  has_update: "HAS_UPDATE",
  undefined: "UNDEFINED",
} as const;

export type ComplianceKey = keyof typeof COMPLIANCE;
// "compliant" | "non_compliant" | "has_update" | "undefined"

export type ComplianceValue = typeof COMPLIANCE[ComplianceKey];
// "COMPLIANT" | "NON_COMPLIANT" | "HAS_UPDATE" | "UNDEFINED"
```

src/Data/Queries/Slices/Resource/GetResources/useGetResources.ts
```diff
@@ -218,11 +142,15 @@ export const useGetResources = (params: GetResourcesParams): GetResources => {
     ...statusFilter,
   };
 
+  const pageSizeNum = Number(pageSize.value);
+  const paginationVariables: Record<string, unknown> = before
```
LukasStordeur: Shouldn't this have a more defined typing?

src/Slices/Resource/UI/ResourcesPage/Page.test.tsx
```diff
@@ -162,211 +151,107 @@ describe("ResourcesPage", () => {
...
     expect(
       await screen.findByRole("generic", { name: "ResourcesPage-Empty" }, { timeout: 5000 })
     ).toBeInTheDocument();
 
-    await act(async () => {
-      const results = await axe(document.body);
-
-      expect(results).toHaveNoViolations();
-    });
+    await act(async () => expect(await axe(document.body)).toHaveNoViolations());
```
LukasStordeur: Why did you change it? It used to be blocked statements.

src/Slices/Resource/UI/ResourcesPage/Page.test.tsx
LukasStordeur, line 1: Plenty of places where you moved from blocked statements to one-liners, was that intended?

src/Slices/Resource/UI/ResourcesPage/Page.tsx
LukasStordeur: You might need to keep the index part so the sorting headers of the tables don't overflow. Not sure if it's still the case.

src/Slices/Resource/UI/ResourcesPage/Page.tsx
```diff
@@ -99,9 +109,22 @@ export const Page: React.FC = () => {
             <Content component="h1" style={{ marginBottom: 0 }}>
               {words("inventory.tabs.resources")}
             </Content>
-            <Label icon={<CubesIcon />} variant="outline" color="blue">
-              {/* TODO: Replace later on with real data */}
-              {mockCompoundResourceData.totalCount}
+            <Label
+              icon={<CubesIcon />}
```
LukasStordeur, line 112: I would have seen it differently, and replaced the CubesIcon with the Spinner (small) component, when there's something deploying. The blue dot, on a blue background is not ideal. In other places, the spinner is used in labels to indicate something is loading/pending

src/Slices/Resource/UI/ResourcesPage/ResourceStateInfo.tsx
```diff
@@ -0,0 +1,106 @@
+import { Flex, FlexItem, Content, Icon } from "@patternfly/react-core";
+import { UnlinkIcon, ClockIcon, CubesIcon } from "@patternfly/react-icons";
+import { words } from "@/UI";
+import { DateWithTooltip, statusGroupIcons, statusMapping } from "@/UI/Components";
+import { BlinkingDot } from "./Components";
+import { ResourceRow } from "./ResourceTableRow";
+
+const RowComponent: React.FC<{
```
LukasStordeur: Rows are typically present in tables. 
Patternfly has a list component that supports icons. https://www.patternfly.org/components/list/#with-icons

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
```diff
@@ -1,64 +1,105 @@
...
+          <Flex
+            gap={{ default: "gapNone" }}
+            flexWrap={{ default: "nowrap" }}
+            alignItems={{ default: "alignItemsCenter" }}
+            style={{
+              height: "100%",
+              justifySelf: "flex-end",
+            }}
+          >
+            <FlexItem
```
LukasStordeur: Do you still require the wrapping flex?
I might be wrong, but I feel like you're over using it with a lot of custom styling

src/UI/Components/CompoundResourceStatus/CompoundResourceStatus.tsx
```diff
@@ -62,6 +72,7 @@ export const CompoundResourceStatus = ({
         value,
         backgroundColor: colorConfig[status],
         label: status,
+        height: "20px",
```
LukasStordeur, line 75: Didn't you remove this in a previous itteration?

src/UI/Components/CompoundResourceStatus/config.tsx
```diff
@@ -1,54 +1,101 @@
...
 
-  [Resource.ComplianceState.undefined]: "var(--pf-t--color--gray--40)",
-  [Resource.LastHandlerRunState.skipped]: "var(--pf-t--color--gray--40)",
+const DEFAULT_COLOR = "var(--pf-t--global--icon--color--severity--minor--default)";
 
-  [Resource.LastHandlerRunState.new]: "var(--pf-t--color--blue--50)",
-};
+/** Resolves the correct colors for every compound state Icon. */
+const resolveColor = (variant: ColorVariant = "state", state?: Resource.CompoundState): string => {
+  if (variant === "default") return DEFAULT_COLOR;
```
LukasStordeur, line 69: Can't you assert if the key is present in the config, rather than have various different if cases that return default?

LukasStordeur (changes requested):

src/Data/Queries/Slices/Resource/GetResources/helpers.ts
AronH99: i copied your comment to the ticket i made to restructure that if that's fine?
https://github.com/inmanta/web-console/issues/6814

src/Slices/Resource/UI/ResourcesPage/Page.test.tsx
AronH99: u like the blocked versions better?
it's just something to reduce lines of code while it's not too hard to read?
i can change it back if u want

LukasStordeur: It's about readability and consistency :)

LukasStordeur: Splitting it also makes it easier to inspect failures when they show up. 
Having one liners can seem practical, but they rarely are :)

src/Data/Common/usePaginatedTable.ts
AronH99: no reference comparrison is always false because {} !== {} which means even if i dont touch the filters it will clear my currentPage which is definitly something i don't want..
Hence the json stringify for the filter.. if u got a better sollution for this do let me know?

src/UI/Components/CompoundResourceStatus/CompoundResourceStatus.tsx
AronH99: this is correct, i removed it in another file which was intended, and added it here to have all of the legend bars being 20px height

src/Data/Queries/Slices/Resource/GetResources/useGetResources.ts
AronH99: removed it because typing was automatically assigned anyway

src/Slices/Resource/UI/ResourcesPage/Page.tsx
AronH99: i'll ask this in the frontend chat what people might find more logical

src/Data/Common/usePaginatedTable.ts
LukasStordeur: This would be true, if the reference wasn't created with useState. It would only be triggering a rerender if a `setFilter({});` would be executed. And if that's the case, then the issue is not in your dependency hook, but the place where you are reassigning an empty object while this is not required.

src/Slices/Resource/UI/ResourcesPage/Page.tsx
AronH99: ticket:
https://github.com/inmanta/web-console/issues/6816

src/Data/Common/usePaginatedTable.ts
AronH99: the thing is that filter changes get added to the url so the usePaginatedTable , to be more specific the useUrlStateWithFilter will pick this up off course and then the useEffect will trigger because a new reference gets created..
Would u like me to adjust the useUrlStateWithFilter to have stable references in between renders or keep the current stringify fix?
Both solutions have advantages and disadvantages i think so yeah..

LukasStordeur: I think it's out of scope for this task. Make a ticket for this. I think updating the useUrlStatexxx will be beneficial.

src/Slices/Resource/UI/ResourcesPage/ResourceTableRow.tsx
AronH99: i ll try to go through patternfly more often if i need something because u were right

src/Data/Common/usePaginatedTable.ts
AronH99: done
https://github.com/inmanta/web-console/issues/6817

src/Slices/Resource/UI/ResourcesPage/Page.test.tsx
LukasStordeur, line 1: I think we'll need to discuss the need of one-liners vs splitting logic in pieces. 
I understand that some files can be long, especially the test files, but this is not the right approach to make them shorter. On top of it, you're adding more changes that aren't really necessary, which then adds more review load on an already considerable pull request.
If you have concerns about certain guidelines, or format, it's ok. But a pull request of this size is not the right place to introduce such new standards :)

src/Slices/Resource/UI/ResourcesPage/Page.test.tsx
```diff
@@ -531,115 +423,87 @@ describe("ResourcesPage", () => {
...
-      await userEvent.type(inputOne, `${filterValueOne}{enter}`);
-
-      const inputTwo = await screen.findByPlaceholderText(placeholderTextTwo);
-
-      await userEvent.type(inputTwo, `${filterValueTwo}{enter}`);
-
-      const updatedRows = await screen.findAllByRole("row", {
-        name: "Resource Table Row",
-      });
+      await userEvent.type(
```
LukasStordeur, line 445: I don't see the added value of these changes. If anything, for me, it makes the code less readable.

src/UI/Components/CompoundResourceStatus/config.tsx
```diff
@@ -1,54 +1,97 @@
...
-  height: t_global_icon_size_font_xl.value,
-  color: "var(--pf-t--color--gray--50)",
+/** default is the neutral icons color and state is whenever the colorConfig should be used for the icons. */
+type ColorVariant = "default" | "state";
+
+const DEFAULT_COLOR = "var(--pf-t--global--icon--color--severity--minor--default)";
+
+/** Resolves the correct colors for every compound state Icon. */
+const resolveColor = (variant: ColorVariant = "state", state?: Resource.CompoundState): string => {
+  if (variant === "default") return DEFAULT_COLOR;
```
LukasStordeur, line 69: Your last return statement falls back to default, there's no need for the early return, no?
Is there even still a place where you pass in "default"?

src/UI/Components/CompoundResourceStatus/config.tsx
```diff
@@ -1,54 +1,97 @@
...
-  width: t_global_icon_size_font_xl.value,
-  height: t_global_icon_size_font_xl.value,
-  color: "var(--pf-t--color--gray--50)",
+/** default is the neutral icons color and state is whenever the colorConfig should be used for the icons. */
+type ColorVariant = "default" | "state";
+
+const DEFAULT_COLOR = "var(--pf-t--global--icon--color--severity--minor--default)";
+
+/** Resolves the correct colors for every compound state Icon. */
+const resolveColor = (variant: ColorVariant = "state", state?: Resource.CompoundState): string => {
```
LukasStordeur, line 68: why is state optional?

LukasStordeur (changes requested):

src/UI/Components/CompoundResourceStatus/config.tsx
AronH99: because we only need to know the state (compound state) for the icons whenever relevant, otherwise it's default for the orphans and legend bar icons

src/UI/Components/CompoundResourceStatus/config.tsx
AronH99: for the orphans i pass them yes in the ResourceTableRow

LukasStordeur: ok.

src/Slices/Resource/UI/ResourcesPage/Page.test.tsx
AronH99: I went through the test and for the most unreadable ones I changed them into splitted logic pieces again.

LukasStordeur (approved):
You can comment out for now the e2e test if you want, so the pipeline passes until you fix them.

PR #6804, opened 2026-04-14
Dependabot update jointjs

LukasStordeur:
# Description

Updating the yarnrc once more, hopefully, it will now work for dependabot again...


AronH99 (approved):
think u investigated this already alot, i approve so that you can test

PR #6808, opened 2026-04-14
update url to jointjs registry

LukasStordeur:
# Description

After digging... it could potentially be that packages hosted on gitlab, require a more explict url for dependabot. This is a longshot, because it's not anywhere in the documentation of jointjs that they are hosting on gitlab, it's just a suspicion. If that doesn't work... well I'm clueless. I'm sorry to bother with recurring prs x)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,7 +2,7 @@ version: 2
 registries:
   jointjs-registry:
     type: npm-registry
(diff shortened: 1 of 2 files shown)


AronH99:
no problem, it's probably a hard thing to figure out, that happens sometimes.

AronH99 (approved):

Issue #6781, opened 2026-03-26
Migrate Resource Discovery page filters to side panel drawer


LukasStordeur:
## Context

The Resources page filters have been migrated to a side panel drawer (using PatternFly `Drawer` + `DrawerPanelContent`). Other pages still use the old inline toolbar filter pattern and should be updated to match.

## Current state

`src/Slices/ResourceDiscovery/UI/TableControls.tsx` renders a `FilterForm` component inline inside a PatternFly `Toolbar`.

## Goal

Migrate the Resource Discovery page filters to the new side panel pattern:

- Move the `FilterForm` component into a `DrawerPanelContent`
- Add a `Drawer` + `DrawerContent` + `DrawerPanelContent` wrapping the table
- Add a filter toggle button to the toolbar (with active filter count badge), replacing the inline `FilterForm`
- Track `isDrawerExpanded` and `activeFilterCount` state on the page component

## Reference implementation

See `src/Slices/Resource/UI/ResourcesPage/Page.tsx` and `src/Slices/Resource/UI/ResourcesPage/Components/ResourceTableControls.tsx` for the pattern to follow.

PR #6811, opened 2026-04-16
Feature/6781 filter drawer refactor

LukasStordeur:
# Description

The Resource Discovery and Desired State Details pages now display filters in a side panel drawer, matching the pattern used on the Resource page. The page title and toolbar are sticky, and filter changes no longer cause the table to glitch thanks to React Query's keepPreviousData.

In addition, I'm trying yet another config for dependabot... 

closes:
- #6781

https://github.com/user-attachments/assets/a9b1fe11-a99b-47e0-8ed8-cec68a209e33


AronH99 (approved):
Reviewed, tested 
=> approved, couldn't find anything wrong with it, seems consistent with the rest of the codebase as well

PR #6815, opened 2026-04-21
Manual dependency update

LukasStordeur:
# Description

Manual update of a set of dependencies, plus reverting the yarnrc/dependabot config.

With the script changes too, now you can omit using two terminals. All you need to do is assign the backend url to the VITE_API_BASEURL


tsconfig.json
```diff
@@ -19,7 +19,6 @@
     "strict": true,
     "resolveJsonModule": true,
     "skipLibCheck": true,
-    "baseUrl": ".",
```
LukasStordeur, line 22: This is deprecated and will be broken in the next TS release.

AronH99 (approved):

Issue #6812, opened 2026-04-20
[REQUEST] Allow to download/copy mermaid diagrams from documentation tab


edvgui:
### Summary

Provide a clear and concise description of the request, and update the title of the ticket.

- **Area of the app**: Documentation tab of service instance
- **Type**: UX
- **affected versions**: iso8+ (all)

### Problem / Use Case

The documentation tab can contain nice mermaid diagrams, which are currently hard to extract.  Only solution is to inspect the DOM and copy the svg manually.

### Proposed Solution(s)

Add a copy/download button next/over the rendered mermaid diagram which allows to copy/download the svg.

### Impact

/

### Additional Context

/

PR #6818, opened 2026-04-22
6812 download diagrams

LukasStordeur:
# Description

Added a toolbar that displays a set of buttons to download the rendered diagrams. The user can download as svg of png. 
The png's do have a transparent background, so in the recording the text is a bit hard to read because I'm using a default darkmode on my device. 

I moved some of the methods in a helper file to make the component file a bit shorter. 

closes:
- #6812 

https://github.com/user-attachments/assets/7d837d34-19cb-4e56-8203-e1e67e826565


LukasStordeur:
I expect some conflicts with the previous pr once that is merged, I'll fix the build issue after that

AronH99:
U have a ts issue somewhere in your code so the integration fails.. can u check it out?

LukasStordeur:
> U have a ts issue somewhere in your code so the integration fails.. can u check it out?

Yes, I was aware of it, it was caused by the dependency updates coming from master :)

AronH99 (approved):
reviewed and tested, works fine so approved

Issue #6508, opened 2025-08-12
Update Resource ID filters + add Desired state filters


LukasStordeur:
Update the resource id filters (Type, agent, value) and add the Desired state filter Purged.
The agent field should pull the list of agents.

PR #6821, opened 2026-04-22
feat: reworked resource filters and added desired state filters

AronH99:
# Description

The following changes were made:
1) The resource filters are redone and all of them are conform with the design and requirements of the ticket.
2) Implemented a useInfiniteQuery from tanstack for the filtering of the agents input/select. 
Along with this I fixed the pagination to be able to go back from any page to the previous page.
3) Also made a small adjustment where the popover of the compound states in the table now have another popover at the icon part where once again u can see what compound state field it stands for (ex: blocked, compliance or lastHandlerRun)
I had that feedback from Arnaud so thought I'd work ahead on this one. 
4) Tests were adjusted to be conform with what's representative for this pr.
5) Made two custom hooks for my components to make use of.

Made a small video as quick example as well:

https://github.com/user-attachments/assets/63030d24-4554-4bff-90bc-f4730ddea1b1

https://github.com/inmanta/web-console/issues/6508


src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableSelectInput.tsx
```diff
@@ -0,0 +1,257 @@
...
+    }
+  };
+
+  const menuContent = (
+    <div
+      ref={menuRef}
+      onScroll={handleScroll}
+      style={{
+        // We do this because we want this to match the width of the inputs.
+        width: inputWidth ? `${inputWidth}px` : "auto",
```
LukasStordeur: Isn't that something patternfly covers by default? I never had to use a wrapping Div element like this for menus or dropdowns so far

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableSelectInput.tsx
LukasStordeur, line 1: You can look at the implementation of MultiTextSelect, it might already have what you need for this, or give you pointers on how to adjust this one.

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableTextInput.tsx
```diff
@@ -47,7 +62,30 @@ export const AddableTextInput: React.FC<AddableTextInputProps> = ({
   };
 
   return (
-    <FormGroup label={label} fieldId={inputId}>
+    <FormGroup
+      label={
+        <span style={{ fontWeight: "var(--pf-t--global--font--weight--body--bold)" }}>{label}</span>
```
LukasStordeur: If the default is not bold on patternfly, we shouldn't be overriding this with an inner span element. Try to avoid as much as possible to use native html elements in favor of Patternfly ones. We only fallback to those in cases where it's not possible otherwise.

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableTextInput.tsx
```diff
@@ -47,7 +62,30 @@ export const AddableTextInput: React.FC<AddableTextInputProps> = ({
...
+      }
+      fieldId={inputId}
+      labelHelp={
+        hint ? (
```
LukasStordeur, line 69: Is the ternary required? Can't you use the && operator like you did for the labelInfo?

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
```diff
@@ -18,40 +30,129 @@ export interface ResourceFilterFormProps {
...
+    useGetAgents().useInfiniteScroll({
+      pageSize: { kind: "PageSize", value: "20" },
+      filter: debouncedSearch ? { name: [debouncedSearch] } : undefined,
+    });
+
+  const agentOptions = useMemo<SelectOption[]>(
+    () =>
+      data?.pages.flatMap((page) =>
+        page.data.map((agent) => ({ value: agent.name, label: agent.name }))
+      ) ?? [],
```
LukasStordeur: So, this is maybe also a good case to illustrate why blocked return statements improve readability. 
Yes, the indentation is there to help a bit, but it would be more clear which part is the inner method being executed, and which part is the memo dependency.

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
```diff
@@ -18,40 +30,129 @@ export interface ResourceFilterFormProps {
...
+        </StackItem>
+        <StackItem>
+          <Switch
+            id={words("resources.filters.desiredState.reportOnly")}
+            aria-label={words("resources.filters.desiredState.reportOnly")}
+            label={
+              <Content component="p">{words("resources.filters.desiredState.reportOnly")}</Content>
+            }
+            isChecked={REPORT_ONLY_STATUSES.every((s) => filter.status?.includes(s)) ?? false}
+            onChange={(_e, bool) => {
```
LukasStordeur: It's better to extract this type of methods from the template, or they will be recreated on each rerender. Which impacts performance.
Same for the isChecked.

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
```diff
@@ -18,40 +30,129 @@ export interface ResourceFilterFormProps {
...
+          />
+        </StackItem>
+        <StackItem>
+          <Switch
+            id={words("resources.filters.desiredState.reportOnly")}
+            aria-label={words("resources.filters.desiredState.reportOnly")}
+            label={
+              <Content component="p">{words("resources.filters.desiredState.reportOnly")}</Content>
+            }
+            isChecked={REPORT_ONLY_STATUSES.every((s) => filter.status?.includes(s)) ?? false}
```
LukasStordeur, line 166: isn't `every` already returning a boolean by design?

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
```diff
@@ -18,40 +30,129 @@ export interface ResourceFilterFormProps {
...
+            onChange={(_e, bool) => {
+              const current = filter.status ?? [];
+              const updated = bool
```
LukasStordeur: bool is more a type than a readable variable. `hasChanged `might be more explicit

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
```diff
@@ -18,40 +30,129 @@ export interface ResourceFilterFormProps {
...
+                ? [...new Set([...current, ...REPORT_ONLY_STATUSES])]
+                : current.filter(
+                    (s) =>
```
LukasStordeur: Try not to use single letters as identifiers/variable names as a general rule of thumb

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
```diff
@@ -18,40 +30,129 @@ export interface ResourceFilterFormProps {
...
+                      !REPORT_ONLY_STATUSES.includes(
+                        s as Resource.ComplianceKey | Resource.LastHandlerRunKey
```
LukasStordeur: Is the casting needed, or do you have some flaw in the implementation somewhere?

src/Slices/Resource/UI/ResourcesPage/ResourceStateInfo.tsx
```diff
@@ -1,11 +1,44 @@
...
+import { Flex, FlexItem, Content, Icon, ListItem, List, Popover } from "@patternfly/react-core";
 import { UnlinkIcon, ClockIcon, CubesIcon } from "@patternfly/react-icons";
+import { Resource } from "@/Core";
 import { words } from "@/UI";
 import { DateWithTooltip, statusGroupIcons, statusMapping } from "@/UI/Components";
 import { BlinkingDot } from "./Components";
 import { ResourceRow } from "./ResourceTableRow";
 
-/** Orphans are not actively a part of the latest intent anymore so limited information is displayed for them. */
+const COMPOUND_STATE_KEYS: (keyof Resource.CompoundStateSummary)[] = [
```
LukasStordeur, line 9: I would have thought we'd had those in the domain typing file, no? 
I

src/Slices/Resource/UI/ResourcesPage/ResourceStateInfo.tsx
```diff
@@ -1,11 +1,44 @@
...
+}: {
+  compoundStateKey: keyof Resource.CompoundStateSummary;
+  row: ResourceRow;
+}) => {
+  const state = row.status[compoundStateKey];
+
+  return (
+    <ListItem
+      icon={
+        <Popover
```
LukasStordeur, line 27: I don't think a popover in a popover is the right way, and I think Bart isn't a fan of that type of thing.

src/UI/Utils/useClickOutside.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, RefObject } from "react";
+
+export const useClickOutside = (
```
LukasStordeur: Why? What are you trying to solve that isn't possible with the PF components?

src/UI/Utils/useDebounce.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect, useState } from "react";
+
+export const useDebounce = (value: string, delay: number): string => {
```
LukasStordeur, line 21: missing docstrings (I've seen other places too, where exported methods/components miss them)

LukasStordeur (changes requested):

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableSelectInput.tsx
AronH99: the main reason i did this because i found some weird issues with patternfly not adjusting the width on resize and also not spanning the whole input (including the + button)
so with this implementation it all remains very consistent.. 
i mean the base implementation is fine i guess but to me it feels bad..

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableSelectInput.tsx
AronH99: i know how to adjust it to have the width deleted but as i said the base implementation feels off, let me know if u still want me to do the base one

AronH99: i'll look into it because maybe it's just an oversight on my part

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableSelectInput.tsx
AronH99: nvm I can use your MultiTextSelect implementation and adjust my component as u said

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableSelectInput.tsx
AronH99: so yes i can use your kind of implementation to adjust my component as well, nvm then, i'll do that

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableTextInput.tsx
AronH99: gives ts issues, so in this case not really

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/AddableSelectInput.tsx
AronH99: sorry patternfly to me is a little hard to navigate sometimes.. i guess i should really spend more time figuring out what the correct components are to use..

LukasStordeur: That's ok, I've been working with it for years, and I know it by heart by now 😂

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
AronH99: i need it because filter.status is defined as string[]
i tried to type it better but it's quite a lot of work since the filter.status can be other things than compound states...
it's a bit complex is what i am saying so for now i'd like to leave it to the refactor types ticket we made maybe?

src/Slices/Resource/UI/ResourcesPage/ResourceStateInfo.tsx
AronH99: no no enums are part of the domain typing for now, so currently i am typing everything properly but no enums to work with so far.
we can leave this to the other ticket as well or no?

src/Slices/Resource/UI/ResourcesPage/ResourceStateInfo.tsx
AronH99: we keep it for now and see what feedback we get or no?

src/UI/Utils/useClickOutside.ts
AronH99: yes i deleted this one entirely

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
LukasStordeur: ok

AronH99:
So what did i do in this last two commits:
1. Addressed all of your comments, all of them should be handled properly now
2. Normally i fixed your pagination bug (this is to be tested again on your environment)
3. Fixed the bug that bart came accross in his testing during the meeting => the options list not popping up for the agents.
4. Added more testing specifically to the helpers of the getResources hook, I got help from Claude to generate these tests quickly, went over them and corrected some of them to fit. 
I think they would be very much useful since it's quite important that we pass the filters properly so yeah very helpful I reckon.
5. Added a comment + a ticket to review the input fields and their functionality again after release (tied to 6).
6. Added the fix where we can filter on substrings for resource type, value and agent (this is also documented with a todo and the ticket)
7. Reworked the AddableSelectInput to make use of pf's components properly

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
```diff
@@ -18,40 +30,143 @@ export interface ResourceFilterFormProps {
...
+
+    return data.pages.flatMap((page) =>
+      page.data.map((agent) => ({
+        value: agent.name,
+        label: agent.name,
+      }))
+    );
+  }, [data]);
+
+  const handlePurgedChange = (_e: unknown, hasChanged: boolean) => {
```
LukasStordeur: `Unknown`? Not `React.FormEvent<HTMLInputElement>`?
And, being pedantic, you're not consistent compared to the other component, where you only pass the used argument. Whichever is fine, but it's good to keep it consistent

LukasStordeur (changes requested):
Ask Claude to block your if statements and return statements :P

AronH99:
That's too much work for Claude I think 😆 
He is on a free subscription base only so no freeloaders who want blocked statements 🤣 
...
On a more serious note:
I am trying to make more use of blocked statements now, it's not something I can make my own over night but I'm definitely working on it

LukasStordeur (approved):

PR #6839, opened 2026-04-28
fix: cypress for resources view has been implemented again

AronH99:
# Description

- Cypress has been reworked and reactivated again for the resources view.
Some bad practices have been removed in favour of working with aria labels or testId's.


src/Slices/ServiceInstanceDetails/Test/mockServer.ts
```diff
@@ -47,7 +47,7 @@ const getHistoryLogsError = http.get("/lsm/v1/service_inventory/mobileCore/1d96a
 const getHistoryLogsDelayed = http.get(
   "/lsm/v1/service_inventory/mobileCore/1d96a1ab/log",
   async () => {
-    await delay(500);
+    await delay("infinite");
```
LukasStordeur, line 50: If this approach works to fix the flakes, shouldn't we apply it to other loading states? 
I'd also rename the name of the variables, to something clear that these calls will never resolve.

LukasStordeur (approved):

src/Slices/ServiceInstanceDetails/Test/mockServer.ts
AronH99: Well I decided to fix this specific issue for now since I saw this one coming back a lot on my pr's.
And with this fix I haven't yet seen it fail. 
I'll adjust the name of this one to make it more clear and we'll see later if any more "loading" issues pop up.
Because it's the only one which needs this fix normally

PR #6846, opened 2026-04-30
feat: added sorting for resources page

### LINKED ISSUE #6505: Add Compound Sorting for Status
Allow the table to be sorted on the different states

### LINKED ISSUE #6817: usePaginatedTable should have stable references between renders
The useUrlStatexxx (filter, sort,...) inside the usePaginatedTable should have stable references in between renders so that we don't need to use methods like JSON.stringify to keep unintended behaviour from happening.
=> also take a look at the useEffect inside of the usePaginatedTable to have a minimum working fix for all pages with paginations.

The unintended behaviour whenever we don't utilize JSON.stringify right now:
Whenever going to next or previous page on the resource view for example 
=> the useEffect inside of usePaginatedTable triggers because the references of empty {} filters are not stable.
So in short this should be given some thought before coming up and implementing a good solution.

AronH99:
# Description

So what happened in this pr:

1. Added multi sort support  to the usePaginatedTable hook. 
It is now possible to use this hook for both multi sorting as well as single sorting (added backwards compatibility because the single sorting is used on other pages/views throughout the web-console).
I also had to rework some logic so I took the liberty to fix the stable references issue as well. 
This was a little bit of a refactor because the useUrlState as base hook now has stable references and it chained into all the other hooks surrounding that but it should now be working and stable.

2. Using the usePaginatedTable hook to implement the new multi sorting functionality on the resources page. 
I added Patternfly drag and drop package, I kind of like how it is now though it could be a little better imo (mostly because you can only drag  the item by the drag icon but honestly in this case that might the best ux solution since u can click the rest of the item to change sorting direction).
Added a StatusSortMenu component where it is possible to:
 - Enable/disable specific status sort keys
 - Reorder active status sorts via drag-and-drop
 - Toggle sort direction (ascending/descending) by clicking an item
 - Has a tooltip for some minor explanation about the sorting.
 So in short all functionalities as requested should be working (see screenshot).
 
 3. Some e2e adjustments for new sorting and some unit test adjustments off course.

closes https://github.com/inmanta/web-console/issues/6505
closes https://github.com/inmanta/web-console/issues/6817

Screenshot 2026-04-30 at 10 16 29


cypress/e2e/scenario-6-resources.cy.js
```diff
@@ -407,6 +406,92 @@ describe("Scenario 6 : Resources", () => {
...
+
+    it("6.8 Resource sorting", () => {
+      cy.visit("/console/");
+      cy.get('[aria-label="Select-environment-test"]').click();
+      cy.get('[aria-label="Sidebar-Navigation-Item"]').contains("Resources").click();
+      cy.get('[aria-label="ResourcesPage-Success"]').should("be.visible");
+
+      // --- Regular column sorting ---
+
+      // Default sort is resource_type asc — clicking it should switch to desc
```
LukasStordeur, line 418: Shouldn't the default be no sort at all?

cypress/e2e/scenario-6-resources.cy.js
```diff
@@ -407,6 +406,92 @@ describe("Scenario 6 : Resources", () => {
...
+      cy.get('[data-testid="status-sort-active-list"]')
+        .contains('[data-testid="status-sort-item"]', "Blocked")
+        .click();
+      cy.get('[data-testid="status-sort-badge"]').should("have.text", "1");
+      cy.get('[data-testid="status-sort-active-list"]').within(() => {
+        cy.get('[data-testid="status-sort-item"]').should("have.length", 1);
+      });
+    });
   } else {
     it("6.2 Resources for OSS", () => {
```
LukasStordeur, line 551: The oss suit also could benefit an assertion or two about the sorting. It doesn't need to be as extensive of course.

src/Core/Domain/Resource/Resource.ts
```diff
@@ -218,10 +208,11 @@ export enum FilterKind {
   Status = "Status",
 }
 
-export type SortKey = "agent" | "status" | "resource_type" | "resource_id_value";
+export type StatusSortKey = "compliance" | "lastHandlerRun" | "blocked" | "isDeploying";
+export type SortKey = "agent" | "resource_type" | "resource_id_value" | StatusSortKey;
 
 export type FilterFromVersion = Omit<Filter, "status">;
-export type SortKeyFromVersion = Exclude<SortKey, "status">;
+export type SortKeyFromVersion = Exclude<SortKey, StatusSortKey>;
```
LukasStordeur, line 225: This one is used on the desired state details page, not the main resource page. That page is as far I know, still using the old endpoint, which doesn't support the compound sorting

src/Core/Domain/Sort.ts
```diff
@@ -22,3 +22,69 @@ export const parse = <Key extends string = string>(value: unknown): Sort<Key> |
...
+ * @example
+ * [{ name: "status", order: "asc" }, { name: "createdAt", order: "desc" }]
+ */
+export type MultiSort<Key extends string = string> = Sort<Key>[];
+
+/**
+ * Checks deep equality between two MultiSort arrays.
+ * Order of entries matters — [a, b] is not equal to [b, a].
+ */
+export const equalsMulti = <Key extends string>(a: MultiSort<Key>, b: MultiSort<Key>): boolean =>
```
LukasStordeur: Can you block the return statements of the various methods?

src/Core/Domain/Sort.ts
```diff
@@ -22,3 +22,69 @@ export const parse = <Key extends string = string>(value: unknown): Sort<Key> |
...
+ */
+export const toggleMulti = <Key extends string>(
+  sorts: MultiSort<Key>,
+  name: Key
+): MultiSort<Key> => {
+  const existing = sorts.find((s) => s.name === name);
+
+  if (!existing) return [...sorts, { name, order: "asc" }];
+  if (existing.order === "asc")
+    return sorts.map((s) => (s.name === name ? { ...s, order: "desc" as Order } : s));
```
LukasStordeur: `as Order` shouldn't be needed here no?

src/Core/Domain/Sort.ts
```diff
@@ -22,3 +22,69 @@ export const parse = <Key extends string = string>(value: unknown): Sort<Key> |
...
+ */
+export type MultiSort<Key extends string = string> = Sort<Key>[];
+
+/**
+ * Checks deep equality between two MultiSort arrays.
+ * Order of entries matters — [a, b] is not equal to [b, a].
+ */
+export const equalsMulti = <Key extends string>(a: MultiSort<Key>, b: MultiSort<Key>): boolean =>
+  a.length === b.length &&
+  a.every((item, i) => item.name === b[i].name && item.order === b[i].order);
```
LukasStordeur: isn't it easier to simply stringify both arrays and compare the string result? I think that would be more efficient too than a map

src/Core/Domain/Sort.ts
LukasStordeur, line 1: Each of the useUrlWithxxxx declare the custom parsing/serialization in their respective files. I'd prefer keeping it consistent if you don't mind, and move the multi-sort specific methods to useUrlWithMultiSort

src/Data/Common/usePaginatedTable.ts
```diff
@@ -1,49 +1,141 @@
...
 
-interface UsePaginatedTableOptions<TFilter = undefined, TSort extends string = string> {
+interface BaseOptions<TFilter> {
   route: RouteKind;
-  defaultSort?: Sort.Type<TSort>;
   defaultFilter?: TFilter;
   filterKeys?: Record<string, "IntRange" | "DateRange" | "Boolean">;
 }
 
+interface SingleSortOptions<TFilter, TSort extends string> extends BaseOptions<TFilter> {
```
LukasStordeur: Those two interfaces don't make sense to me. 
I'd rather opt for a single interface `SortOptions`
with a `mode: "single" | "multi"` 

Using boolean literals, for an optional property, and then having a second interface to have the opposite, is convoluted

src/Data/Common/usePaginatedTable.ts
```diff
@@ -1,49 +1,141 @@
...
+  setCurrentPage: (currentPage: CurrentPage) => void;
+  pageSize: PageSize.Type;
+  setPageSize: (size: PageSize.Type) => void;
+  filter: TFilter;
+  setFilter: (filter: TFilter) => void;
+  sort: Sort.MultiSort<TSort>;
+  setSort: (sort: Sort.MultiSort<TSort>) => void;
+};
+
+export function usePaginatedTable<TFilter = undefined, TSort extends string = string>(
```
LukasStordeur: Euhm, I do understand the idea behind it. You're trying to use function overloads, but this is really not the way to go for something like this. It's overly complicated, and while technically working, not practical at all

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+ * simple selectable menu items.
+ *
+ * @Props {Props} - Component props.
+ *  @prop {Sort.MultiSort<Resource.SortKey>} sort - Current multi-sort state containing all active and inactive sort definitions.
+ *  @prop {(sort: Sort.MultiSort<Resource.SortKey>) => void} setSort - Callback to update the sort state.
+ *
+ * @returns {React.ReactElement} The rendered status sort dropdown menu.
+ */
+export const StatusSortMenu: React.FC<Props> = ({ sort, setSort }) => {
+  //todo check overlayProps
```
LukasStordeur: Is this a leftover todo?

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  const activeCount = activeStatusSorts.length;
+
+  const onToggle = (key: Resource.StatusSortKey) => {
+    setSort(Sort.toggleMulti(sort, key));
+  };
+
+  const onDragFinish = (_event: DragEndEvent, newOrder: DraggableObject[]) => {
+    const reorderedStatusSorts = newOrder.map(
+      (item) => activeStatusSorts.find((s) => s.name === item.id)!
```
LukasStordeur: The non-null assertion `!` is risky here. If `item.id` doesn't match any active sort entry (e.g. stale drag state), `find` returns `undefined` and the assertion silently produces an invalid entry in the array. Consider filtering out the `undefined` cases or throwing an explicit error instead. You can apply a `filter(Boolean)` for example to eliminate undefined/null/empty strings in one pass.

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+  key: Resource.StatusSortKey;
+}
+
+const STATUS_SORT_ITEMS: StatusSortItem[] = [
+  { label: "Blocked", key: "blocked" },
+  { label: "Compliance", key: "compliance" },
+  { label: "Last Handler Run", key: "lastHandlerRun" },
+  { label: "Is Deploying", key: "isDeploying" },
+];
+export const STATUS_SORT_KEYS = STATUS_SORT_ITEMS.map((i) => i.key);
```
LukasStordeur: `STATUS_SORT_KEYS` is exported from a UI component but consumed in `ResourcesTable.tsx` and potentially other non-UI layers. Since it is the runtime counterpart to `Resource.StatusSortKey` (a domain type), it would fit better in `src/Core/Domain/Resource/Resource.ts` alongside the type itself, keeping the UI component free of domain constants.

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+            onClick={() => setIsOpen((prev) => !prev)}
+            aria-label="Sort by status fields"
+            aria-expanded={isOpen}
+            style={{
+              fontWeight: 600,
+              fontSize: "inherit",
+              position: "absolute",
+              bottom: 0,
+              right: 22,
+            }}
```
LukasStordeur: The project convention is to avoid inline styles and use `styled-components` or PatternFly utilities instead. What's the reason you needed to set the position absolute and right: 22?

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.css
```diff
@@ -0,0 +1,24 @@
+/* Targets the main wrapper of the dnd we otherwise wouldn't have access to. */
+.statusSortDragWrapper > div > div {
```
LukasStordeur: Selecting `.statusSortDragWrapper > div > div` relies on `@patternfly/react-drag-drop`'s internal DOM structure. Any patch to that package can silently break this styling. Is there a `className` prop or a CSS custom property exposed by `DragDropSort` that would give a more stable hook?

src/Data/Common/UrlState/useUrlStateWithMultiSort.ts
```diff
@@ -0,0 +1,36 @@
...
+import { provide, Location, Replace, StateConfig, Update } from "./helpers";
+import { useUrlStateHandler } from "./useUrlState";
+
+/**
+ * Synchronizes a {@link Sort.MultiSort} value with the URL query string.
+ *
+ * Reads and writes multiple sort criteria under the `sort` key as a
+ * comma-separated string (e.g. `?sort=status.asc,createdAt.desc`).
+ *
+ * Built on top of {@link handleUrlState} using the MultiSort
```
LukasStordeur: Stale JSDoc link — `{@link handleUrlState}` no longer exists; it was renamed to `useUrlStateHandler`. Same for `{@link handleUrlStateWithSort}` on line 13 → `useUrlStateWithSortHandler`.

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+        )}
+        popperProps={{ position: "center" }}
+      >
+        <DropdownList
+          style={{ display: "flex", flexDirection: "column", gap: "6px", width: "220px" }}
+          data-testid="status-sort-menu"
+        >
+          {/* Active sorts — draggable to reorder */}
+          {activeStatusSorts.length > 0 && (
+            <div className="statusSortDragWrapper" data-testid="status-sort-active-list">
```
LukasStordeur: Don't use plain html, always favor using patternfly components or elements.

src/Slices/Resource/UI/ResourcesPage/ResourcesTable.tsx
```diff
@@ -12,43 +13,55 @@ import {
...
 
 export const ResourcesTable: React.FC<Props> = ({ rows, sort, setSort, ...props }) => {
   const onSort: OnSort = (_event, index, order) => {
-    setSort({
-      name: getColumnNameForIndex(index) as Resource.SortKey,
-      order,
-    });
+    const name = getColumnNameForIndex(index) as Resource.SortKey;
+
+    // Whenever sorting via regular column, we want to keep the status sorts intact
```
LukasStordeur: Do we?

src/Data/Common/usePaginatedTable.ts
```diff
@@ -1,49 +1,141 @@
...
     route,
     keys: filterKeys,
     default: defaultFilter,
   });
-  const [sort, setSort] = useUrlStateWithSort<TSort>({
-    default: defaultSort ?? { name: "" as TSort, order: "asc" },
+
+  const [singleSort, setSingleSort] = useUrlStateWithSort<TSort>({
+    default: (!isMulti ? options.defaultSort : undefined) ?? {
+      name: "" as TSort,
```
LukasStordeur: The `name: "" as TSort` cast is a type-safety hole — it forces an empty string through a generic `extends string` constraint. Any downstream code that uses `sort.name` without first checking for the empty-string sentinel (e.g. `getIndexForColumnName`) could produce silent bugs.

src/Data/Common/usePaginatedTable.ts
```diff
@@ -1,49 +1,141 @@
...
  */
-export function usePaginatedTable<TFilter = undefined, TSort extends string = string>({
-  route,
-  defaultSort,
-  defaultFilter,
-  filterKeys,
-}: UsePaginatedTableOptions<TFilter, TSort>) {
-  //TODO: The useUrlStatexxx should have stable references i between renders so that we don't need to use methods like JSON.stringify
-  //https://github.com/inmanta/web-console/issues/6817
+export function usePaginatedTable<TFilter = undefined, TSort extends string = string>(
```
LukasStordeur: The overload + dual-hook + `isMulti` branching pattern makes this significantly harder to follow than it needs to be. I'd suggest a different approach: **leave `usePaginatedTable` completely unchanged** and add a dedicated `usePaginatedTableWithMultiSort` hook alongside it.

The two hooks are nearly identical — the only difference is swapping `useUrlStateWithSort` for `useUrlStateWithMultiSort`. The minor duplication is completely fine for hooks this small, and you avoid every complication introduced here: no overloads, no parallel hooks on the same URL key, no `isMulti` runtime flag, no `BaseOptions`/`SingleSortOptions`/`MultiSortOptions` dance, and no `"" as TSort` sentinel.

```ts
// usePaginatedTable.ts — unchanged

// usePaginatedTableWithMultiSort.ts — new file
export function usePaginatedTableWithMultiSort<
  TFilter = undefined,
  TSort extends string = string
>({
  route,
  defaultSort = [],
  defaultFilter,
  filterKeys,
}: {
  route: RouteKind;
  defaultSort?: Sort.MultiSort<TSort>;
  defaultFilter?: TFilter;
  filterKeys?: Record<string, "IntRange" | "DateRange" | "Boolean">;
}) {
  const [currentPage, setCurrentPage] = useUrlStateWithCurrentPage({ route });
  const [pageSize, setPageSize] = useUrlStateWithPageSize({ route });
  const [filter, setFilter] = useUrlStateWithFilter<TFilter>({
    route,
    keys: filterKeys,
    default: defaultFilter,
  });
  const [sort, setSort] = useUrlStateWithMultiSort<TSort>({ default: defaultSort, route });

  useEffect(() => {
    setCurrentPage({ kind: "CurrentPage", value: "" });
    // setCurrentPage is stable — omitted intentionally to avoid an infinite loop
  }, [sort, filter]);

  return { currentPage, setCurrentPage, pageSize, setPageSize, filter, setFilter, sort, setSort };
}
```

`Page.tsx` then just calls `usePaginatedTableWithMultiSort` instead of `usePaginatedTable`.

LukasStordeur (changes requested):

cypress/e2e/scenario-6-resources.cy.js
AronH99: It was the default as well whenever i started this but we can remove it if u'd like? not sure why it was there in the first place? @LukasStordeur

src/Slices/Resource/UI/ResourcesPage/ResourcesTable.tsx
AronH99: it would make most sense i guess no?
Why would i need to support multi sorting otherwise?
If you would click on type, agent, value sorting + new compound status sorting..

It's a possibility so why not make it available in this way?

src/Core/Domain/Sort.ts
AronH99: I looked into this and apparently using every should be more efficient since it will short-circuit on the first mismatch it finds, whilest stringify needs to walk to the entire array and objects and can only compare in the end..

I also looked at the equals function which was as well in this file and this was essentially the extension of that.

cypress/e2e/scenario-6-resources.cy.js
LukasStordeur: Leave it for now, we can discuss this later to finetune it if needed

src/Core/Domain/Sort.ts
LukasStordeur: But if you make a string out of the array, you don't need to loop/map over it 🤔

src/Slices/Resource/UI/ResourcesPage/ResourcesTable.tsx
LukasStordeur: No, it doesn't make sense, how do I set priority then? Which sort is taken in account first? The multisort on the status is controled, you can't say the same on the order of the others. Which will return inconsistent results

LukasStordeur: I'd stick to one column at the time, which can happen to have a multi sort like the status column

src/Core/Domain/Sort.ts
AronH99: no but apparently it's more efficient because with stringifying u also need to visit every element/property to then allocate more memory to it because u are making new strings out of the arrays.
In the worst case the every does the same traversal but with no allocations and in most cases it's faster because it will return early.

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
AronH99: the reason for that is because otherwise visually this:

Screenshot 2026-05-07 at 10 40 35

otherwise won't line up, i tried some other things as well but visually nothing really worked but this

LukasStordeur: https://www.patternfly.org/components/menus/menu-toggle#with-a-badge

```tsx
  <Fragment>
    <MenuToggle badge={<Badge>{counter}</Badge>}>Count</MenuToggle>
    <MenuToggle variant="plainText" badge={<Badge screenReaderText="additional items">{counter}</Badge>} />
  </Fragment>
```

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.css
AronH99: i can make the selector into:
.statusSortDragWrapper div:has(> [aria-roledescription="sortable"]) 

but i am afraid that there are no relevant selectors i can use for patternfly's wrapping div...

Screenshot 2026-05-07 at 10 57 31

AronH99: it's not possible to add classNames no, i tried..

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
AronH99: i know but it will still look bad :/
look:

Screenshot 2026-05-07 at 11 04 02

AronH99: the type, agent value are glued to the bottom of their headers but anything else has more padding or more height and it will stand above the rest visually..

AronH99: also the font is different aswell so without custom styling i cant match it to the rest of the columns

AronH99: its honestly what i am finding myself having the most issues on with working with patternfly, it's just really hard to mix and match sometimes

LukasStordeur: is it the same if separate the text from the toggle?

AronH99: if i remove the toggle and go with the "plain" variant its still the same yes:

Screenshot 2026-05-07 at 11 27 45

AronH99: it's for sure less code which is nice but unfortunately it doesnt really fit :/

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.css
LukasStordeur: Why is the custom css needed in the first place?

src/Data/Common/UrlState/useUrlStateWithMultiSort.ts
AronH99: nice catch

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.css
AronH99: Screenshot 2026-05-07 at 11 34 46

unfortunately all of the drag and drop variants are not working very well, so i took the best variant and made minimal css changes to it to come to:

Screenshot 2026-05-07 at 11 36 12

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
LukasStordeur: But, if the dropdown is a different element than the text in the header, it shouldn't affect the text? Can you commit the version with the changes? 
Custom css should be prevented as a general rule, I'm not eager to let that pass. Same for plain html, it hints something is not used properly.

AronH99:
So what happened now:
1. Addressed all comments, everything should be fixed now on that regard.
2. I honestly reworked the StatusSortMenu, now everything should be semantically perfect as the docs of Patternfly state.
I also decided to drop drag and drop because I came across another interesting solution which you can see on the screenshot.
I used Patternfly's inherit MenuItemActions to be able to order the menu items which seems more intuitive and honestly works slightly better in my opinion than the drag and drop system (and don't need the package either which is a nice bonus).
Let me know what u think.
3. Adjusted e2e test.

Ps: going to make another component unit test for my StatusSortMenu tomorrow but u can review it now already.

Screenshot 2026-05-07 at 16 27 27

src/Slices/DesiredStateDetails/UI/VersionResourceTable.tsx
```diff
@@ -8,8 +8,8 @@ interface Props {
   version: string;
   rows: RowFromVersion[];
   tablePresenter: VersionResourceTablePresenter;
-  sort: Sort.Type<Resource.SortKeyFromVersion>;
-  setSort: (sort: Sort.Type<Resource.SortKeyFromVersion>) => void;
+  sort: Sort.Type<Resource.DesiredStateSortKey>;
+  setSort: (sort: Sort.Type<Resource.DesiredStateSortKey>) => void;
```
LukasStordeur: It's not exactly correct. It's still a resource table, but for a specific desired state version

src/Core/Domain/Resource/Resource.ts
```diff
@@ -218,10 +208,24 @@ export enum FilterKind {
...
 
 export type FilterFromVersion = Omit<Filter, "status">;
-export type SortKeyFromVersion = Exclude<SortKey, "status">;
+
+interface StatusSortItem {
+  label: string;
+  key: StatusSortKey;
+}
+
+export const STATUS_SORT_ITEMS: StatusSortItem[] = [
```
LukasStordeur: Those are displayed values, and belong in the words.tsx. You can easily skip all the filtering too then.
```ts
words(`resources.sort.label.${key}`);
```

LukasStordeur (changes requested):
I'll review the rest when I get to the office :)

src/Core/Domain/Resource/Resource.ts
```diff
@@ -218,10 +208,24 @@ export enum FilterKind {
...
+const STATUS_SORT_KEY_SET: ReadonlySet<string> = new Set(STATUS_SORT_KEYS);
+
+/**
+ * Type guard that checks whether a given SortKey is a StatusSortKey.
+ * Used to safely narrow sort keys to status-related keys.
+ */
+export const isStatusSortKey = (key: SortKey): key is StatusSortKey => STATUS_SORT_KEY_SET.has(key);
+
+/** @deprecated Use Sortkey instead, only in use for desired state resource table */
+export type OldSortKey = "agent" | "resource_type" | "resource_id_value";
```
LukasStordeur: I'm not certain versioned resources will have the same keys as the new compound status. I wouldn't deprecate those just yet and keep the "Version" key word in the name.

src/Slices/DesiredStateDetails/UI/Page.tsx
```diff
@@ -31,7 +31,7 @@ export const Provider: React.FC = () => {
 export const Page: React.FC<{ version: string }> = ({ version }) => {
   const [isDrawerExpanded, setIsDrawerExpanded] = useState(false);
   const { currentPage, setCurrentPage, pageSize, setPageSize, filter, sort, setSort } =
-    usePaginatedTable<Resource.FilterFromVersion, Resource.SortKeyFromVersion>({
+    usePaginatedTable<Resource.FilterFromVersion, Resource.OldSortKey>({
```
LukasStordeur: The "FromVersion" is used in more places. I would revert this for now, and only focus on the scoped changes for the main resource page.

LukasStordeur (changes requested):

AronH99:
So as discussed in the office I think i have come to the most elegant solution possible right now, I will also keep u posted about the Patternfly bug report (the visual glitch).
Video:


https://github.com/user-attachments/assets/586d4246-13de-4705-ac65-4719c71ba708

cypress/e2e/scenario-6-resources.cy.js
```diff
@@ -1,4 +1,5 @@
 import environmentHelpers from "../support/environmentHelpers";
+import "cypress-real-events/support";
```
LukasStordeur: Why do you need this? This library is only supporting Chromium based browsers. 
Which event can't be properly handled by Cypress? We are already using drag and drop events for the composer component.

src/Core/Domain/Resource/Resource.ts
```diff
@@ -218,10 +208,23 @@ export enum FilterKind {
...
+
+const STATUS_SORT_KEY_SET: ReadonlySet<string> = new Set(STATUS_SORT_KEYS);
+
+/**
+ * Type guard that checks whether a given SortKey is a StatusSortKey.
+ * Used to safely narrow sort keys to status-related keys.
+ */
+export const isStatusSortKey = (key: SortKey): key is StatusSortKey => STATUS_SORT_KEY_SET.has(key);
+
+export type SortKeyFromVersion = "agent" | "resource_type" | "resource_id_value";
```
LukasStordeur: Why not keep it with the `Exclude` util?
Keeping one source of truth is always better than redundancy.

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+describe("StatusSortMenu", () => {
+  describe("active sort badge count", () => {
+    it("shows 0 when no active sorts", () => {
+      const { badge } = setup();
+      expect(badge()).toHaveTextContent("0");
+    });
+
+    it("reflects the number of active status sorts", () => {
+      const { badge } = setup(
+        makeSort([
```
LukasStordeur: What's the added value of having this method, beyond the typecheck? All it does is return the array you passed in 🤔 
It's not concistently used either, so I'm unsure if it is intentional, or something left over that should be removed?

src/Slices/Resource/UI/ResourcesPage/ResourcesTable.tsx
```diff
@@ -12,43 +14,46 @@ import {
...
-  const smallHeaders = ["status"];
+
+  const activeRegularSort = sort.find(
+    (sortEntry) => !Resource.STATUS_SORT_KEYS.includes(sortEntry.name as Resource.StatusSortKey)
+  );
+
   const heads = columnHeads.map(({ apiName, displayName }, columnIndex) => {
+    if (apiName === "status") {
+      return (
+        <Th style={{ textAlign: "end", overflow: "visible" }} key={displayName}>
```
LukasStordeur, line 32: Why overflow visible?

src/Slices/Resource/UI/ResourcesPage/ResourcesTable.tsx
```diff
@@ -12,43 +14,46 @@ import {
...
-      order,
-    });
+    const name = getColumnNameForIndex(index) as Resource.SortKey;
+    setSort([{ name, order }]);
   };
-  const activeSortIndex = getIndexForColumnName(sort.name);
-  const smallHeaders = ["status"];
+
+  const activeRegularSort = sort.find(
+    (sortEntry) => !Resource.STATUS_SORT_KEYS.includes(sortEntry.name as Resource.StatusSortKey)
```
LukasStordeur: Claude: 
The `as Resource.StatusSortKey` cast here is semantically incorrect — you're casting a value that might not be a `StatusSortKey` in order to pass it to `includes`, which silences TypeScript on the unsafe coercion. The domain already exports `isStatusSortKey` exactly for this purpose:

```ts
const activeRegularSort = sort.find(
  (sortEntry) => !Resource.isStatusSortKey(sortEntry.name)
);
```

This is type-safe and more readable — no cast required.

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+                  handleToggle(key);
+                }}
+              >
+                <Icon status="info" isInline>
+                  {activeEntry?.order === "desc" ? <SortAmountDownIcon /> : <SortAmountUpIcon />}
+                </Icon>
+              </DataListCell>
+            ),
+          ]}
+        />
```
LukasStordeur: Claude:
`isActive && (<DataListCell .../>)` evaluates to `false` when `isActive` is `false`, so `dataListCells` receives `Array<ReactElement | false>`. PatternFly types `dataListCells` as `React.ReactNode[]` — `false` is a valid `ReactNode` in React 18 but it is not a `ReactElement`, so depending on the exact PF overload this may produce a TypeScript error or a runtime warning.

Prefer the conditional spread to keep the array type clean:

```tsx
dataListCells={[
  <DataListCell key={`label-${key}`} ...>...</DataListCell>,
  ...(isActive
    ? [<DataListCell key={`sort-icon-${key}`} ...>...</DataListCell>]
    : []),
]}
```

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+    setSort(toggleMulti(sortStatus, key));
+  };
+
+  const onDrop = (_: unknown, newItems: DraggableObject[]) => {
+    const newOrder = newItems.map((item) => item.id as Resource.StatusSortKey);
+    setSort(reorderSorts(newOrder, sortStatus));
+  };
+
+  const items = itemOrder.map((key) => {
+    const activeEntry = sortStatus.find((entry) => entry.name === key);
```
LukasStordeur: Claude:
`sortStatus` is correctly memoized, but every derived value that flows from it — `activeSortStatus`, `inactiveSortStatus`, `itemOrder`, and especially `items` (which creates new React elements each time) — is recomputed on every render. `DragDropSort` will receive a new `items` array reference on every render even when the sort state hasn't changed, which may cause unnecessary internal diffing or re-renders inside the PF drag-drop library.

`items` in particular should be wrapped in `useMemo` with `[sortStatus]` as the dependency (since it's the only thing that drives the computation). The other derived arrays are cheap but could follow the same pattern for consistency.

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+        overlayProps={{ isCompact: true }}
+      >
+        <DataList aria-label="draggable-sort-items" isCompact />
+      </DragDropSort>
+    </Menu>
+  );
+
+  return (
+    <Tooltip content={words("resources.column.status.toolTip")}>
+      <MenuContainer
```
LukasStordeur, line 191: Claude:
Wrapping `MenuContainer` with `Tooltip` is risky for a few reasons:

1. **The tooltip fires while the menu is open.** Once the user opens the menu, hovering over the toggle button still pops the tooltip on top of the menu, which is distracting and probably not the intended behaviour.
2. **Focus management conflict.** PF's `MenuContainer` manages its own focus trap and keyboard navigation. `Tooltip` adds its own layer of ARIA attributes (`aria-describedby`) that can interfere with the menu's accessibility tree.

A common alternative is to put the tooltip only on a static info icon next to the toggle (e.g. `QuestionCircleIcon`) rather than wrapping the interactive control, which sidesteps both issues.

Me:
None of the table headers that have sorting have a tooltip.

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  const activeSortStatus = sortStatus.map((entry) => entry.name);
+  const inactiveSortStatus = Resource.STATUS_SORT_KEYS.filter(
+    (key) => !activeSortStatus.includes(key)
+  );
+  const itemOrder = [...activeSortStatus, ...inactiveSortStatus];
+
+  const handleToggle = (key: Resource.StatusSortKey) => {
+    setSort(toggleMulti(sortStatus, key));
+  };
```
LukasStordeur: Claude:
`toggleMulti(sortStatus, key)` operates on `sortStatus`, which contains only the status-related entries of `sort`. So `setSort(...)` replaces the entire multi-sort with a status-only array, silently discarding any active regular column sort (e.g. "Agent asc"). The same applies to `onDrop`.

This is intentional per the design discussion ("one column at a time, which can happen to have a multi-sort like the status column"), but the contract is completely implicit. A future reader or author touching this code could easily introduce a bug thinking the two sort categories are additive.

At minimum, a short comment explaining the mutual exclusivity would prevent regressions. Alternatively, the type signature itself could enforce it — e.g. a discriminated union `{ kind: "regular"; sort: Sort.Type<...> } | { kind: "status"; sorts: MultiSort<StatusSortKey> }` would make the invariant impossible to violate by accident.

cypress/e2e/scenario-6-resources.cy.js
LukasStordeur, line 1: Claude:
Two observations on the drag-and-drop e2e test in "6.8 Resource sorting":

**`cy.wait(1000)` after the drop** — Hard waits are brittle; CI machines run at variable speeds and 1000 ms can be either too short (flaky on slow runners) or needlessly slow. The drop already has an observable side effect: the `data-testid` attributes on the list items change order. Replace the fixed wait with an assertion on the new order that Cypress will retry automatically:
```js
// Instead of cy.wait(1000)
cy.get('[data-testid^="status-sort-item-"]').first()
  .should("have.attr", "data-testid", "status-sort-item-compliance-active");
```

**Arbitrary pixel offsets in `realMouseMove`** — `realMouseMove(0, -5)` and `realMouseMove(0, 5)` are magic numbers that happen to work at a given viewport size but are not tied to the actual item height. If the list item height ever changes (e.g. PF upgrade), the drag might no longer land on the intended target. Consider using the bounding box of the target element to compute the delta instead.

LukasStordeur (changes requested):
Claude:

**Summary of new comments:**

- `ResourcesTable.tsx` line 28 — `as Resource.StatusSortKey` is a semantically wrong cast; `isStatusSortKey` already exists for this purpose
- `StatusSortMenu.tsx` lines 106–132 — `isActive && <DataListCell/>` puts `false` into the `dataListCells` array; conditional spread is cleaner and type-safe
- `StatusSortMenu.tsx` line 94 — `items` (and its upstream derived arrays) are recomputed every render; should be `useMemo`-ised
- `StatusSortMenu.tsx` line 86 — `handleToggle`/`onDrop` silently discard any active non-status column sort; the mutual exclusivity is entirely implicit and a future author will not know it's intentional
- `StatusSortMenu.tsx` line 170 — `Tooltip` wrapping `MenuContainer` fires the tooltip while the menu is open and may interfere with PF's focus trap
- `cypress/e2e/scenario-6-resources.cy.js` — `cy.wait(1000)` hard wait after drag-and-drop is flaky; the pixel offsets in `realMouseMove` are fragile magic numbers

LukasStordeur (commented):
Additionaly:
1. No unit tests for usePaginatedTableWithMultiSort
2. Grayed out items are hinting at disabled states. I wouldn't use that.

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.test.tsx
AronH99: yh forgot to remove indeed thx

src/Slices/Resource/UI/ResourcesPage/ResourcesTable.tsx
AronH99: Otherwise the tableHeaderCell will hide the entire StatusSortMenu, because that's it's default

cypress/e2e/scenario-6-resources.cy.js
```diff
@@ -476,36 +475,62 @@ describe("Scenario 6 : Resources", () => {
...
+
+          cy.wrap($buttons[1])
+            .trigger("pointerdown", {
+              force: true,
+              isPrimary: true,
+              button: 0,
+              clientX: startX,
+              clientY: startY,
+            })
+            .wait(1000)
```
AronH99: the wait is needed here for stability, recommended on a dnd thread from 2024

LukasStordeur: But, it remains arbitrary, while asserting if the expected result is there is not. 
Can you reference the thread? 2024 is already a bit outdated.

AronH99: https://github.com/clauderic/dnd-kit/issues/208

Here is where I read about it since Patternfly uses that one under the hood.
Seems like quite a few people were having those issues and still having those issues with it...

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
```diff
@@ -0,0 +1,198 @@
...
+  // status sorts intentionally clears any active regular-column sort.
+  const handleToggle = useCallback(
+    (key: Resource.StatusSortKey) => {
+      setSort(toggleMulti(sortStatus, key));
+    },
+    [setSort, sortStatus]
+  );
+
+  const onDrop = useCallback(
+    (_: unknown, newItems: DraggableObject[]) => {
```
LukasStordeur: Isn't it an event? Is it really unknown?

AronH99: no i just noticed now that i can import it under a different name as type, though it will remain unused

cypress/e2e/scenario-6-resources.cy.js
LukasStordeur: I think they're talking about the arguments you need to pass the trigger event, not the extra waiting time. 
Did the test flake without the waiting time?

src/Slices/Resource/UI/ResourcesPage/Components/Sort/StatusSortMenu.tsx
LukasStordeur: yes, it being unused is fine, but the type should still remain correct :D

cypress/e2e/scenario-6-resources.cy.js
AronH99: Didn't test that until now, it seems like it isn't flaking locally so I'll push the changes and see if it flakes via Jenkins.

EDIT: checks have passed

changelogs/6505-resources-sorting.yml
```diff
@@ -0,0 +1,6 @@
+description: Added sorting on the resources page. It is now possible to sort by compound state and deployment status.
+issue-nr: 6505
+change-type: patch
```
LukasStordeur: can you make it a minor?

LukasStordeur (approved):
I approved, but wait until the patch release is out before merging :)

PR #6850, opened 2026-05-04
feat: users can now search for specific versions in the compare tab of a service instance attributes

### LINKED ISSUE #6143: Add search in compare versions instance details
Add the possibility to search for a specific version in the dropdowns for the versions in the Attributes/compare tab.

AronH99:
# Description

What changed?
1) Users can now search for specific versions in the compare tab of a service instance attributes (see ticket).
2) Made a new component SearchWithMenu to accommodate the requested changes.
@LukasStordeur Should I delocalize this component and put it in the overall UI components folder?
It's a really generic one but I am not sure if it will fit our needs for the future as well... You have any idea?
3) Adjusted unit tests and made unit tests for the new component. 

closes https://github.com/inmanta/web-console/issues/6143

Screenshot 2026-05-04 at 14 51 55


src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/SearchWithMenu.tsx
```diff
@@ -0,0 +1,147 @@
...
+  };
+
+  const handleClear = () => {
+    onChange("");
+    onClear?.();
+    suppressCloseRef.current = true;
+    setIsOpen(true);
+  };
+
+  const handleDocumentClick = useCallback((e: MouseEvent) => {
```
LukasStordeur: This is something Patternfly offers by default on all their dropdowns/menus, you don't need to redefine it.

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/SearchWithMenu.tsx
```diff
@@ -0,0 +1,147 @@
...
+              {option}
+            </MenuItem>
+          ))}
+        </MenuList>
+      </MenuContent>
+    </Menu>
+  );
+
+  return (
+    <Popper
```
LukasStordeur: Why work with a popper? 
https://www.patternfly.org/components/menus/select/react-templates/#typeahead
That component should cover most of what you need, no?

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/SearchWithMenu.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const suppressCloseRef = useRef(false);
+
+  const filter = useCallback(
+    (option: string, value: string): boolean => {
+      if (!value) return true;
+      if (typeof filterStrategy === "function") return filterStrategy(option, value);
+      const lower = value.toLowerCase();
+      if (filterStrategy === "all") return true;
+      if (filterStrategy === "includes") return option.toLowerCase().includes(lower);
+      return option.toLowerCase().startsWith(lower);
```
LukasStordeur: New lines + some blocked statements

LukasStordeur (changes requested):
Responding to the question about moving the component.
Most of the implementation is redundant and available in patternfly, so I'd first simplify your version to use that, and then move it to the shared folder

AronH99:
All right comments are addressed, necessary changes should have been made

src/UI/Components/SearchSelect/SearchSelect.test.tsx
LukasStordeur, line 1: Why do we need to test the inner implementation of the Patternfly component? I would expect the test to be blackboxed, and only assert on the result without mocking the inner methods. 

I'd rather expect the test then to type something ,and confirm the list has been updated. That will indirectly assert onChange has been called or that your list is actually filtering
Focussing the menu and selecting an item => indirectly asserts that the menu was opened, otherwise, you wouldn't be able to select an item.

src/UI/Components/SearchSelect/SearchSelect.test.tsx
```diff
@@ -0,0 +1,184 @@
+import React, { useState } from "react";
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, it, expect, vi } from "vitest";
```
LukasStordeur: You don't need to import from vitest, those are globally available

LukasStordeur (changes requested):
I'm not sure we need the complex filter strategy for a list of numbers to be filtererd 🤔 
Looks a bit overkill, and not worth the added maintenance. (Don't make something that might maybe used one day. That will either be forgotten, or not exactly what will be needed in the future, and will cost you efforts to maintain while it's not even used)
 
5 items max seems a bit weird. Isn't the list scrollable? Slicing it means that potential results are hidden and can confuse the user in thinking they don't have more items.

Also, what happens now when you don't fill in a version that doesn't exist and leave the input field? 

https://www.patternfly.org/components/menus/custom-menus#with-inline-search-filter
Patternfly suggests to have the searchfield within the menu, when it's a select component. (I'm sorry I didn't think of that one when reviewing last time, I was led off with the typeahead thing, but based on the usecase, the one with the search field in the menu seems better)

AronH99:
Screenshot 2026-05-22 at 11 33 30

Implemented your entire suggestion, was a good solution for this one indeed

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesCompare.tsx
```diff
@@ -93,101 +90,84 @@ export const AttributesCompare: React.FC<Props> = ({ instanceLogs, selectedVersi
...
+            </FlexItem>
+          </Flex>
+        </FlexItem>
       </Flex>
       <DiffEditor
         height={"calc(100vh - 525px)"}
         options={{
           readOnly: true,
           renderSideBySide: true,
+          scrollbar: { vertical: "hidden" },
```
LukasStordeur: don't hide scrollbars, I know it looks prettier, but it's not recommended for UX reasons.

src/UI/Components/SearchSelect/SearchSelect.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <MenuContent maxMenuHeight="200px">
+        <MenuList>{menuItems}</MenuList>
+      </MenuContent>
+    </Menu>
+  );
+
+  return (
+    <MenuContainer
+      menu={menu}
+      popperProps={{ placement: "bottom" }}
```
LukasStordeur: I notice you're often overriding the placement. But if you don't override it, it will adjust based on the available space. If you're at the bottom of a page, it will open it on top of the menuToggle. There's no need to override the placement.

LukasStordeur (changes requested):

LukasStordeur (approved):

PR #6854, opened 2026-05-06
improv: added dedicated json editor for dict fields

### LINKED ISSUE #6836: [REQUEST] Improve dict editor
### Summary

Service instances forms can offer an input to edit dicts. However, this is a single line without much assistance.

Image

### Problem / Use Case

Improve the UX

### Proposed Solution(s)

Create an input for dicts that either is a monaco instance with a formatted string (one item per line) (cfr the attributes in the resources) or a table with key / value paris

### Impact

Is there an impact expected on existing features/workflows/user-groups?

### Additional Context

Add any other context, sketches, links, or implementation notes that might be helpful for the frontend team.

AronH99:
# Description

What happened in this pr:
1. Made a dedicated JSON editor for the dict fields to have a DictFieldInput (see screenshot)
2. Made unit tests or made adjustments to existing ones.

I tried making minimal adjustments since I really felt like this whole form system was overly complex and could use a rework but maybe that's just me?
Let me know if I missed something?

closes https://github.com/inmanta/web-console/issues/6836

Screenshot 2026-05-06 at 11 03 19


AronH99:
Screenshot 2026-05-22 at 09 36 55
Screenshot 2026-05-22 at 09 45 17

It now looks like this, which honestly is quite clean imo, wasn't too much of a hassle either tbf.
@LukasStordeur u can review it now i reckon :) 
I think I have correctly applied everything but not 100% sure because of the complexity of the form.

src/UI/Components/ServiceInstanceForm/Components/DictFieldInput/DictFieldInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+    monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
+      validate: true,
+      schemaValidation: "ignore",
+    });
+
+    // Initial height
+    setHeight(Math.min(editor.getContentHeight(), 400));
+
+    // Resize height accordingly
+    editor.onDidContentSizeChange((e) => {
```
LukasStordeur, line 40: Should do it automatically, there's a height="sizeToFit"

src/UI/Components/ServiceInstanceForm/Components/DictFieldInput/DictFieldInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const [isInvalid, setIsInvalid] = useState(false);
+  const [height, setHeight] = useState(100);
+
+  const handleEditorDidMount: ComponentProps<typeof CodeEditor>["onEditorDidMount"] = (
+    editor,
+    monaco
+  ) => {
+    // Disable schema validation — without this Monaco rejects all
+    // properties because it has no schema to validate against.
+    monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
```
LukasStordeur: When passing the language={Language.json}
it will automatically do validation. You don't need to override the setDiagnosticsOptions at all.
You can test it here and replace the language to json. You'll see that the text will be highlighted as incorrect JSON syntax. 

https://www.patternfly.org/components/code-editor#with-sizetofit-height-height-will-growshrink-with-content

src/UI/Components/ServiceInstanceForm/Components/DictFieldInput/DictFieldInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+        const markers = monaco.editor.getModelMarkers({ resource: model.uri });
+        setIsInvalid(markers.some((m) => m.severity === monaco.MarkerSeverity.Error));
+      }
+    });
+  };
+
+  const handleChange = (val: string) => {
+    setText(val);
+    const parsed = toDict(val);
+    if (parsed !== null) {
```
LukasStordeur: You're not going to like this, but null is actually valid JSON, and is also allowed for the `dict?` types, so you will want to take those values into account :)

src/UI/Components/ServiceInstanceForm/Components/DictFieldInput/helpers.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Converts a JavaScript object into a formatted JSON string.
+ *
+ * If the input is not a valid object or is empty, returns an empty JSON object string ("{}").
+ *
+ * @param {unknown} value - The value to convert into a JSON string.
+ * @returns {string} A pretty-printed JSON string representation of the object.
+ */
+export function toText(value: unknown): string {
```
LukasStordeur, line 14: I would use a try catch instead of the if cases.

LukasStordeur (changes requested):
There are a few things.
1. `null` is a valid value. Right now, your editor will show `{}`  while in fact, the value is `null`. 
2. Dicts are usually received as normal objects. But not always. It can also be received as a whole stringified object. like this one: `'{"default":"value"}'` So it would be good to avoid regression and add a test to cover that. It's fine to send it back as a proper object, but the app should not crash trying to load such a value.
3. With the json input field, we don't need to convert optional values that are empty objects to `null`. We used to do that because the text fields do not have the concept of `null`. Now you have the issue that null will be displayed as an empty object, while being submitted as null. 
4. I'd also like to see test cases with other sorts of values. This is for example valid for the API. 
```json
{
  "a": [],
  "b": "e",
  "c": null,
  "d": 1
}
```

src/UI/Components/ServiceInstanceForm/Components/DictFieldInput/DictFieldInput.tsx
AronH99: I know and maybe thats where the use case of this field probably comes in.
If we don't do the custom height logic it will keep on scaling without setting a proper height limit.
Now if this is not an issue I can just remove it, however, if people use it for setting a lot of json this might be inconvenient because of not having a scrollbar in the json field...

LukasStordeur: Ok, then it would be good to add the reasoning in a comment.

AronH99:
All concerns should be addressed now:

Screenshot 2026-05-28 at 10 50 44

LukasStordeur (approved):

PR #6856, opened 2026-05-06
6837 composer blueprint

LukasStordeur:
# Description

- create placeholder connections with the minimal required amount of embedded entities. It was agreed on not placing placeholders of inter-service relations considering those often come from the existing inventory.
- Cherry picked a dependabot pr that was stuck in a loop. 

closes:
- #6837 

https://github.com/user-attachments/assets/801778e9-b06f-4919-931e-7e531ef32e20


AronH99 (approved):
Went through all code couldn't find anything inherently wrong, then started the environment and tested as well. Works perfectly so approved

PR #6862, opened 2026-05-07
Bugfix: Don't add optional embedded-entities in form regardless of nested fields optionality

LukasStordeur:
# Description 

The old check `!== null` let undefined through as truthy, making `showList = true` and rendering the optional entity's fields regardless of its children's optionality. It appeared to "work" for optional inner entities because those fields rendered empty/indeterminate. 

image

### DIFF
--- a/changelogs/unreleased/optional-embedded-entity-form.yml
+++ b/changelogs/unreleased/optional-embedded-entity-form.yml
@@ -1,0 +1,5 @@
+description: Bugfix for optional embedded entities in the service instance form, ensuring that they are not automatically added when they are optional, even when they contain required nested attributes.
+change-type: patch
+destination-branches: [master, iso9]
+sections:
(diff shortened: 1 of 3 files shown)


AronH99 (approved):
This reminds me of something in the same file:
get(originalState, makePath(path, field.name)) !== undefined)

is also used often to decide whether something should be disabled or not, but I can't tell for sure if this is intended or should catch nulls aswell

(not related to this ticket hence why I approved)

LukasStordeur:
> This reminds me of something in the same file: get(originalState, makePath(path, field.name)) !== undefined)
> 
> is also used often to decide whether something should be disabled or not, but I can't tell for sure if this is intended or should catch nulls aswell
> 
> (not related to this ticket hence why I approved)

There's often a distinction between null and undefined, where null is ok, but undefined not.

Issue #6877, opened 2026-05-18
[BUG] Halo size inconsistent composer


LukasStordeur:
The halo where the action icons are displayed around entity shapes is inconsistently sized.

Issue #6860, opened 2026-05-07
Expand/Collapse relations Composer


LukasStordeur:
Allow the user to expand or collapse relations on the canvas

PR #6878, opened 2026-05-18
6860 expand collapse relations composer

LukasStordeur:
# Summary

This PR adds expand/collapse controls to entity shapes in the Instance Composer, allowing users to manage complex service graphs by hiding and revealing nested relation layers.

## Expand Collapse icons buttons
Two new action buttons are added to the halo (right-click context menu) of any shape that has connected relations:

Expand/Collapse layer -> collapses all nested relation layers recursively, or expands only the direct (first-layer) children. The icon updates dynamically between a "collapse" and "expand" state.
Expand all layers  -> reveals the entire subtree of connected shapes at once, regardless of the current collapsed state.

## Validation surfacing for collapsed layers
When a shape has collapsed children, the red validation halo and tooltip on the parent are updated to reflect errors in the hidden subtree.

## Reveal instances on delete root shape
When a shape with collapsed children is deleted from the canvas, its direct (first-layer) children are automatically revealed before removal so they remain visible and accessible.

## Known issue
The halo size around the shapes is not consistent, this was already present and will be handled in subsequent ticket #6877

https://github.com/user-attachments/assets/16f7c792-3d9a-46d2-b09a-735114263be7

--------

Closes:
- #6860


AronH99:
Screenshot 2026-05-19 at 09 25 31

So as shown for the embedded-entity-service some services all get correctly hidden and some don't.
I went through the code it's hard to spot "issues" at this kind of code but something else I noticed was maybe we could show a popover for the link icon as well with a short explanation? Because the other 2 control tools have some explanation too?

LukasStordeur:
Resolved the issue with the embedded entities. It was not related to the collapse/expand feature, but an edge case in the placeholder creation where multiple embedded entities with the same type, but different name were not all properly added to the connection pool from the root instance. The reason the links were visible, was because the embedded entities shapes had the ids of the root instance to know they were connected.

AronH99 (approved):

PR #6891, opened 2026-05-20
ISO8- Add web_order sorting and multi-expand support to the documentation tab

LukasStordeur:
> ISO 8 PR

# Summary

Add two optional annotation properties to Service Instance Details documentation tab collapsibles:

| Property | Type | Behaviour |
|---|---|---|
| `web_order` | `number` (optional) | Sort collapsibles ascending. `-1` or absent → keep declaration order. |
| `web_default_open` | `boolean` (optional) | When `true`, the collapsible is expanded on first render. Absent/`false` → collapsed. If no default item found, expand first one by default. |

Additionally, allow **multiple collapsibles to be open simultaneously** (currently only one at a time is allowed).

closes: 
- #6883

### DIFF
--- a/changelogs/unreleased/6883-sort-documentation-tab-items.yml
+++ b/changelogs/unreleased/6883-sort-documentation-tab-items.yml
@@ -1,0 +1,9 @@
+change-type: patch
+description: Add web_order sorting and multi-expand support to the documentation tab collapsibles
+destination-branches:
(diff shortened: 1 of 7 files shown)


AronH99 (approved):
Thought i already reviewed/looked into this one but forgot to approve i guess

PR #6900, opened 2026-05-21
Add References tab + resolve mutators in Desired State

edvgui:
> Claude opening this PR on behalf of @edvgui.

  ## Summary

  Surfaces the `references` and `mutators` carried on a resource's desired state in the resource details page.

  ### New "References" tab

  Lists every entry in `details.attributes.references` as an expandable row showing the reference's `type` and `id`. Expanding a row reveals its arguments in a Desired-State-lookalike
  layout:

  - **Linked args** (`type: "reference"` / `"resource"`) render as blue clickable chips. Clicking a `reference` chip expands and scrolls to the target row in the same table.
  - **mjson args** are rendered as a JSON block. Nested references listed under `arg.references` (JSONPath → reference id) are substituted in place and rendered as inline clickable chips
  inside the JSON text.
  - **literal args** go through the existing `AttributeClassifier` + `AttributeList` pipeline so JSON / XML / multi-line formatting matches the Desired State tab exactly.

  Expansion state is URL-backed (`useUrlStateWithExpansion`) so it survives reloads and cross-tab navigation.

  ### Mutator resolution on the Desired State tab

  `core::Replace` mutators are interpreted client-side (see `inmanta_core.references.ReplaceValue`):

  1. Extract `(destination, referenceId)` pairs from `attributes.mutators`.
  2. Substitute the value at each `destination` (JSONPath) in a clone of the attributes with a printable sentinel string.
  3. Render mutated attribute rows specially — pure-sentinel values become a single chip, mixed-string values become inline chips, complex values become a JSON code block with inline chips.

  `mutators` and `references` keys are hidden from the Desired State tab (they have dedicated rendering / the References tab).

  ### Cross-tab navigation

  Clicking a chip in the Desired State tab switches to the References tab and auto-expands the target row. A small `useEffect`-based dance defers the expansion to a follow-up render so the
  two URL writes (active tab + expansion) don't clobber each other — each `useUrlState` write captures `location.search` at render time and would otherwise overwrite the sibling key.

  ### Structured / JSON toggle

  A toggle above the Desired State card lets the user flip between the structured view (with mutator chips) and a read-only `CodeEditor` showing `details.attributes` verbatim — useful as an
  escape hatch when something looks off.

  ## Implementation notes

  - **Sentinel trick.** Inline chips in serialized JSON cannot ride a Monaco-backed `CodeEditor`. Instead we substitute reference holes with printable sentinels
  (`@@INMANTA_REF_START@@<uuid>@@INMANTA_REF_END@@`), `JSON.stringify`, then split the resulting text on the sentinel regex and interleave `<Label>` chips between text chunks. Earlier draft
  used a real null byte — `JSON.stringify` escapes it to a 6-char `\u0000` sequence, which broke the regex; printable ASCII sidesteps that.
  - **JSONPath subset.** In-house parser handles `$` root, `[<n>]`, `[<"key">]`, `.key`, and bare-name paths (`value`, `foo.bar`). No wildcards / filters / recursive descent — sufficient for
   what the orchestrator currently emits via `jsonpath_ng`. If/when we see richer expressions, swap to `jsonpath-plus`.
  - **Mutator support.** Only `core::Replace` is interpreted. Other mutator types are ignored (the original attribute value is rendered as-is) until a use case appears.

  ## Preview

https://github.com/user-attachments/assets/8fbb95f9-15c4-4a61-8c50-19a59017d64a

  ## Test plan

  - [ ] Open a resource with `attributes.references` populated; **References** tab appears between Desired State and Requires.
  - [ ] Expand a reference row — args render with `type · id` chips for linked args; literal args use the standard formatting.
  - [ ] Expand a `mjson`-bearing reference — clickable chips appear inline in the JSON block at each `references` JSONPath.
  - [ ] Click a chip — target row expands and scrolls into view.
  - [ ] Open a resource with `core::Replace` mutators (see `resource.json`) — mutated attribute renders as a chip in the Desired State tab; clicking it switches to References tab with the
  target row expanded.
  - [ ] Toggle Desired State → JSON — full `details.attributes` is shown in a read-only editor, including `mutators` and `references` arrays.
  - [ ] Resource with no references shows the empty state on the References tab.


bartv:
Implementation in https://github.com/inmanta/web-console/issues/7354

PR #6908, opened 2026-05-27
fix: reduce redundant environment detail fetches by aligning staleTime with refetch interval

AronH99:
Claude here on behalf of Aron:

## Problem

`GET /api/v2/environment/{id}?details=true` was firing **5 times** on every page load instead of once per poll cycle. Root cause: `staleTime` was not set on the `useContinuous` variant of `useGetEnvironmentDetails`, so it defaulted to `0`. This caused React Query to treat freshly-fetched data as immediately stale — any observer re-attachment during app initialisation triggered a new network request instead of serving the cached response.

## Fix

Added `staleTime: REFETCH_INTERVAL` to `useContinuous` to align the freshness window with the polling cadence. Data is now considered fresh for 5 s after arrival; re-renders within that window read from cache. The network is only hit on the initial fetch and each legitimate 5 s poll tick.

```diff
 useContinuous: (id: string) =>
   useQuery({
     queryKey: getEnvironmentDetailsKey.single(id),
     queryFn: () => get(`/api/v2/environment/${id}?details=true`),
     retry: false,
     select: (data) => data.data,
     refetchInterval: (query) => (query.state.error ? false : REFETCH_INTERVAL),
+    staleTime: REFETCH_INTERVAL,
   }),
```

`gcTime` is intentionally unchanged — this does not affect how long inactive query data is retained in memory.

Halt and resume actions are unaffected: both `HaltButton` and `ResumeButton` call `client.refetchQueries()` on success, which is a forced refetch that bypasses `staleTime` entirely and updates the UI immediately.

## Measured results

| Metric | Before | After | Delta |
|---|---|---|---|
| `GET /api/v2/environment/{id}?details=true` | 5 calls | 2 calls | −60% |
| `POST /api/v2/graphql` | 15 calls | 6 calls | −60% |
| Total API calls on page load | 24 | 12 | −50% |

The 2 remaining environment calls are expected: 1 on mount + 1 at the first 5 s poll boundary.

## Notes

- The GraphQL reduction is a side effect — the redundant environment re-renders were cascading into other query observers being invalidated in the same render cycles.
- The same pattern (missing `staleTime` on `useContinuous` hooks) may apply to other continuous queries using `REFETCH_INTERVAL` and is worth a follow-up audit.

## Test fix

`EnvironmentControls.test.tsx` uses a shared `testClient` singleton across all tests. With `staleTime: 0` (before), every mount immediately refetched so cross-test cache pollution was invisible. With `staleTime: REFETCH_INTERVAL`, cached data from one test's MSW handler survived into the next test, causing stale responses to be served. Fixed by adding `testClient.clear()` to `afterEach` to restore proper test isolation.

## Test plan

- [ ] Environment details appear correctly in the sidebar on the Resources page
- [ ] Halt / Resume buttons in `EnvironmentControls` reflect live environment state
- [ ] Network tab shows `GET /api/v2/environment/{id}?details=true` firing once on load, then ~once per 5 s — not 5× during initialisation
- [ ] Settings → Environment tab still loads correctly (`useOneTime` is unchanged)

### DIFF
--- a/changelogs/unreleased/fix-environment-details-stale-time.yml
+++ b/changelogs/unreleased/fix-environment-details-stale-time.yml
@@ -1,0 +1,5 @@
+description: Fixed redundant environment detail fetches on page load by aligning staleTime with the polling interval, reducing API calls by 50%.
+change-type: patch
+destination-branches: [master, iso9]
+sections:
+  bugfix: "{{description}}"
--- a/changelogs/unreleased/fix-environment-details-stale-time.yml
+++ b/changelogs/unreleased/fix-environment-details-stale-time.yml
@@ -1,5 +1,3 @@
 description: Fixed redundant environment detail fetches on page load by aligning staleTime with the polling interval, reducing API calls by 50%.
 change-type: patch
 destination-branches: [master, iso9]
-sections:
-  bugfix: "{{description}}"
--- a/src/Data/Queries/Slices/Environment/GetEnvironmentDetails/useGetEnvironmentDetails.ts
+++ b/src/Data/Queries/Slices/Environment/GetEnvironmentDetails/useGetEnvironmentDetails.ts
@@ -35,6 +35,7 @@ export const useGetEnvironmentDetails = (): GetEnvironmentDetails => {
         retry: false,
         select: (data) => data.data,
         refetchInterval: (query) => (query.state.error ? false : REFETCH_INTERVAL),
+        staleTime: REFETCH_INTERVAL,
       }),
   };
 };
--- a/src/UI/Root/Components/Sidebar/EnvironmentControls/EnvironmentControls.test.tsx
+++ b/src/UI/Root/Components/Sidebar/EnvironmentControls/EnvironmentControls.test.tsx
@@ -36,7 +36,10 @@ describe("EnvironmentControls", () => {
   const server = setupServer();
 
   beforeAll(() => server.listen());
-  afterEach(() => server.resetHandlers());
+  afterEach(() => {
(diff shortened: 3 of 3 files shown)


AronH99:
@LukasStordeur 

So this is a pr claude found on it's own whilst I was experimenting with the performance issues on the resources page, and I think it's a good change.
It essentially reduces the number of calls the useGetEnvironmentDetails needs to make during initialization of a page.
Then I made a skill for my own to test if it could make a pr for this on it's own and this worked quite well i reckon.
Let me know what you think :)

LukasStordeur:
The refetchInterval is working as follow on a query
- A component mounts and subscribes to the query, the query is mounted and the interval starts.
- if a there's a second subscriber to the query mid-interval, that will join the existing interval, and shouldn't trigger a refetch.
- All components unmounts resets the refetchInterval.

If you use staleTime AND refetchInterval, with the same exact value, the assumption is that they would both remain in sync, but that is not the case. They have no awareness of eachother. Invalidating queries for example will not reset the refetchInterval, eventhough it will trigger a manual refetch. A query that is being slower (can be spikes in the connection, or whatever reason) will desync the two parameters. The staleTime starts counting when the query is resolved, while refetchInterval resets when sending out the api call.

Here again, it tries to solve a rendering issue without addressing the root cause, which would probably be related to components mounting and unmounting when not needed. If your entire tree that contains the query needs to remount due to a statechange, yes, that might reset the interval and refetch the data. Using staleTime is covering that up.

AronH99:
All right, makes sense, i'll close this pr

PR #6909, opened 2026-05-27
fix: reduce unnecessary re-renders on Resources page

AronH99:
Claude here on behalf of Aron:

## Problem
Two independent issues on the Resources page:

**1. React re-renders** — every `ResourceTableRow` re-rendered on every polling cycle because inline object literals passed as Flex/FlexItem props always produced new references, defeating memoization. With 100 rows visible this produced ~100 wasted renders per cycle. The `rows` array was also recreated unconditionally on every parent render.

**2. Layout/paint thrash from the Spinner** — the PF v6 `<Spinner>` in the deploying-progress label has no compositor layer of its own. Each animation tick invalidates the ancestor table + all SVG status icons, triggering a full Layerize → Layout → Paint → Commit cycle every ~19 ms. On a 200k-resource environment: spinner visible = 41 FPS, spinner removed = 60 FPS. A single 15×15 px spinner was consuming ~30% of the frame budget.

## Fix
**React render reduction:**
- Wrapped `ResourceTableRow` and `ResourcesTable` in `React.memo`
- Hoisted all inline prop objects (`gapNone`, `gapMd`, `noWrap`, `alignCenter`, `statusCellStyle`, `bullseyeStyle`, `inlineFlex`) to module-level constants so memo's shallow comparison passes
- Moved `rows` into `useMemo` in `Page.tsx`, keyed on `data?.resources`
- Memoized `popoverBody` in `ResourceTableRow` with `useMemo`
- Stabilized `onSort` with `useCallback`

**Spinner compositor isolation:**
- Added `willChange: "transform"` + `isolation: "isolate"` to the deploying `Spinner` — forces it onto its own compositor layer so its animation no longer invalidates the ancestor table and surrounding SVG icons

### React render reduction
Measured live (102 resources, 100 visible) using `performance.getEntriesByType('measure')`.

| Metric | Before | After | Delta |
|---|---|---|---|
| Total React render measures | 29,848 | 4,706 | **−84%** |
| `ResourceTableRow` renders | 1,400 | 0 | **−100%** |
| `Flex` renders | 2,933 | 248 | **−92%** |
| `FlexItem` renders | 1,928 | 185 | **−90%** |
| FCP | 600 ms | 500 ms | −100 ms |
| JS heap used | 130.0 MB | 127.9 MB | −2.1 MB |

### Spinner paint thrash (200k-resource env => taken from Bart's findings on Slack)

| Metric | Value |
|---|---|
| Layerize events | 721 in 14 s (~52/s) |
| Layout events | 725 in 14 s (~52/s) |
| Paint events | 2,168 in 14 s (~155/s) |
| FPS with spinner | 41 |
| FPS without spinner | 60 |
| Frame budget consumed by spinner | ~30% |

## Notes
`ResourceStateInfo` inside the popover could similarly be memoized as a follow-up. The spinner thrash is also worth filing upstream — `pf-v6-c-spinner` shouldn't invalidate ancestors on every tick.

## Test plan
- [ ] Resources page loads and displays rows correctly
- [ ] Sorting columns still works
- [ ] Status popover opens and shows correct resource state info
- [ ] Deploying spinner appears next to the deploy count when resources are deploying
- [ ] No visual regressions on orphan rows (striped styling intact)
- [ ] React DevTools Profiler confirms `ResourceTableRow` does not appear in flamegraph between poll cycles
- [ ] FPS improves visibly during active deploy on large environments


AronH99:
@LukasStordeur 
This pr essentially it does 2 things:
1) By memoizing components and or their props we get stable references meaning less re-rendering cycles (less re-rendering).
2) Added will-change: transform + isolation: isolate to the deploying spinner so it animates on its own layer and stops triggering repaints on the rest of the page (svg's in particular).

I verified this pr as well whilst I was experimenting with claude a little, for the performance numbers which were quite interesting.

bartv:
I wonder if the pulsing dot is also causing too many re-renders. They are deeper in the tree but still

AronH99:
@bartv If I let Claude analyze it it says it's fine because:
Unlike the spinner's strokeDashoffset which runs on the main thread, the opacity animations of the blinking dot don't have any effect on surrounding components.
I looked into it and you can indeed find some information about those properties not repainting vs repainting the browser. 
Learned something new today which probably mostly matters in settings like these

LukasStordeur (changes requested):
I'm not entirely convinced.

I understand the reasoning, and while being aware objects declared in the template would be recreated on each render, I don't think it would have that big of an impact. Storing them in variables shouldn't make that big a difference, unless there's an underlying re-rendering issue. In which case, this is just masking the underlying problem. 

Right now, in the rows you're extracting the objects into variables to keep a stable reference, but not on the spinner, which is apparently the one that was pointed out being a problem.

Having the rule set is not a bad thing however to not redeclare objects in the template, but then we need to apply it consistently too. I did a quick search and there's potentially a linting rule that can enforce that.

I do believe the correct approach would be to have a better control on the re-rendering of the component(s). And have test cases asserting we don't have unnecessary rerenders. 
Without tests, we cannot trust the feedback about the potential improvements. They could be biased or made up by Claude.

The real change that would have impact, is memoizing the rows, because there's a bit of logic in creating the rows.

LukasStordeur:
I did a manual comparison for the spinner in a paternfly codesandbox with 150 spinners that render at an interval of 2 seconds to simulate the case.
This is the tracing on 10s, without the styling isolation.
image
 And this is the result with the styling isolation:
image

We only have 1 Spinner on the page, which means the performance gain is marginal to none on a single component.

AronH99:
@LukasStordeur 
So apparently modern browsers will handle these repainting issues on their own so there is indeed not much to be gained from adding the willChange: "transform" so I dropped that change.
I restructured the rest of the changes a little bit to only include the things which cause the most re-rendering.
The way it is now the amount of re-renders should be down from 37.5K to about 1.5k.
But u have a good point, I will add some tests to catch regression related to re-rendering of the ResourcesTable and ResourceTableRow.

src/Slices/Resource/UI/ResourcesPage/ResourcesTable.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+const rowRenderSpy = vi.fn();
+
+vi.mock("./ResourceTableRow", () => ({
+  ResourceTableRow: (props: { row: ResourceRow }) => {
+    rowRenderSpy(props.row.id);
+
+    return (
+      <tbody>
+        <tr>
+          <td data-testid={`row-${props.row.id}`}>{props.row.type}</td>
```
LukasStordeur: This has no real testing value imo. Mocking the component you want to test is like testing a different component entirely. You bypass all the internal lifecycles that could triger rendering issues.

```ts
const TrackedComponent = (props) => {
  renderSpy()
  return <OriginalComponent {...props} />
}
```
Keep the original component instead.

LukasStordeur (changes requested):

LukasStordeur (approved):

Issue #6778, opened 2026-03-26
Migrate Desired State page filters to side panel drawer


LukasStordeur:
## Context

The Resources page filters have been migrated to a side panel drawer (using PatternFly `Drawer` + `DrawerPanelContent`). Other pages still use the old inline toolbar `FilterWidget` pattern and should be updated to match.

## Current state

`src/Slices/DesiredState/UI/Components/TableControls.tsx` renders a `FilterWidget` inline inside a PatternFly `Toolbar`, alongside a `CompileWidget` and a `CompareSelectionWidget`.
The filter widget lives in `src/Slices/DesiredState/UI/Components/FilterWidget.tsx`.

## Goal

Migrate the Desired State page filters to the new side panel pattern:

- Add a `Drawer` + `DrawerContent` + `DrawerPanelContent` wrapping the table
- Move `FilterWidget` content into a `DrawerPanelContent` component
- Add a filter toggle button to the toolbar (with active filter count badge), replacing the inline widget
- The `CompileWidget` and `CompareSelectionWidget` should remain in the toolbar
- Track `isDrawerExpanded` and `activeFilterCount` state on the page component

## Reference implementation

See `src/Slices/Resource/UI/ResourcesPage/Page.tsx` and `src/Slices/Resource/UI/ResourcesPage/Components/ResourceTableControls.tsx` for the pattern to follow.

PR #6914, opened 2026-05-28
Feature/6778 desired state filter drawer

LukasStordeur:
# Description

## Summary
Migrates the filters on the Desired State page from the inline toolbar pattern to a collapsible side panel drawer, bringing it in line with the existing Resources page UX.

## Key changes:

Replaced the single FilterWidget.tsx with a proper FilterWidget/ folder containing:
ConnectedFilterWidget — owns URL-persisted filter state via useUrlStateWithFilter, memoized to prevent unnecessary re-renders of the filter panel when the table refreshes
FilterWidgetComponent — pure presentation component rendering status, date-range, and version-range filter controls inside a FilterDrawerPanelContent
Added a reusable FilterDrawer module (FilterDrawerPanelContent, FilterToggleButton) under src/UI/Components/FilterDrawer/ so other pages can adopt the same pattern
Updated TableControls and Page.tsx on the Desired State slice to wire up the drawer toggle and pass the onClose callback
Simplified Page.test.tsx now that filter state is managed inside ConnectedFilterWidget rather than the page itself

closes:
- #6778 

Screenshot 2026-05-27 153919


changelogs/unreleased/6778-desired-state-filter-drawer.yml
```diff
@@ -0,0 +1,5 @@
+description: The Desired State page filters have been migrated to a side panel drawer, matching the pattern already used by the Resources page.
+destination-branches: [master]
+change-type: major
```
AronH99, line 3: is this not a minor? it's a non breaking change right?

src/Slices/DesiredState/UI/Components/FilterWidget/ConnectedFilterWidget.tsx
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  const filterWithDefaults = useMemo(
+    () =>
+      filter.status
+        ? filter
+        : {
+            ...filter,
+            status: [
+              DesiredStateVersionStatus.active,
```
AronH99: Something small but, you already indicate it in a comment here but can't we share the constants between the ConnectedFilterWidget and the Page.tsx?
Something like:
```ts
export const DEFAULT_DESIRED_STATE_STATUSES = [
  DesiredStateVersionStatus.active,
  DesiredStateVersionStatus.candidate,
  DesiredStateVersionStatus.retired,
];
```

That way there is one source of truth to avoid future issues?

AronH99 (commented):
1 more thing maybe:

Image

I can't really clear the "default" filters which should be possible right?
Otherwise part of the code (hasActiveFilters ternary) is sort of redundant if this should not be possible.

changelogs/unreleased/6778-desired-state-filter-drawer.yml
LukasStordeur: I'm waiting on Bart's response, but if it's only iso10, we need it to be a major to make the divergence or the merge-tool will reject this.

LukasStordeur: Confirmed: this is meant for 10 only.

LukasStordeur:
https://github.com/user-attachments/assets/48d160e0-13cf-48bb-9e75-54be1ddc83c7

AronH99 (approved):

src/index.tsx
```diff
@@ -1,4 +1,7 @@
 import "@patternfly/react-core/dist/styles/base.css";
+import "@patternfly/react-styles/css/components/_index.css";
+import "@patternfly/react-styles/css/layouts/_index.css";
+import "@patternfly/react-styles/css/utilities/_index.css";
```
AronH99: Small question maybe, why do we need all of these imports because I can imagine some of it is redundant but probably needed to fix a certain issue?

LukasStordeur: I added them because certain css variables seemed to not resolve. But in hindsight perhaps it was a red herring related to the Form tag. I'll remove them.

Issue #6903, opened 2026-05-22
[REQUEST] Don't send r-only attributes when updating the service


edvgui:
### Summary

When editing a service instance, this happens a lot: 

<img width="705" height="245" alt="Image" src="https://github.com/user-attachments/assets/150d64de-324d-458b-a8f9-3be5ec116766" />

### Problem / Use Case

The reason is that the attribute is being updated by the compiler or a resource while the user is editing the service.  This is perfectly fine because the user can not edit this attribute either, so there is no conflict.  Except that the web console tries to send this attribute (the old version of it) during update, which the server rightfully rejects as it is different from the current version and the user is not allowed to change it.  

### Proposed Solution(s)

The web-console should not try to send read-only attributes during updates (only expert mode can do that).  Ideally the json editor should also not contain the read-only values, so objects can easily be copy-pasted.

### Impact

Editing services can be a real pain

### Additional Context

/

PR #6915, opened 2026-05-28
Replace PATCH v2 with PUT for edit instance (#6903)

LukasStordeur:
# Description

## Summary

  - Replaces `PATCH /lsm/v2/service_inventory/<entity>/<id>` (used when `strict_modifier_enforcement = true`) with `PUT
  /api/v1/service_inventory/<entity>/<id>?current_version=<v>&ignore_read_only_attributes=true`
  - Adds a new `usePut` (env-aware) HTTP primitive to the query helpers layer
  - Introduces a `PutAttributes` data module (`usePutAttributes` hook + `getBodyPut` helper) with body shape: `{ put_id: <uuid>, attributes: <all sanitized
  attributes> }`
  - Cleans up all dead PATCH v2 code (`BodyV2`, `getBodyV2`, `apiVersion` branching in `usePatchAttributes`)
  - `EditForm` calls `usePutAttributes` for strict-enforcement services and `usePatchAttributes` (v1, diff-only) for all others — both hooks called unconditionally
   per React rules

closes:
- 6903

_Changes made in collaboration with Claude_

### DIFF
--- a/changelogs/unreleased/6903-put-edit-instance.yml
+++ b/changelogs/unreleased/6903-put-edit-instance.yml
@@ -1,4 +1,4 @@
-description: Edit instance now uses PUT /api/v1/service_inventory with ignore_read_only_attributes instead of PATCH v2 for services with strict modifier enforcement.
+description: Edit instance now uses PUT /lsm/v1/service_inventory with ignore_read_only_attributes instead of PATCH v2 for services with strict modifier enforcement.
 issue-nr: 6903
 change-type: patch
 destination-branches: [master, iso9]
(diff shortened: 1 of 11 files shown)


AronH99 (changes requested):
Still got a test failing because of the api/v1 vs lsm/v1.
I Think u need to cleanup all of the api/v1 issues and replace them to lsm/v1 as intended no?

LukasStordeur:
> Still got a test failing because of the api/v1 vs lsm/v1. I Think u need to cleanup all of the api/v1 issues and replace them to lsm/v1 as intended no?

Yes was already fixing it :)

AronH99 (approved):

PR #6916, opened 2026-05-28
fix: flakiness desired-state

AronH99:
# Description

Improve stability of the desired state e2e test by fixing a race condition on filter apply and a false "No Dry runs exist" assertion on subsequent runs.
Second one is more quality of life first one is the one I came across a couple of times in the failing builds on Jenkins and hopefully it will resolve this problem.

=> Part of build mastering


cypress/e2e/scenario-4-desired-state.cy.js
```diff
@@ -278,7 +284,7 @@ describe("Scenario 4 Desired State", () => {
     cy.get('[role="menuitem"]').contains("Compare with current state").click();
 
     cy.get("h1").should("have.text", "Compliance Check");
-    cy.get('[aria-label="ReportListSelect"]').contains("No Dry runs exist").should("be.visible");
```
LukasStordeur, line 281: you removed the assertion, how do you know now that there are no dry runs yet available?

AronH99: We don't really care if they exist or not right?
The point of this e2e test was to see if a dry run could be done/performed properly?
Or am I wrong?
Hence why I removed it, let me know if that's not correct then I'll adjust again

LukasStordeur: yes we do care actually :) this is also to assert there are no leaks, or that we start with a clean slate, and that clean slate is expected to not have any dryruns available in the select.

LukasStordeur: The flake is because the correct version is not opened in the page.

AronH99: All right I'll only keep the other change then

LukasStordeur (approved):

PR #6918, opened 2026-05-29
improv: added unit testing for the AttributesTable

### LINKED ISSUE #6722: [REQUEST] Attribute page for service instance needs unit testing
### Summary

Unit testing needs to be added for the AttributesTabContent component.
Specifically testing for the bug where we need to make a difference between attributes who have children and those who don't 
=> clickable vs normal pointer

- **Area of the app**:  Inventory
- **Type**:  Testing
- **affected versions**: Iso10

### Problem / Use Case

Stability upgrade/improvement

Image

AronH99:
# Description

Added unit testing for the service instance details AttributesTable (AttributesTab).
This should only be for iso 10 right (this what was defined on the ticket)?

closes https://github.com/inmanta/web-console/issues/6722


changelogs/unreleased/attributesTable-unit-test.yml
```diff
@@ -0,0 +1,3 @@
+description: Added unit testing for AttributesTable on the service instance details page
+change-type: patch
+destination-branches: [master]
```
LukasStordeur: Why only master?

LukasStordeur: Missing ticket reference.

LukasStordeur (changes requested):

changelogs/unreleased/attributesTable-unit-test.yml
AronH99: It was specified on the ticket I wrote, but apparently I misunderstood the, should be fixed

LukasStordeur (approved):

PR #6922, opened 2026-05-29
Refactor compound resource state types to use as const objects as single source of truth

### LINKED ISSUE #6814: Rework/Review the STATUS_FIELD mapping inside GetResources/helpers.ts
There was a concern where the mapping which was used for the GetResources filters (STATUS_FIELD_MAP) was a little complex or that at least it could be improved upon with less code.
=> https://github.com/inmanta/web-console/pull/6803#discussion_r3116209888

AronH99:
# Description

Refactored the domain resource type file to use single source of truth principle for the typing related to it.
It's quite a bit more clear what's what and I think the naming leaves little room for interpretation at the moment.
I believe this is more closer to what you probably had envisioned as well?

Sidenote:
Tried a claude skill to make changelog + commit + branch (and push)

closes https://github.com/inmanta/web-console/issues/6814


src/Core/Domain/Resource/Resource.ts
```diff
@@ -17,75 +17,79 @@ import { Sort } from "../Sort";
...
+  temporarily_blocked: "TEMPORARILY_BLOCKED",
+} as const;
 
-/**
- * Union of all compound state values.
- * Combines {@link LastHandlerRun}, {@link Compliance}, and {@link Blocked}.
- */
-export type CompoundState = LastHandlerRun | Compliance | Blocked;
+export type LastHandlerRunKey = keyof typeof LAST_HANDLER_RUN;
+export type LastHandlerRun = (typeof LAST_HANDLER_RUN)[LastHandlerRunKey];
```
LukasStordeur: I'd add that this reflects the value in the naming. It might get confusing otherwise, and get mistaken for the overall shape of a Last handler run. same with Compliance/Blocked, I'd name them xxxValue, so you have a distinct pair Key-Value.

LukasStordeur (changes requested):
Much better indeed.

src/Core/Domain/Resource/Resource.ts
AronH99: good idea

LukasStordeur (approved):

PR #6927, opened 2026-06-01
Add orphaned filter to the Resources page status filter

### LINKED ISSUE #6866: Resource view : Add a way to hide/show orphaned resources in filters
Currently, the user can't filter manually on orphaned resources. Atm, only the initial default filters adds the !orphaned filter key.

AronH99:
# Description

So this took a little longer than anticipated because:
The toggle group as we decided to use from PatternFly is only available from version 6.5.x and onwards.
I was first looking as to why the component would look so bad (see frontend channel slack for screen) but it seemed like they made the component way better looking in 6.5.
I was looking around whether or not I should update anything else to 6.5 or higher from PatternFly, stumbled upon some issues here as well so I decided to keep the impact minimal and only update the relevant package I need for this PR.
Let me know what you think?
Also needed to adjust some e2e test to make more improvements out of testing the filtering features of the resource page.

closes https://github.com/inmanta/web-console/issues/6866

https://github.com/user-attachments/assets/3a9ebc4b-39ea-461b-a6d2-7c1ac4e86909

Note: commit, changelog, branch pushed by Claude


LukasStordeur (changes requested):
1. Every form element requires a label. The other filter sections all have a label.
2. I'm not sure "Both" is the right solution. One could expect then both chips to be added. Even if it results in the same.  A two button toggle "Include/Exclude", that are not exclusive might be enough then. So each can be added, or removed. That pattern can easily be reused on other boolean filters, so we'd have a reusable component. 
You can use the multi-select variant for that. 

3. I'm not certain what you mean by: only available from 6.5, we've been using toggle group buttons in plenty of places. Open the form to create/edit an instance for example. Or the Instance Detail page with the Attributes tab open. 

_I resolved some of the css issues that were related to the Form tag not being used, so certain css-variables wouldn't resolve properly. You might require the same fix for this case. You can find that solution back in the DesiredState FilterSidePannel PR. You'll need to prevent-default submit action to avoid the UI to rerender entirely on `enter`_ 

4. We better keep the PF packages in sync when it comes to versions. I wanted to implement a grouping for Dependabot, but need to look into it further to set it up properly.

AronH99:
Screenshot 2026-06-02 at 13 19 41

Screenshot 2026-06-02 at 13 20 09

Screenshot 2026-06-02 at 13 19 32


So I took the liberty to make everything more in line with the intended styling of PatternFly and everything looks consistent now. 
I couldn't get the Orphaned and Include/Exclude on the same line without breaking the Form style of PatternFly. 
Though I think it still looks good now.
The new component is mutually exclusive but all of the options can be turned off, I think this way the user also can't make any mistakes against selecting all of the options at the same time...
Let me know if u have any other concerns..

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/FilterWidgetComponent.tsx
```diff
@@ -146,46 +147,48 @@ export const FilterWidgetComponent: React.FC<FilterWidgetComponentProps> = ({
...
-            removeAgentChip={removeAgentChip}
-            removeValueChip={removeValueChip}
-            removeStatusChip={removeStatusChip}
-            clearTypeFilters={clearTypeFilters}
-            clearAgentFilters={clearAgentFilters}
-            clearValueFilters={clearValueFilters}
-            clearStatusFilters={clearStatusFilters}
-          />
-        </Stack>
+        <Form>
```
LukasStordeur: You need to prevent the default submit event, or the whole page will rerender if a user hits enter.

LukasStordeur: Claude:
Separate from the missing `onSubmit` handler (already flagged): this `<Form>` creates **nested `<form>` elements** in the DOM, because `ResourceFilterForm` already renders `<Form onSubmit={(e) => e.preventDefault()}>` internally. Nested forms are invalid HTML — browsers will silently break the inner one.

Two clean options:
- Remove the `<Form>` here and wrap only `StatusFilterSelect` with its own `<Form onSubmit>` (since that's the tab that actually needs it for the switch/toggle labels).
- Remove the `<Form>` from `ResourceFilterForm` and rely solely on this outer one — but then the `onSubmit` prevention must live here.

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/StatusFilterSelect.tsx
```diff
@@ -89,14 +90,34 @@ export const StatusFilterSelect: React.FC<StatusFilterSelectProps> = ({
...
+                value: "!orphaned",
+                buttonId: "orphaned-exclude",
+              },
+            ]}
+          />
+        </FormGroup>
+      </StackItem>
+      <StackItem>
+        <FormGroup label={words("resources.filters.status.isDeploying")}>
+          <Switch
```
LukasStordeur: `id={words("resources.filters.status.isDeploying")}` resolves to `"Is Deploying"` — an ID containing spaces is **invalid HTML** and silently breaks the `htmlFor` → `id` association that `FormGroup`'s label relies on. The label will not be programmatically linked to the switch.

Use a stable, space-free ID here (e.g. `"is-deploying"`). Same fix needed for the Purged switch in `ResourceFilterForm.tsx` — `"Purged"` has no spaces so it's technically fine, but using a translated string as an HTML ID is fragile (breaks if the translation ever changes).

src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
```diff
@@ -0,0 +1,57 @@
+import { ToggleGroup, ToggleGroupItem } from "@patternfly/react-core";
```
LukasStordeur, line 1: Claude:
Missing `import React from "react"`. Every other component in this area (`ResourceFilterForm`, `StatusFilterSelect`) imports it explicitly. Either add the import here or clean up the inconsistency across the board — but right now this file is the odd one out.

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/ResourceFilterForm.tsx
```diff
@@ -130,14 +130,14 @@ export const ResourceFilterForm: React.FC<ResourceFilterFormProps> = ({
           </Title>
         </StackItem>
         <StackItem>
```
LukasStordeur, line 131: Claude: 
This `FormGroup` — and the entire "Desired State" section — is rendered in a `Stack` **outside** the `<Form>` element above it. The `<Form>` closes before this stack, so PatternFly's form grid styles (`.pf-v6-c-form`) won't apply here. Verify the layout actually renders correctly; if it doesn't, this section needs to move inside the Form.

Separate issue: the `Switch` still has `aria-label` set to the same text as the `FormGroup` label. Since `aria-label` takes precedence over a programmatically associated `<label>`, the FormGroup label becomes redundant for accessibility — the `htmlFor` link is effectively ignored by screen readers. Pick one strategy: either keep `aria-label` on the Switch (and drop the FormGroup wrapper), or use the FormGroup label with `htmlFor` (and remove `aria-label` from the Switch). Same applies to the `isDeploying` switch in `StatusFilterSelect.tsx`.

cypress/e2e/scenario-6-resources.cy.js
```diff
@@ -13,6 +13,15 @@ const expectFilteredLessThan = (alias) => {
   });
 };
 
+// Helper to verify row count is more than initial
```
LukasStordeur, line 16: `expectMoreThan` doesn't follow the naming pattern of `expectFilteredLessThan`. Should be `expectFilteredMoreThan`.

LukasStordeur (changes requested):
Your E2E tests are failing on the resource view, I doubt this is a flake :)

AronH99:
@LukasStordeur
1) For both switches I implemented PatternFly's idiomatic way of working again
2) For the e2e tests they should be way more stable now, and I tried 5000 different ways to make the drag and drop work without the wait() but it's seems like an impossible task, even claude analyzing it via google mcp "found" some fixes but then again nothing worked.
So it's either this or I need to remove it temporarily because can't get it to work otherwise.
I think it's acceptable because we still get the benefit of it being tested...
Do let me know what u think

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/StatusFilterSelect.tsx
```diff
@@ -88,14 +88,32 @@ export const StatusFilterSelect: React.FC<StatusFilterSelectProps> = ({
...
+      </StackItem>
       <StackItem>
         <Switch
-          id={words("resources.filters.status.isDeploying")}
-          aria-label={words("resources.filters.status.isDeploying")}
+          id="is-deploying"
           label={words("resources.filters.status.isDeploying")}
           isChecked={selectedStatuses?.includes("isDeploying") ?? false}
           onChange={(_event, hasChanged) => handleIsDeploying(hasChanged)}
-          isReversed
```
LukasStordeur, line 83: Why do we remove the isReversed? Isn't it supposed to show the label first, then the toggle? Not that it matters much since it will be replaced in a future iteration.

src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
```diff
@@ -0,0 +1,57 @@
...
+ *  @prop {string[]} selected - The full list of currently active filter values. Only values matching this group's options are considered.
+ *  @prop {function} onChange - Called with the updated full list after applying mutual-exclusion logic.
+ *
+ * @notes At most one option can be active at a time. Selecting an inactive option deactivates any
+ * other option in the group. Selecting the active option deactivates it, leaving no selection.
+ * Unrelated values in the selected list are preserved.
+ *
+ * @returns {React.FC} The OptionalToggleGroup component.
+ */
+export const OptionalToggleGroup = <T extends string>({
```
LukasStordeur, line 29: Missing the React Component typing.

src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
LukasStordeur: Not resolved :)

LukasStordeur (commented):
The e2e are fine for now, we'll see how it goes

LukasStordeur (changes requested):

src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
AronH99: I removed them in my files.
You can see the runtime JSX for React in all files unless you need to import something for actual use (this is a comment so gets excluded from bundle)

src/Slices/Resource/UI/ResourcesPage/Components/FilterWidget/StatusFilterSelect.tsx
AronH99: Looked pretty good like this as well, don't think it matters too much.

src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
AronH99: see comment above

LukasStordeur: And I'm saying you're missing the return type React.FC :)

LukasStordeur (changes requested):

LukasStordeur (approved):

PR #6945, opened 2026-06-04
Fix the zoom level of the composer on render

LukasStordeur:
# Description

This should prevent the view from clipping shapes.


AronH99 (approved):

AronH99 (changes requested):
Sorry I did found something weird when running the branch, on initial load I get:

Image

LukasStordeur:
I added a scale limit to prevent the fitting to take up the entire screen when you have a large screen and few shapes.

AronH99 (approved):

PR #6960, opened 2026-06-08
Fix DictField JSON editor content color in dark mode

### LINKED ISSUE #6948: [BUG] DictField should use Light/Dark theme
### Summary

The new DictField using the CodeEditor isn't setting the isDarkTheme when the UI is in dark mode.

AronH99:
# Description

I know this PR was somewhat out of scope for this ticket but I found these changes to be very necessary for longterm health of the console. 
I will explain what actually happened:

1) I fixed the bug whereby editor content color didn't change immediately into the requested theme. 
In this case DictField but there were some other components that needed this changes as well.
2) Made a useTheme hook to collect all theming logic from the console so it could be used throughout.
3) Adjusted unit tests where needed.

These changes mainly were needed because of the bug of the code editor but also some added benefits now are:

- Improved tests
- Consistent logic, 1 source of truth 
=> useful because of upgrades to PatternFly which might break the current implementation for example.

So all in all not too much work for quite a good change in my opinion.
If you want I can as well adjust the ticket linked to this PR?

closes https://github.com/inmanta/web-console/issues/6948

Screenshot 2026-06-08 at 11 59 13

### DIFF
--- a/src/UI/Components/DarkmodeOption/useTheme.ts
+++ b/src/UI/Components/DarkmodeOption/useTheme.ts
@@ -1,0 +1,99 @@
+import { useCallback, useEffect, useState } from "react";
+
+const STORAGE_KEY = "theme-preference";
+
+/**
+ * Framework-agnostic attribute set on <html> to signal the active theme.
+ * The MutationObserver watches this instead of any UI-library-specific class,
+ * so the reactive hook stays stable across PatternFly version upgrades.
+ */
+const THEME_ATTR = "data-theme";
+
+/**
+ * PatternFly-specific class names applied alongside data-theme.
+ * When upgrading PatternFly, only these two constants need updating.
+ */
+const PF_DARK_CLASS = "pf-v6-theme-dark";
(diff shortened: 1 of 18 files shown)


src/UI/Components/DarkmodeOption/useTheme.ts
```diff
@@ -0,0 +1,99 @@
...
+ * When upgrading PatternFly, only these two constants need updating.
+ */
+const PF_DARK_CLASS = "pf-v6-theme-dark";
+const PF_LIGHT_CLASS = "pf-v6-theme-light";
+
+/**
+ * Retrieves the user's theme preference from localStorage or the system's color scheme.
+ * @returns {"dark" | "light"} The resolved theme preference.
+ */
+export const getThemePreference = (): "dark" | "light" => {
```
LukasStordeur: You're extracting other strings, which I do like. These might be worth storing in a type too

LukasStordeur (approved):
Approving, the minor improvement shouldn't require re-review :)

PR #6962, opened 2026-06-08
Fix heading-order axe violation in toast alerts

### LINKED ISSUE #6785: Heading-order issue with the rework of notifications on the status page
### Summary

The heading-order rule we disabled on the Slices => Status => UI => Page.test.tsx should actually be removed and then the tests need to pass.

### Steps to Reproduce
Remove the heading-order from the page's test and u will see the error showing up when running the test.

### Expected Behavior
Should just pass the test.

### Actual Behavior
Gives an error about headings

AronH99:
# Description

Toast alerts in `AppAlertProvider` rendered their title as `h4` (PatternFly's default), which caused an axe `heading-order` violation when the page already had an `h1`. Fixed by setting `component="div"` on toast alerts — the title is not a document heading, and the live region already handles screen reader announcements. The suppression rule in the Status page test is removed.

closes https://github.com/inmanta/web-console/issues/6785

### DIFF
--- a/src/Slices/Status/UI/Page.test.tsx
+++ b/src/Slices/Status/UI/Page.test.tsx
@@ -208,7 +208,7 @@ describe("StatusPage", () => {
     expect(within(errorContainer).getByText("error")).toBeVisible();
 
     await act(async () => {
-      const results = await axe(document.body, {});
+      const results = await axe(document.body);
 
       expect(results).toHaveNoViolations();
     });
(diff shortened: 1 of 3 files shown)


src/Slices/Status/UI/Page.test.tsx
```diff
@@ -208,12 +208,7 @@ describe("StatusPage", () => {
     expect(within(errorContainer).getByText("error")).toBeVisible();
 
     await act(async () => {
-      const results = await axe(document.body, {
-        // See ticket #6785
-        rules: {
-          "heading-order": { enabled: false },
-        },
-      });
+      const results = await axe(document.body, {});
```
LukasStordeur:
```suggestion
      const results = await axe(document.body);
```

LukasStordeur (changes requested):

LukasStordeur (approved):

PR #6965, opened 2026-06-09
feat(facts): expand JSON values inline with formatted code block

### LINKED ISSUE #6934: [REQUEST] - Display values of Facts in codelog when parseable JSON
### Problem / Use Case
The problem is that we sometimes can get a stringified json object. And displaying it in the colums is not ideal. The length of the JSON object can be extensive, and even contain pieces of xml.
Image

### Proposed Solution

Whenever the value is parseable JSON, the row should be expandable, and we should display a link button in the cell "View Value". 
When the user clicks on the cell, the collapsed section should expand. In the collapsed section, we should display the full value in a codelog component. 

The user should be able to copy the full value thanks to a copy button on the codelog component.

### Impact

This should be implemented on both ISO 9 and 10.

AronH99:
# Description

Facts table rows whose `value` field contains a valid JSON object or array now show a **"View Value"** link button instead of the raw string. Clicking it expands an inline `CodeBlock` with the pretty-printed JSON and a copy-to-clipboard button. Rows with plain-text values are unaffected.

closes https://github.com/inmanta/web-console/issues/6934

Screenshot 2026-06-09 at 09 19 08


src/Slices/Facts/UI/FactsTablePresenter.ts
```diff
@@ -16,14 +13,11 @@ export class FactsTablePresenter implements TablePresenter<Fact, Fact> {
...
       { displayName: words("facts.column.resourceId"), apiName: "resource_id" },
     ];
-    this.numberOfColumns = this.columnHeads.length + 1;
+    this.numberOfColumns = this.columnHeads.length;
   }
 
   createRows(sourceData: Fact[]): Fact[] {
-    return sourceData.map((fact) => ({
-      ...fact,
-      updated: fact.updated ? datePresenter.getFull(fact.updated) : undefined,
```
LukasStordeur, line 25: Are you certain you can remove that?

src/Slices/Facts/UI/FactsRow.tsx
```diff
@@ -35,6 +55,23 @@ export const FactsRow: React.FC<Props> = ({ row }) => {
           </Link>
         </Td>
       </Tr>
+      {valueIsJson && (
+        <Tr isExpanded={isExpanded}>
+          <Td colSpan={numberOfColumns}>
```
LukasStordeur: I think I know what you're trying to do, but I don't see a visual difference.

src/Slices/Facts/UI/FactsRow.tsx
```diff
@@ -35,6 +55,23 @@ export const FactsRow: React.FC<Props> = ({ row }) => {
...
+            <ExpandableRowContent>
+              <CodeBlock
```
LukasStordeur: Did you ask Guillaume what sorts of value we can expect?
if xml or python is also an option besides Json, we might want to apply the same logic as in the AttributeList component.

LukasStordeur (changes requested):
I don't see the chevron to expand/collapse the row back. It might be intuitive to you to close the row when clicking back on "View Value". 
https://www.patternfly.org/components/table/design-guidelines#expansion
They do show the chevron in the guidelines, although admitedly, they do not explecitly mention it is required.

src/Slices/Facts/UI/FactsTablePresenter.ts
AronH99: Yes because the DateWithToolTip already handles this logic internally, and if updated is undefined or null it is also catched by the current code

src/Slices/Facts/UI/FactsRow.tsx
AronH99: Good point, I'll ask

src/Slices/Facts/UI/FactsRow.tsx
AronH99: numberOfColumns part?
Is mainly for the CodeBlock to take up the whole width of the Table otherwise it only takes a small portion of it.
Or do you mean something else?

AronH99:
I addressed your comments now just waiting for Guillaume to get back to us about the possible implementation of the xml, python block.

AronH99:
So what happened:

I essentially replaced the CodeBlock with CodeEditor now to try and keep the custom code minimal and to remain consistent with the other implementation of the CodeEditors in the AttributesList.
I think this is quite an elegant solution, asked Guillaume for feedback and he thought it looked good as well (shared same screenshots on slack FE).
Normally all his feedback has been processed (raw copy value was good enough for him).
Let me know what you think?

Screenshot 2026-06-10 at 11 58 03

Screenshot 2026-06-10 at 12 17 53

src/Slices/Facts/UI/FactsRow.tsx
```diff
@@ -35,6 +66,24 @@ export const FactsRow: React.FC<Props> = ({ row }) => {
...
+            <ExpandableRowContent>
+              <CodeEditor
+                isReadOnly
+                isDarkTheme={isDark}
+                code={getCodeContent(row.value, valueType)}
+                language={LANGUAGE_MAP[valueType]}
+                isLanguageLabelVisible
+                isDownloadEnabled
+                customControls={<CodeEditorCopyControl code={row.value} />}
+                height="200px"
```
LukasStordeur: You might want to use the `getDefaultHeightEditor`. (Once the pr to resolve PF's issue is merged, you should automatically get the proper height capping at 300. )
That will prevent showing too much whitespace for smaller objects :D

src/Slices/Facts/UI/helpers.ts
LukasStordeur, line 1: The `AttributeClassifier`class is already doing a lot of what you're doing in this file. The `classifyKeyValue` does all the checks to know what type of string it is. 

Maybe even reuse the `AttributesList`? It should be capable of working with a single record item, no? That would keep the visuals consistent too, with the same action buttons on the code editor

src/Slices/Facts/UI/helpers.ts
```diff
@@ -0,0 +1,147 @@
...
+ */
+export function getValuePreview(value: string): string {
+  return value.length > VALUE_PREVIEW_LENGTH ? `${value.slice(0, VALUE_PREVIEW_LENGTH)}…` : value;
+}
+
+/**
+ * Maps each expandable `ValueType` to the corresponding PatternFly
+ * `CodeEditor` language for syntax highlighting.
+ */
+export const LANGUAGE_MAP: Record<Exclude<ValueType, "plain">, Language> = {
```
LukasStordeur: using `Language[valueType]` in the factsRow Editor seems just as good? This feels redundant in my opinion.

src/UI/words.tsx
```diff
@@ -823,6 +823,7 @@ const dict = {
   "facts.column.value": "Value",
   "facts.column.resourceId": "Resource Id",
   "facts.empty.message": "No facts found",
+  "facts.value.viewButton": "View Value",
```
LukasStordeur: I think you removed that one :D

LukasStordeur (changes requested):

AronH99:
@LukasStordeur 
Final result of what happened:

1. Refactored the code to use the AttributeClassifier for the fact rows.
I modified AttributeClassifier itself only slightly to cover for the string values being JSON and some other very minor tweaks but in general it's the same logic which is in use for both cases (attributes + facts) of the web console.

2. Made a generic ReadOnlyCodeEditor which sounds exactly like what it does, did some refactoring on this to use the same component for both cases (attributes + facts).
I know this was slightly out of scope however the flow has been simplified by doing so, and is streamlined for future implementations related to this. 
Thought this was a really good change.
Slight note though, I dropped the python case since that was one which was not at all in use anywhere in the application (might have been at some point).
I cleaned most of that part because I asked around and both Guillaume and Sander said they wouldn't use it like that and it would even be a safety concern/weird.  

3.Adjusted unit tests wherever necessary.

So yeah let me know what you think, give your feedback/concerns/questions but overall this seems to be the right direction to go in concerning scalability/maintenance.
I also manually tested the implementation of the attributes by comparing it to mgmt console and the display of the attributes seemed exactly the same so that should be covered too.

Some screenshots of how it looks now:
Screenshot 2026-06-12 at 15 43 47

Screenshot 2026-06-12 at 15 48 21

src/UI/Components/ReadOnlyCodeEditor/ReadOnlyCodeEditor.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+}));
+
+const plainValue = "const answer = 42;";
+
+afterEach(() => {
+  vi.clearAllMocks();
+});
+
+describe("ReadOnlyCodeEditor", () => {
+  test("GIVEN a value WHEN rendered THEN shows it with copy, download, and expand controls", () => {
```
LukasStordeur: I'm tempted to say those are redundant for the most part, since `CodeEditorCopyControl` and `CodeEditorHeightToggleControl` each are covered in `CodeEditorControls.test.tsx`. 
Duplicating tests adds extra maintenance when we need to make changes, so I'd keep them contained.

src/UI/Components/ReadOnlyCodeEditor/ReadOnlyCodeEditor.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    // The component shows the value as-is; it does not reformat.
+    expect(screen.getByText((content) => content.includes('"a": 1'))).toBeVisible();
+
+    fireEvent.click(screen.getByRole("button", { name: words("copy") }));
+
+    expect(copy).toHaveBeenCalledWith(formatted);
+  });
+
+  test("GIVEN a language WHEN rendered THEN the language label is shown", () => {
```
LukasStordeur: Additionally, I'd argue that this is where I'd pull the line between things we need to test, and things that should be blackboxed and Patternfly's responsibility to test. We can't cover all the inner workings of a third party library, it would add too much tests that are too susceptible to change. You could be mocking a PotatoComponent that renders an element with that test-id, it would pass, and yet, it won't not prove the third party component is behaving as expected.

src/UI/Components/ReadOnlyCodeEditor/ReadOnlyCodeEditor.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Read-only code editor with dark-theme, syntax highlighting, download, copy,
+ * and an expand/collapse height toggle. Shows `value` verbatim — callers format
+ * it. Pass `language` for highlighting and a label; omit it for generic content.
+ *
+ * @note Uses CodeEditorCopyControl (copy-to-clipboard) instead of the editor's
+ * built-in copy, which relies on navigator.clipboard and breaks on http origins.
+ */
+export const ReadOnlyCodeEditor: React.FC<Props> = ({ value, language }) => {
```
LukasStordeur, line 20: I do see the added value of a CodeEditor with all the presets we generally use, such as theme, height and expansion. I wouldn't restrict it to only read-only, since that's a simple enough attribute we can propagate. 
Maybe for a follow-up ticket, I think it would be a nice improvement to push it one step further, and have other places use this preset component elsewhere.

src/UI/Components/DesiredStateAttributes/AttributeClassifier.ts
```diff
@@ -38,6 +39,21 @@ export class AttributeClassifier {
       .sort((a, b) => (a.key < b.key ? -1 : 1));
   }
 
+  /**
+   * Detects the code-editor language of a single raw string value by running it
```
LukasStordeur:
```suggestion
   * Detects the language of a single raw string value by running it
```
I'd keep things loose coupled. This class purpose is to classify attributes, regardless of it being used later for a CodeEditor or something else. I'd simply return the plain string value if you really need a dedicated method to extract it. 

You're retargetting this `detectLanguage`  with a wrapping method in your helpers.ts in the facts table. That's the perfect place to bridge the two parts. (Arguably, the location of the wrapper methods could be moved later on the the generic CodeEditor, since it would be used in both the DesiredState Details, and the Facts, and who knows where next.)

src/UI/Components/DesiredStateAttributes/AttributeClassifier.ts
```diff
@@ -98,6 +122,27 @@ export class AttributeClassifier {
     };
   }
 
+  /**
+   * Maps a classified kind to the code-editor language it should render as.
+   *
+   * @param kind - The classified attribute kind (or undefined when nothing matched).
+   * @returns A {@link Language}, `undefined` for generic code, or `null` for inline text.
+   */
+  private kindToLanguage(
```
LukasStordeur: Here too, I don't think it's the classifier's responsibility to deal with Editor specific parts.

src/UI/Components/DesiredStateAttributes/AttributeClassifier.ts
```diff
@@ -98,6 +122,27 @@ export class AttributeClassifier {
     };
   }
 
+  /**
+   * Maps a classified kind to the code-editor language it should render as.
+   *
+   * @param kind - The classified attribute kind (or undefined when nothing matched).
+   * @returns A {@link Language}, `undefined` for generic code, or `null` for inline text.
```
LukasStordeur: I'm not a fan to return null for a, or undefined for b kind of logic. It obscures the meaning down the line, and forces you to go on a search hunt to understand what null or undefined means for this specific case.

 As a general rule of thumb, they should mean: 
- `undefined` is something that hasn't been set yet. (can also be manually set for initialisations, but still would mean the same)
- `null` something that we are purposefully assigning to indicate it has no real value.

src/UI/Components/DesiredStateAttributes/AttributeClassifier.ts
```diff
@@ -159,14 +220,25 @@ export class AttributeClassifier {
...
   private isXml(value: string): boolean {
-    const trimmedValue = value.trimStart().trimEnd();
+    const trimmed = value.trim();
 
-    return trimmedValue.startsWith("<") && trimmedValue.endsWith(">");
+    if (!trimmed.startsWith("<")) {
+      return false;
+    }
+    try {
+      const doc = new DOMParser().parseFromString(trimmed, "application/xml");
```
LukasStordeur: While JSON is generally safe to parse, xml or html, are not exactly safe. It's true that the older method wasn't bulletproof, and would still try to render xml that has mistakes in it, but using DOMParser might open us to XSS vulnerabilities for a very minor improvement. 

Additionally, the size of the xml strings can be considerable. As in, couple of thousand lines for certain configs. 
Trimming the start and end is cheap, I'm not certain parsing it just to cover malformed XML is worth it.

src/UI/Components/DesiredStateAttributes/AttributeList.tsx
```diff
@@ -21,6 +20,12 @@ interface Props {
 
 type AttributeTextVariant = "default" | "monospace";
 
+/** Maps a code-editor attribute kind to its highlighting language (generic `Code` has none). */
```
LukasStordeur: in this case I'd add `Code: undefined` explicitly. Always prefer explicit over implicit.

LukasStordeur (changes requested):
I'd also restructure the folders a bit, the `AttributesLis`t is no longer dedicated to the `DesiredStateAttributes`. So I would move it to the root Components folder. 
Some helpers in the Facts folder are also generic.

I'm under the impression the tab spacing of the JSON is now bigger? 

The other places using the classifier could reuse the `getClassifier` method I believe too. Either make a ticket for the follow-up part if you want to split it up, or tackle it here.

I also believe you initially mentionned liking the structure with helper, and type files. I'd keep that consistent, where methods would be in helpers.ts, and anything type/consts related would go in a type.ts file. 

Last question, shouldn't text values, single/multiline, also have the option to be copied? 

I also remember Guillaume mentionning he'd prefer a copy button to be available in the row without needing to expand it first, was it consiously discarded? (Either is fine for me. Just want to confirm.)

src/UI/Components/ReadOnlyCodeEditor/ReadOnlyCodeEditor.tsx
AronH99: Let's keep this for a separate ticket for now since this would probably take a lot more work to make things consistent over all of the Editors with more risk of unintentional bugs.

https://github.com/inmanta/web-console/issues/6998

AronH99:
So what happened:

- Moved the AttributeClassifier and it's data/helpers/test to the common folder under data
- Refactored throughout the entire console the use of the AttributeClassifier to now come from the same place (common folder under data).
- AttributeList and AttributeValue (child component) are now shared UI components.
- FactsRow now makes use of that AttributeValue, same with the expanded row, also makes use of that AttributeValue, both only accept attributes which are converted/handled by the AttributeClassifier
- Duplicate unit tests should be removed

=> In general this refactor now makes use of both the AttributeList/Value components everywhere in the application and looks like this on the facts page:

Screenshot 2026-06-17 at 13 21 09

With this all of the concerns you had should be addressed and everything should be consistent except for the extra ticket I made about the CodeEditor.

src/Data/Common/AttributeClassifier/AttributeClassifier.spec.ts
```diff
@@ -9,28 +10,6 @@ test("GIVEN AttributeClassifier WHEN provided with a mixed attributes object THE
 });
 
 test("GIVEN AttributeClassifier WHEN provided with a custom multiline classifier THEN returns the correct list of ClassifiedAttributes", () => {
```
LukasStordeur: The title of the test doesn't seem correct anymore

src/Data/Common/AttributeClassifier/AttributeClassifier.ts
```diff
@@ -159,14 +183,17 @@ export class AttributeClassifier {
...
+   * than parsing: XML payloads can be thousands of lines, and running untrusted
+   * markup through DOMParser risks XSS for a marginal gain in precision.
    *
    * @param value - String to check
-   * @returns True if the string appears to be XML (starts with < and ends with >)
+   * @returns True if the trimmed string starts with `<` and ends with `>`
    */
   private isXml(value: string): boolean {
-    const trimmedValue = value.trimStart().trimEnd();
+    const trimmed = value.trim();
```
LukasStordeur, line 209: Don't forget the trimEnd too. It can happen that an extra whitespace can be there after the >.

src/Data/Common/AttributeClassifier/Mock.ts
```diff
@@ -0,0 +1,63 @@
+import { JsonFormatter } from "../JsonFormatter";
+import { XmlFormatter } from "../XmlFormatter";
+import { ClassifiedAttribute } from "./ClassifiedAttribute";
+
+const json = new JsonFormatter();
+const xml = new XmlFormatter();
+
+export const attributes = {
```
LukasStordeur, line 8: This is mock data, I think the file name Data.ts can lead to confusion. We generally use Mock.ts I believe in other places

src/Data/Common/AttributeClassifier/helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+ * facts, resource attributes, logs and discovery all classify values identically
+ * instead of each re-constructing the classifier inline.
+ *
+ * @param options.includeAllKeys - When `true`, no key is filtered out. Defaults
+ *   to `false`, which ignores the `version` and `requires` keys (the class default).
+ */
+export function createAttributeClassifier({
+  includeAllKeys = false,
+}: { includeAllKeys?: boolean } = {}): AttributeClassifier {
+  return new AttributeClassifier(
```
LukasStordeur: Why aren't those arguments handled within the class itself if they're the same everywhere else, and just have the includeAllKeys needed in the constructor. That way, you can omit the need of a wrapper createClass method, and just use `new AttributeClassifier()`

src/Slices/Facts/UI/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+let classifier: ReturnType<typeof createAttributeClassifier> | undefined;
+
+/**
+ * Classifies a single raw fact value via the shared classifier. The resulting
+ * {@link ClassifiedAttribute} `kind` decides how the row renders: structured and
+ * generic "code" kinds expand into the editor, everything else shows inline.
+ *
+ * @param value - The raw fact value string.
+ */
+export function classifyValue(value: string): ClassifiedAttribute {
```
LukasStordeur: This should move to the Common folder I believe. It's not specific to the Facts page itself.
You can have a `classifyRecord`, and a `classifyValue` for example.

src/Slices/Facts/UI/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * Returns `true` when a fact value should expand into the code editor (structured
+ * or generic "code") rather than render inline as plain text.
+ *
+ * @param value - The raw fact value string.
+ */
+export function isExpandableValue(value: string): boolean {
+  return isEditorKind(classifyValue(value).kind);
```
LukasStordeur: Do we really need to reclasiffy each value for this check? Shouldn't `isExpandableValue` use a classified value as argument?

src/Slices/Facts/UI/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Returns a truncated preview of a value. Values longer than
+ * `VALUE_PREVIEW_LENGTH` are clipped and suffixed with `…`.
+ *
+ * @param value - The raw fact value string.
+ * @returns The full value if short enough, otherwise the first
+ *   `VALUE_PREVIEW_LENGTH` characters followed by `…`.
+ */
+export function getValuePreview(value: string): string {
+  return value.length > VALUE_PREVIEW_LENGTH ? `${value.slice(0, VALUE_PREVIEW_LENGTH)}…` : value;
```
LukasStordeur: There is a PF component that would automatically truncate https://www.patternfly.org/components/truncate

src/Slices/Facts/UI/Page.test.tsx
```diff
@@ -11,8 +11,41 @@ import { testClient } from "@/Test/Utils/react-query-setup";
 import { words } from "@/UI";
 import { TestMemoryRouter } from "@/UI/Routing/TestMemoryRouter";
 import { Mock } from "@S/Facts/Test";
+import { VALUE_PREVIEW_LENGTH } from "./helpers";
 import { FactsPage } from ".";
 
+vi.mock("@patternfly/react-code-editor", () => ({
+  CodeEditor: ({
```
LukasStordeur: We do have a default mock for the code editor in the test-setup.ts (root folder), which is I believe imported by default vitest. I think that should be enough for the tests you need to add for coverage. (And that would prevent maintaining multiple code-editor mocks)
We need to test our own implementations primarily, and avoid testing inner PF workings. 

Since we are mocking most of it, we are testing if our mock is doing the job, not if PF is properly displaying a copy button here or there. So the test is a false positive, and doesn't add value.

src/UI/Components/ReadOnlyCodeEditor/helpers.test.ts
```diff
@@ -0,0 +1,47 @@
+import { Language } from "@patternfly/react-code-editor";
+import { getDefaultHeightEditor, isEditorKind, languageForKind } from "./helpers";
+
+vi.mock("@patternfly/react-code-editor", () => ({
```
LukasStordeur: same remark here, there should be a global mock available, and that one contains all the languages too.

LukasStordeur (changes requested):
One minor observation, and I suspect Guillaume will ask about it, the copy in the row is only available for text values, but I think he'll want it on all.

AronH99:
> One minor observation, and I suspect Guillaume will ask about it, the copy in the row is only available for text values, but I think he'll want it on all.

I get what you are aiming at with the copying button but let's just keep it this way for now.
It also doesn't really make sense to have a copy button and a click open button in same table cell.
We'll process any further feedback regarding that later on.

src/Data/Common/AttributeClassifier/AttributeClassifier.ts
AronH99: trim is both start and end, so that should be the same

LukasStordeur: right, my bad, misread it.

LukasStordeur (changes requested):

src/Data/Common/AttributeClassifier/helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+ * JSON/XML formatters wired in, and multiline, non-structured strings treated as
+ * generic "code" (shown in the editor without highlighting). Centralised so
+ * facts, resource attributes, logs and discovery all classify values identically
+ * instead of each re-constructing the classifier inline.
+ *
+ * @param options.includeAllKeys - When `true`, no key is filtered out. Defaults
+ *   to `false`, which ignores the `version` and `requires` keys (the class default).
+ */
+export function createAttributeClassifier({
+  includeAllKeys = false,
```
LukasStordeur: 🤖 Claude · [Blocking]: Open thread from Lukas still unresolved in the code. `createAttributeClassifier` always wires in the same `JsonFormatter`, `XmlFormatter`, and `Code` fallback — those are constants, not configuration. Per the thread's suggestion, they should be defaults inside `AttributeClassifier`'s constructor, so callers can just write `new AttributeClassifier()` (or `new AttributeClassifier({ includeAllKeys: true })`). The factory wrapper adds an indirection and an extra import without buying anything; folding the defaults into the class makes the API self-contained.

src/Slices/Facts/UI/helpers.ts
LukasStordeur: 🤖 Claude · [Blocking]: Open thread from Lukas still unresolved in the code. `classifyValue` (and `isExpandableValue` below) just wrap `createAttributeClassifier().classify(...)` — pure data logic with zero dependency on anything Facts-specific. Per the thread, this belongs in `src/Data/Common/AttributeClassifier/helpers.ts` alongside `createAttributeClassifier`, perhaps as `classifyValue` / `classifyRecord`. Leaving generic code in a feature slice hides it from other slices that may want the same thing, and is a clear placement convention violation.

src/Data/Common/JsonFormatter.ts
```diff
@@ -2,6 +2,6 @@ import { Formatter } from "@/Core";
 
 export class JsonFormatter implements Formatter<unknown> {
   format(source: unknown): string {
-    return JSON.stringify(source, null, 4);
+    return JSON.stringify(source, null, 2);
```
LukasStordeur, line 5: 🤖 Claude · [Should-fix]: The indentation change from 4 → 2 spaces (same in `XmlFormatter`) is silent but affects **every** place that renders attributes across the app — DesiredState details, ResourceDetails, HistoryTab, LogTab, ResourceDiscovery, and now Facts. That's a broad visual change that isn't mentioned in the PR description or the linked issue. Worth an explicit note confirming this was a deliberate product/design choice so reviewers don't miss it.

src/Slices/Facts/UI/Page.test.tsx
LukasStordeur: 🤖 Claude · [Should-fix]: Lukas's thread ("use the global code-editor mock from `test-setup.ts`") was marked resolved, but this local `vi.mock` is still here. If the global mock renders `code` as text (so `findByText(content => content.includes(...))` works without an override), this block should be removed. If it doesn't, the thread was resolved prematurely and the local mock is doing real work — either way, worth confirming before merge.

src/UI/Components/ReadOnlyCodeEditor/helpers.test.ts
LukasStordeur: 🤖 Claude · [Should-fix]: Same point as `Page.test.tsx` — Lukas flagged this ("global mock includes all languages") and the thread was marked resolved, but the local `vi.mock` is still present. If the global `@patternfly/react-code-editor` mock in `test-setup.ts` exports `Language`, this block is redundant and can go.

src/Slices/Facts/UI/FactsRow.tsx
```diff
@@ -1,27 +1,54 @@
...
+          />
+        )}
         <Td dataLabel={words("facts.column.name")}>{row.name}</Td>
         <Td dataLabel={words("facts.column.updated")}>
           {row.updated && <DateWithTooltip timestamp={row.updated} />}
         </Td>
         <Td modifier="breakWord" dataLabel={words("facts.column.value")}>
-          {row.value}
+          {isExpandable ? (
+            <Button variant="link" isInline onClick={() => setIsExpanded((prev) => !prev)}>
```
LukasStordeur: 🤖 Claude · [Nit]: This button toggles `isExpanded` but carries no `aria-expanded` attribute. The PF `<Td expand={...}>` chevron does communicate the expanded state correctly, but users activating this text button get no state feedback from it. Adding `aria-expanded={isExpanded}` here keeps both toggle controls consistent for assistive technology.

src/Slices/Facts/UI/FactsTable.tsx
```diff
@@ -5,21 +5,25 @@ import { SortKey } from "@/Slices/Facts/Core/Types";
...
 export const FactsTable: React.FC<Props> = ({ rows, tablePresenter, sort, setSort, ...props }) => {
-  const onSort: OnSort = (event, index, order) => {
+  const onSort: OnSort = (_event, index, order) => {
     setSort({
       name: tablePresenter.getColumnNameForIndex(index) as SortKey,
       order,
     });
   };
   const activeSortIndex = tablePresenter.getIndexForColumnName(sort.name);
+  const hasExpandableRows = rows.some((row) => isExpandableValue(row.value));
```
LukasStordeur: 🤖 Claude · [Nit]: `isExpandableValue(row.value)` classifies each value via the lazy singleton, and then `FactsRow` classifies the same value again on render — so each row gets classified twice. The singleton is cheap enough that it's fine at current scale; just noting it in case the table grows large. If `classifyValue` moves to Common (per the open thread), one option would be to precompute the classification here and pass it down, but that's a bigger shape change — leave it as-is if the double-call isn't a real concern.

LukasStordeur (changes requested):
## 🤖 Claude pre-review — PR #6965 — feat(facts): expand JSON values inline with formatted code block

**Context map**

| | |
|---|---|
| **Type** | feat + refactor |
| **Goal** | Facts table rows with JSON/XML/long values now expand inline into a read-only code editor; shared classifier + editor infrastructure extracted along the way |
| **Surface area** | `src/Slices/Facts/` (main feature) · `src/Data/Common/AttributeClassifier/` (moved + extended) · `src/UI/Components/AttributeList/` + `ReadOnlyCodeEditor/` (new shared) · `JsonFormatter` + `XmlFormatter` (indentation changed) · 5 caller sites cleaned up |
| **Size** | +722 / −524, 36 files |
| **Test delta** | Positive — new helper unit tests, new Page expand coverage, AttributeList test migrated |
| **Blast radius** | High on formatters (indentation change silently affects every attribute view) · Medium on classifier (all callers updated) · Low on Facts feature itself |
| **Risk hotspots** | Formatter indentation change; two open reviewer threads still unaddressed in the code; local CodeEditor mock duplication |

**Verdict: Request changes**

Two open threads from Lukas remain unaddressed in the code: `createAttributeClassifier` should fold its fixed args into `AttributeClassifier`'s constructor (clean API), and `classifyValue`/`isExpandableValue` belong in `Common` rather than the Facts slice (generic code in a feature folder). Both are clear convention violations per the repo's standards and need to be resolved before merge.

2 blocking, 3 should-fix, 2 nits inline below.

AronH99:
So what happened:

1) After the slack thread I posted, it seemed like I could cleanup the multi line logic for the attributes so I did go ahead and did that (+test cleanup).
Everything should be consistent now and go through the AttributeList/AttributeClassifier. 
Those are cleaned up as well to simplify them.

2) Visually nothing really has changed except for the fact that json and xml now format 2 spaces instead of 4 
=> looks better everywhere

Generally speaking that's mostly it, nothing really big, just cleanup and simplification and some small feedback stuff addressed.

src/Slices/Facts/UI/FactsRow.tsx
LukasStordeur: Can you add a response when you're resolving a comment without applying changes? I can't tell if it's a willing ommision or not :)

src/Slices/Facts/UI/FactsTable.tsx
```diff
@@ -1,25 +1,36 @@
...
 export const FactsTable: React.FC<Props> = ({ rows, tablePresenter, sort, setSort, ...props }) => {
-  const onSort: OnSort = (event, index, order) => {
+  const onSort: OnSort = (_event, index, order) => {
     setSort({
       name: tablePresenter.getColumnNameForIndex(index) as SortKey,
       order,
     });
   };
   const activeSortIndex = tablePresenter.getIndexForColumnName(sort.name);
+  const classifiedRows = rows.map((row) => ({
```
LukasStordeur: Might be worth memoizing this, or it will be recalculated every 5s with the polling we do for the data.

LukasStordeur (changes requested):

src/Slices/Facts/UI/FactsRow.tsx
AronH99: sorry placed it wrongly, should be fixed now

src/Slices/Facts/UI/FactsTable.tsx
AronH99: Maybe a good candidate to test my performance skill on

LukasStordeur (approved):

AronH99:
@LukasStordeur 
Hmm yes so the conclusion is that this will most likely only have a minor impact overall.
However when the facts table is a little bigger and especially the rows where the value contains a lot of json or xml it can have more impact so as best practice Claude definitely recommended implementing the useMemo + memoization of the rows (same pattern as resources table now).
I also disabled the tooltip of the Truncate component since that would be really messy whenever we have medium to long json/xml, and not necessary either because we show it in code blocks anyway

Issue #6935, opened 2026-06-03
[REQUEST] - Environment Switcher


LukasStordeur:
### Summary

Replace the HOME page with an environment Switcher. 

### Problem / Use Case

The environment overview page is rarely used, and takes too much space for what it is.

### Proposed Solution
Improve the menu dropdown to contain a search field that filters the environment in the list based on name match or project. 
Add the logo of each environment to the options.
Remove the Home page, and make Dashboard the default landing page. When no environment is selected, and no existing environment is available, we should redirect to the Create Environement. 
Add the option to create a new environment.

### Impact

ISO-10 only

### Additional Context

Follow-Up task: Add Project Management menu option.

PR #6971, opened 2026-06-09
Feature/6935 environment switcher

LukasStordeur:
# Description

  The Environment overview page has been removed, and the Environment Dashboard will now be the default when loading the application. 
  The last selected environment is stored in local storage so the application will by default load that environmnent when the id is not provided in the URL. 
  When no environment is available to load, the user will be redirected to the create environment page.
  The main menu in the application header now also contains a search box to quickly find and switch to an environment.

image

closes:
- #6935

I ran all the e2e test suits manually, OSS, Local-auth, Oidc, and keycloak


cypress/e2e/scenario-8-composer.cy.js
```diff
@@ -7,15 +7,15 @@
...
+const { checkStatusCompile, selectEnvironment } = environmentHelpers;
 
 const isIso = Cypress.expose("edition") === "iso";
 
 if (isIso) {
   describe("Scenario 8 Composer", () => {
     before(() => {
       // Setup: Ensure a clean test state for this scenario.
-      const envName = "PXSDC Test Env";
+      const envName = "Test PXSDC";
```
AronH99: Small thing, maybe just put this as a const on top to share it between your tests, that way if it changes again you only need to adjust 1 time instead of 3?

cypress/e2e/scenario-8-composer.cy.js
```diff
@@ -7,15 +7,15 @@
...
       const projectName = "PXSDC Test Project";
```
AronH99: This one has it as well I saw but that was the case earlier as well..

cypress/e2e/scenario-9-orders.cy.js
```diff
@@ -1,19 +1,19 @@
...
 
 const isIso = Cypress.expose("edition") === "iso";
 
 if (isIso) {
   describe("Scenario 9 : Order Overview", () => {
     before(() => {
-      clearEnvironment();
-      forceUpdateEnvironment();
+      clearEnvironment("test (lsm-frontend)");
+      forceUpdateEnvironment("test (lsm-frontend)");
```
AronH99: same here, as a const on top?

src/UI/Components/DarkmodeOption/DarkmodeOption.tsx
```diff
@@ -15,7 +16,11 @@ export const DarkmodeOption: React.FC = () => {
   };
 
   return (
-    <DropdownItem key="darkmode-toggle" onClick={toggleTheme}>
+    <DropdownItem
```
AronH99: Something claude picked up on:
We render a DropdownItem here whilst it really should be a MenuItem since it's rendered inside of a MenuList, this is something small which mainly would impact accessibility but it's a valid concern though

src/UI/Root/Components/Header/EnvSelector/EnvSelector.tsx
```diff
@@ -44,77 +69,160 @@ export const EnvSelector: React.FC<Props> = ({
...
+              {words("userManagement.title")}
+            </MenuItem>
+          )}
+          {showLogout && (
+            <MenuItem onClick={() => authHelper.logout()} icon={<SignOutAltIcon />}>
+              {words("dashboard.logout")}
+            </MenuItem>
+          )}
+          <DarkmodeOption />
+          {authHelper.getUser() && (
```
AronH99, line 175: a shame we can't use theMenuFooter for something like this but I guess this is the best solution with the custom tokens right now..

src/UI/Root/Components/PageFrame/Provider.tsx
AronH99, line 68: We can maybe make this null so we can avoid the string comparisons?

src/UI/Root/Components/PageFrame/Provider.tsx
```diff
@@ -15,17 +17,39 @@ export const Provider: React.FC<React.PropsWithChildren<Props>> = ({
...
+    if (redirectingToCreateEnv) {
+      notifyInfo({
+        title: words("home.noEnvironments.toast.title"),
+        message: words("home.noEnvironments.toast.message"),
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  if (environmentId === words("error.environment.missing")) {
```
AronH99, line 39: That way we can just if (environmentId === null) ...

AronH99 (commented):
Tested the branch, seemed pretty good to me, found no issues other than some small stuff. 
Good work :)

src/UI/Root/Components/Header/EnvSelector/EnvSelector.tsx
LukasStordeur: Yes, I tried, but the footer caused more UI issues, and neededeven more added style to get the whole box a bit grayish and faded text color, without absurd margins
image
It probably works well when you just need a button or don't care about the additional margins

src/UI/Root/Components/PageFrame/Provider.tsx
LukasStordeur: I think in this case, the string is more explicit. 
It allows to finegrain the logic to make a distinction between access reasons, and genuine no environments at all.
We don't want to redirect when the user doesn't have access to the rest. The difference then between null and undefined is not exactly clear when you come back to the code later.
Comparing a string, or null, isn't that different in performance either.

src/UI/Components/DarkmodeOption/DarkmodeOption.tsx
LukasStordeur: Forgot about it, I readjusted the menu to use MenuItems after adding the icons to the darkmode, and missed it in prereviewing my pr. Good catch.

AronH99 (approved):

PR #6972, opened 2026-06-09
Improve stability of the resources e2e tests

AronH99:
# Description

Fixes the flaky e2e tests for the resources view.
Ran these multiple times, should be more reliable now.

EDIT: Passed on the first time so it seems to work properly, we will see I guess...


cypress/e2e/scenario-6-resources.cy.js
```diff
@@ -440,15 +440,25 @@ describe("Scenario 6 : Resources", () => {
...
+      // and finish within one 5 s cycle, making a should("exist") check unreliable on CI.
+      cy.intercept("POST", "/api/v2/graphql", (req) => {
+        if (req.body?.query?.includes("isCompiling")) {
+          req.alias = "envPreviewPoll";
+        }
+      });
       cy.get('[aria-label="Sidebar-Navigation-Item"]').contains("Resources").click();
-      cy.get('[aria-label="CompileReportsIndication"]', { timeout: 30000 }).should("exist");
+      cy.wait("@envPreviewPoll");
+      cy.wait("@envPreviewPoll");
```
LukasStordeur: I don't think this has much value considering you are waiting for 90s right after to make sure the compile is done.
You removed the assertion that was confirming a compile was running, which was the one breaking if I'm not mistaken. So, of course keeping only the next one affirming the compile is not running will pass. It might cause flakes further down the line.

The question is, why did we need to check if the compile was running and done? => We want to make sure the deletion of the instance is finished. 
How do we do confirm that with certitude? We can do that in two ways.
1. We check the api response that fetches the inventory and do a hard check on the instance status. 
2. We confirm the row is gone from the table on the inventory. 

Option 2 is the easiest, and makes an assertion on expected result. Option 1 can be a bit more complicated since you'd need to extract the proper id if you have more than one deleted instance on that service.

Once you confirmed that the table is empty, you can safely assume the instance is deleted. It's a safer check than the compilation status.

cypress/e2e/scenario-6-resources.cy.js
```diff
@@ -516,8 +526,15 @@ describe("Scenario 6 : Resources", () => {
...
+        if (req.body?.query?.includes("GetResources")) {
+          req.alias = "sortResetQuery";
+        }
+      });
       cy.get("button").contains("Type").click();
-      cy.get('[aria-label="ResourcesPage-Success"]').should("be.visible");
+      cy.wait("@sortResetQuery");
       cy.get("#PaginationWidget-top-top-toggle > .pf-v6-c-menu-toggle__text > b:first-of-type", {
         timeout: 20000,
       }).should("have.text", "1 - 20");
```
LukasStordeur, line 520: I think this covers up the issue rather than anything else.
If I translate the reasoning behind the change correctly:

A timeout of 20s wasn't enough, so now, we're adding an additional cy.wait in the middle of it. Adding at most 5s to the total timeout before confirming the UI has updated the numbers in the pagination dropdown. 

So, we need to wait approximately 25 seconds to reset the sort? That doesn't sound reasonable at all.

LukasStordeur (changes requested):

cypress/e2e/scenario-6-resources.cy.js
AronH99: I don't think it is a, bad solution but now that you mention it, I was fixing something which could actually be simplified for sure, thanks

cypress/e2e/scenario-6-resources.cy.js
AronH99: yeah this can be solved in a different/easier way as well, good point
Think I found the flake for this one

cypress/e2e/scenario-6-resources.cy.js
LukasStordeur: Replacing a breaking assertion with extra lines of code that in the end are just awaiting extra time before a 90s timeout assertion, will just bloat the test without real purpose :)

cypress/e2e/scenario-6-resources.cy.js
```diff
@@ -440,15 +440,23 @@ describe("Scenario 6 : Resources", () => {
...
+        }
+      );
 
-      // Intercept the GraphQL resource query — needed because keepPreviousData keeps
-      // stale rows visible while the new request is in flight, so we must wait for
-      // each response before asserting row counts.
-      cy.intercept("POST", "/api/v2/graphql").as("resourcesQuery");
+      // Target GetResources queries only — a generic graphql intercept is consumed by
+      // background env-preview polls, causing cy.wait to resolve against stale data.
+      cy.intercept("POST", "/api/v2/graphql", (req) => {
```
LukasStordeur: You don't need the intercept anymore, no?

cypress/e2e/scenario-6-resources.cy.js
```diff
@@ -440,15 +440,23 @@ describe("Scenario 6 : Resources", () => {
...
 
-      // Navigate to Resources, then wait for the recompile to start and finish
       cy.get('[aria-label="Sidebar-Navigation-Item"]').contains("Resources").click();
-      cy.get('[aria-label="CompileReportsIndication"]', { timeout: 30000 }).should("exist");
-      cy.get('[aria-label="CompileReportsIndication"]', { timeout: 90000 }).should("not.exist");
+      // Once the recompile finishes, "test" resources are orphaned and hidden by the default
+      // Not Orphaned filter. Retries until no value cell starts with "test-".
+      cy.get('[aria-label="Resource Table Row"] td[data-label="Value"]', { timeout: 90000 }).should(
+        ($cells) => {
+          const testCells = $cells.filter((i, el) => el.textContent.trim().startsWith("test-"));
```
LukasStordeur, line 453: This would work well, if we had any previous assertions confirming we had resources starting with "test-"

LukasStordeur (changes requested):

cypress/e2e/scenario-6-resources.cy.js
AronH99: The intercept was already present previously to:
```ts
 cy.get("#orphaned-exclude").click();
      cy.wait("@resourcesQuery");
      expectFilteredMoreThan("initialRowCount");

So that one was already there, just changed it a little

File: cypress/e2e/scenario-6-resources.cy.js (hunk 4 above)
Comment (AronH99) [reply cypress/e2e/scenario-6-resources.cy.js:453]: We have, here:
```ts
// Add a new instance via resource-states service
      cy.get('[aria-label="Sidebar-Navigation-Item"]').contains("Service Catalog").click();
      cy.get("#resource-states").contains("Show inventory").click();
      cy.get("#add-instance-button").click();
      cy.get("#name").type("test");
      cy.get("button").contains("Confirm").click();
      
      
      => so we delete these and that's what we need to confirm, are about 25-26 of them or so

**Comment by AronH99** [comment]
I think I tackled the root cause of the problems now so hopefully this should do it, decided to remove the query stuff after all

**Comment by AronH99** [comment]
All right, was another issue but think I am in the right direction still

File: cypress/e2e/scenario-6-resources.cy.js [hunk 5]
```diff
@@ -439,21 +439,33 @@ describe("Scenario 6 : Resources", () => {
...
+        }
+      );
 
-      // Restore page size to 250 so all resources are visible for accurate counting
+      // Restore page size to 250 so all resources are visible for accurate counting.
       cy.get("#PaginationWidget-top-top-toggle").click();
       cy.get('[data-action="per-page-250"]').click();
-      cy.wait("@resourcesQuery");
+      // keepPreviousData keeps the old page on screen while the new request is in flight,
+      // so retry until the header reads "1 - N of N": all resources rendered on one page.
```
LukasStordeur: I'm not entirly sure I understand how this is more robust than finding the text to be `1 - 250` 🤔 
The 20s timeout would retry the assertion regardless of the added logic to extract the total and recreate the `1 -  250` textContent

AronH99: Because the default is 100 and we know that before the deletion there were more than 100 resources, for example:
1 - 100 of 102 becomes => 1 - 76 of 76 (or something like this).
Then we know oke the deletion has happened and completed properly before we establish another rowCount and move on to the other assertions.

LukasStordeur: Yes, the default is 100, but we just clicked on setting it to 250, no? 
So your check will always pass, no matter how many are removed?

AronH99: Hmm yes I think that's a mistake I should move the 250 per page down and then the check should be useful to see whether or not those resources are deleted properly.
But I guess I can just delete it as well since we only have less than 100 resources left at that point (old leftover then)

AronH99: Deleted

LukasStordeur (approved):

PR #6976, opened 2026-06-11
Display version creation dates in the compare view version selects

### LINKED ISSUE #6958: Add dates to the version dropdown of the attributes comparison of a service instance
Add the deployed time/date behind the version dropdown of the service instance attributes comparison.

Image

AronH99:
# Description

- The version dropdowns in the Service Instance Details compare view now show each version's creation date as a muted secondary text (description) below the version number
- Adjusted tests

closes https://github.com/inmanta/web-console/issues/6958

Screenshot 2026-06-11 at 10 51 01


src/Slices/ServiceInstanceDetails/Utils/AttributeUtils.ts
```diff
@@ -218,18 +218,21 @@ export const sortTreeRows = (
   return sortData(tableData);
 };
 
+// A version available in the history logs, with the timestamp of when it was created.
+export interface VersionInfo {
```
LukasStordeur: Can't you Pick the two properties out of InstanceLog?

src/UI/Components/SearchSelect/SearchSelect.tsx
```diff
@@ -18,17 +18,24 @@ export const TEST_IDS = {
   input: "swm-input",
 };
 
+export interface SearchSelectOption {
```
LukasStordeur: I think we can reuse the SelectOptionProps from patternfly?

LukasStordeur (changes requested):
I'm not sure visually it's looking nice to have the date as a description. I'd personnaly append it to the label of the selectOption instead to have it all on the same line. We'd use up more of the available whitespace too.
SelectOptions can have both value and label, one being used onSelect, and the other only for display purposes.

AronH99:
@LukasStordeur I know, I tried doing that as well, however if you append the timestamp to the value outside of maybe customizing the SearchSelect component even further then the search will also include the formatted timestamp which is something we definitely want to avoid, for this reason I took the approach with the description since that's what was supported nicely by PatternFly.
I can append the "description" part to the option.value without passing it to the description prop of SelectOption but I really think it's less clear/nice looking that way (though this avoids the search part issue).
Let me know what you had in mind.

Screenshot 2026-06-11 at 16 48 29

LukasStordeur:
I'm a bit confused by your explanation about the searchfield. 
There are two distinct properties on a `SelectOption`, `value` and `label`. When you perform a search, you will be filtering on `value`. When you have an `onSelect` event, it should return the full `selectOption` item, which should both contain `value`, and `label`. I don't think it's a bad thing to display the timestamp (so the `label`) in the selected box.

I'm also a bit concerned to see the 6.5 documentation of the select, and see that they don't mention the `description` attribute anywhere, while I do find it back in the type definition from the package itself. They also seem to not suport custom JSX within the `SelectOption`, or at least, their examples don't work with it any longer. So it could mean they are stepping away from that, and will deprecate it in next major release.

AronH99:
@LukasStordeur 

And yes let's clear things up:
What I meant was with the way SearchSelect was before there wouldn't be a way to pass the timestamp without having the filter issues.
Now the way I see it we can do two things:
1. Go with your suggestion, change the options prop of the SearchSelect to accept label and value and use the label only as a display value.
2. Go with my suggested fix and make use of the description, which in the end is kind of the same because SearchSelect needs to be adjusted either way.
The description is mentioned here:
https://www.patternfly.org/components/menus/select/#selectoption
It is also used in the examples.

So I reckon this is kind of a style preference, I think you like the first one and for me the second one seems more clear/straightforward.
But if you would absolutely prefer the 1st suggestion I don't mind going with that, otherwise I can also ask some other people to give feedback on this?

LukasStordeur:
Ok, they have two different pages about the Select... How handy. 

My issue with the `description`, is that it doesn't really look good. There's so much whitespace it feels just off. You can share it to Bart if you want a second opinion on the visual :) 

With the alternative documentation page, I'll adjust my suggestion. The `children` is technically equivalent to the `label` but supports a `React.Node`, so the displayed informaion. And in that, you can style it however you like and append the timestamp after the version, may it be in a more grayish font color, or the spacing, etc. 
`onSelect` will work with the `value`, or whatever you pass into it really, and so should the filtering.

AronH99:
No it's fine I'll go with your suggestion

AronH99:
Screenshot 2026-06-15 at 10 18 37

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesCompare.tsx
```diff
@@ -43,17 +54,24 @@ export const AttributesCompare: React.FC<Props> = ({ instanceLogs, selectedVersi
...
-    }
-  }, [instanceLogs]);
+  const availableVersions: SelectOptionProps[] = useMemo(
+    () =>
+      getAvailableVersions(instanceLogs).map(({ version, timestamp }) => ({
+        value: version,
+        children: (
+          <Flex gap={{ default: "gapSm" }} alignItems={{ default: "alignItemsCenter" }}>
+            {version}
+            <Content style={{ color: "var(--pf-t--global--text--color--subtle)" }}>
```
LukasStordeur: I'm not sure the variable is properly applied, the text color in the screenshot doesn't look faded.

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesCompare.tsx
```diff
@@ -43,17 +54,24 @@ export const AttributesCompare: React.FC<Props> = ({ instanceLogs, selectedVersi
...
-  useEffect(() => {
-    if (instanceLogs && instanceLogs.length) {
-      const versions = getAvailableVersions(instanceLogs);
-
-      setAvailableVersions(versions);
-    }
-  }, [instanceLogs]);
+  const availableVersions: SelectOptionProps[] = useMemo(
+    () =>
+      getAvailableVersions(instanceLogs).map(({ version, timestamp }) => ({
```
LukasStordeur: What's the added bonus of mapping twice over the logs?
getAvailableVersions is a bit redundant in my opinion. You're deconstructing version and timestamp in both places.

LukasStordeur (changes requested):

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesCompare.tsx
AronH99: It is applied properly but really poorly visible, was looking at other colors but didn't seem to find any better fitting ones. 
We can try another color or adjust only the version itself to be more dark?

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesCompare.tsx
AronH99: forgot to cleanup

src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesCompare.tsx
AronH99: Tried darker colors for the version but that didn't seem to be a good fit either

LukasStordeur: --pf-t--color--gray--50 maybe? It's the one used for the subtle icons, it's slightly lighter.

LukasStordeur: That, or we can stick the timestamp in a label with outline? Just an idea

AronH99: Screenshot 2026-06-16 at 11 34 37

We can but that might make it have accessibility issues, unsure if we care a lot about those but it's on the edge

LukasStordeur: Which accessibility concern do you have? For the label, or the color?

AronH99: If the timestamp is too light it clashes a little with the background color, but anyway I tried the label idea and think that looks good?

Screenshot 2026-06-16 at 11 40 02

LukasStordeur: Yes, just maybe add the outline version instead of the plain blue one?

AronH99: Screenshot 2026-06-16 at 11 58 27

LukasStordeur (approved):

Issue #6963, opened 2026-06-08
"Show discovery resource" link doesn't work


edvgui:
Both in the discovered resource details and in the discovered resources table, clicking on the link brings you to the wrong page.  Looks like url re-encoding issue.

Reproducible on mgmt orchestrator for any discovered resource.

PR #7000, opened 2026-06-16
6963 discovery resource uri

LukasStordeur:
# Description

Replace the rid extraction with the resource id now that it is provided. When the page was initially developped, those fields weren't there yet, and we weren't informed after they were added. 

https://github.com/user-attachments/assets/7b4acc9c-71f3-446c-ab4d-4cf4595c6f45

closes : 
- #6963


AronH99 (approved):
pretty straightforward so approved

PR #7017, opened 2026-06-18
Replace Moment with Day.js

LukasStordeur:
# Description

- Replaced moment and moment-timezone (~1.3 MB) with day.js (~2 KB core + plugins), reducing bundle size significantly. The public` DatePresenter` interface is unchanged; only the implementation class was renamed from `MomentDatePresenter` to `CustomDatePresenter`.
- Fixed a pre-existing timezone bug on the main Facts page: `FactsTablePresenter.createRows()` was pre-formatting the updated timestamp with `getFull()`, then passing the result to `<DateWithTooltip>` which re-parsed it as a UTC string and applied the timezone offset a second time. Removed the pre-formatting so the raw ISO timestamp reaches `DateWithTooltip` directly, consistent with every other table in the app.


src/Data/Queries/Slices/Parameters/GetParameters/getUrl.ts
```diff
@@ -17,7 +17,7 @@ import { GetParametersParams } from "./useGetParameters";
  */
 export function getUrl(
   { pageSize, filter, sort, currentPage }: GetParametersParams,
-  timezone = moment.tz.guess()
+  timezone = Intl.DateTimeFormat().resolvedOptions().timeZone
```
AronH99: dayjs.tz.guess() is possible as well

src/Data/Queries/Slices/Resource/GetResourceLogs/getUrl.ts
```diff
@@ -42,8 +42,10 @@ const serializeTimestampFilter = (filter?: DateRange.Type[]): string[] | undefin
 
   return filter.map(
     (timestampWithOperator) =>
-      `${RangeOperator.serializeOperator(timestampWithOperator.operator)}:${moment
-        .tz(timestampWithOperator.date, moment.tz.guess())
+      `${RangeOperator.serializeOperator(timestampWithOperator.operator)}:${dayjs(
+        timestampWithOperator.date
+      )
+        .tz(Intl.DateTimeFormat().resolvedOptions().timeZone)
```
AronH99: same here, some other places have it as well so maybe go through them all?

test-setup.ts
```diff
@@ -100,8 +99,15 @@ const linkShapeMock = vi.hoisted(() => ({
   }),
 }));
 
-// Set default timezone
-moment.tz.setDefault("Europe/Brussels");
+// Set default timezone for tests to Europe/Brussels,
+// matching the previous moment.tz.setDefault("Europe/Brussels") behavior.
+// This ensures MomentDatePresenter picks up Brussels timezone at instantiation time.
+const _originalResolvedOptions = Intl.DateTimeFormat.prototype.resolvedOptions;
```
AronH99:
```ts
vi.spyOn(dayjs.tz, "guess").mockReturnValue("Europe/Brussels");
```

I think my suggestion can be used as well here instead of the current change.

AronH99 (changes requested):
Some minor things, little cleanup tho

src/Data/Queries/Slices/Parameters/GetParameters/getUrl.ts
LukasStordeur: One's the native API available in the browser, the other is a wrapper that uses the same api under the hood. Using the native API has the advantage is less suceptible to be changed in the future. 
We can change it, but they're pretty much the same at the moment.

AronH99: I know, i thought that this change would be a little clearer/more straightforward but if u want to keep it like this, also fine

AronH99 (approved):

PR #7020, opened 2026-06-19
Manage projects feature

### LINKED ISSUE #6936: [REQUEST] Add Project Management menu option.
### Summary

Add the option to manage projects in the main menu dropdown.

### Problem / Use Case

Project cannot be deleted at the moment.

### Proposed Solution(s)

See https://claude.ai/design/p/fa27ecf6-a729-409b-8d07-ba8ca7360c41?via=share&file=Environment+Health+Dashboard.html

### Impact
Only ISO-10

### Additional Context

Should be done after #6935

LukasStordeur:
# Description

Closes #6936

Adds a "Manage projects" option to the environment selector dropdown. Clicking it opens a modal that lists all projects with their environment counts, and allows deletion of projects that have no environments.

### New feature

- `EnvSelector` — new "Manage projects" menu item (folder icon) that opens the modal via `ModalContext`
ManageProjectsModal — modal content built with PatternFly layout components (Split, Stack, Flex); lists all projects fetched via `GET /api/v2/project?environment_details=true`; each row shows the project name, environment count, and a Delete button
- Deletion requires inline confirmation ("Confirm delete" / "Cancel") before the `DELETE /api/v1/project/{id}` request is sent; projects with environments have the Delete button disabled with a lock icon and remaining-environment count
- `useDeleteProject` — new React Query mutation hook that invalidates the projects cache on success

## Supporting changes

- `ModalProvider` — backdrop click now closes any modal that has showClose: true (the default); PF6's Modal.onClose only covers the X button, so this adds a document-level click listener that checks whether the click landed inside .pf-v6-c-backdrop but outside [role="dialog"]. Unlike for the dropdowns, they didn't provide an onOpenChange property to deal with backdrop clicks.

https://github.com/user-attachments/assets/51c20b35-44b0-4509-9902-f2eed8fd2a71


src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
```diff
@@ -392,4 +393,63 @@ describe("EnvironmentSelector", () => {
...
+      })
+    );
+
+    render(
+      <ErrorBoundary>
+        <TestMemoryRouter initialEntries={["/?env=123"]}>
+          <QueryClientProvider client={testClient}>
+            <AuthProvider config={undefined}>
+              <AuthTestWrapper>
+                <ModalProvider>
```
AronH99, line 433: Why not just include ModalProvider in the setup, It should be harmless for the other tests no?
Then you can remove this long setup

src/UI/Root/Components/ModalProvider/ModalProvider.tsx
```diff
@@ -143,6 +143,27 @@ export const ModalProvider: React.FC<PropsWithChildren> = ({ children }) => {
     }
   };
 
+  useEffect(() => {
```
AronH99, line 146: Should we suddenly add a clickOutside to the global implementation of modals?
I am not entirely sure?
=> If we do the pf class selector I really would avoid that if possible

AronH99 (commented):
Have 2 questions more than requests I guess

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
LukasStordeur: Hm, I guess? But maybe that's part of a different refactor, it might affect other tests and involve more changes than expected.

AronH99: What I mean is just the setup in this file, not a global one or anything

src/UI/Root/Components/ModalProvider/ModalProvider.tsx
LukasStordeur: This is the global provider, and you don't need to add anything really. I'm not sure there's another way to add the functionality at the moment. 
This should allow modals to be closed by default when the user clicks outside. I think that's something we'd want? I do personnaly find it a bit annoying when modals don't close when I click on the backdrop

src/UI/Root/Components/Header/EnvSelector/EnvironmentSelector.test.tsx
LukasStordeur: Oh, well, because this test is contained to only the project management, the others not really. It doesn't hurt, but doesn't add much either I think.

src/UI/Root/Components/ModalProvider/ModalProvider.tsx
AronH99: Yes.I know, for me it's fine, I prefer it as well but I was just hoping to avoid the use of:
```ts
".pf-v6-c-backdrop"

AronH99 (approved):
I'll approve already anyway

src/UI/Root/Components/ModalProvider/ModalProvider.tsx
LukasStordeur: I can make a ticket to look into it :D

LukasStordeur: https://github.com/inmanta/web-console/issues/7024

PR #7021, opened 2026-06-19
Replace moment dayjs iso9

LukasStordeur:
# Description

ISO 9 pr for replacing moment with day.js


AronH99:
You are still busy fixing this build before I need to review right?

LukasStordeur:
> You are still busy fixing this build before I need to review right?

Should be good now :) It's the same as for master

AronH99 (approved):

PR #7029, opened 2026-06-22
Add include/exclude icons to resource status toggles and share the component

### LINKED ISSUE #6995: Change the resources filters (purged, isDeployed, orphaned) to have check/cross icons.
# Description
Change the resources filters (purged, isDeployed, orphaned) to have check/cross icons just as the compound state filters in the dropdown are made.

# Notes
- The goal of this is to make the ToggleGroup less ambiguous.
- Also look into the toggle groups active/focus state which is a little confusing (see video).

https://github.com/user-attachments/assets/42dc8310-fdf4-4b51-bceb-a14a71e8407e

AronH99:
# Description

- The orphaned, deploying, and purged resource status filters now render check/cross icons instead of "Include"/"Exclude" text
- OptionalToggleGroup accepts either icon + ariaLabel or just a label to display just icons or just text toggles (this is to accommodate future use)
- Some cleanup where the IncludeExcludeSelect and helpers are moved to the proper folder/files

closes https://github.com/inmanta/web-console/issues/6995

https://github.com/user-attachments/assets/7b62c09f-cb8b-49d4-8a30-42a6994fec37


src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
```diff
@@ -1,9 +1,14 @@
...
 export interface OptionalToggleGroupOption<T extends string = string> {
-  label: string;
   value: T;
   buttonId: string;
+  label?: string;
+  icon?: {
+    active: React.ReactNode;
+    inactive: React.ReactNode;
+  };
+  ariaLabel?: string;
```
AronH99: 🤖 Claude · [Should-fix]: `label`, `icon`, and `ariaLabel` are all independently optional here, so the type permits an option with **no accessible name at all** — e.g. an `icon` with no `ariaLabel` renders an icon-only `ToggleGroupItem` whose icons are `aria-hidden`, leaving the button unlabelled. The call sites are fine today, but the type doesn't enforce the contract the docstring describes. A discriminated union makes the invalid combos unconstructable:

```ts
type OptionalToggleGroupOption<T extends string = string> = {
  value: T;
  buttonId: string;
} & (
  | { label: string; icon?: never; ariaLabel?: never }
  | { icon: IncludeExcludeIconPair; ariaLabel: string; label?: never }
);
```

Bonus: the inline `{ active; inactive }` shape duplicates `IncludeExcludeIconPair` — reuse that named type (see the placement note on it) rather than redeclaring it.

src/UI/Components/IncludeExcludeIcons/IncludeExcludeIcons.tsx
```diff
@@ -0,0 +1,34 @@
+import { CheckCircleIcon, CheckIcon, TimesCircleIcon, TimesIcon } from "@patternfly/react-icons";
+
+/**
+ * Icon pair shown for an include/exclude control. The `active` icon (filled, coloured)
+ * is rendered when the option is selected, the `inactive` icon (outlined, muted) otherwise.
+ */
+export interface IncludeExcludeIconPair {
```
AronH99, line 7: 🤖 Claude · [Should-fix]: `includeIcons` / `excludeIcons` / `IncludeExcludeIconPair` now serve generic `OptionalToggleGroup` consumers (`StatusFilterSelect`, `ResourceFilterForm`) that have nothing to do with `IncludeExcludeSelect` — yet they import this include/exclude vocabulary out of a sibling component's folder. It compiles fine because everything re-exports through `@/UI/Components`, but the folder name is misleading for what's now a shared vocabulary. Consider a neutral home (e.g. its own `IncludeExcludeIcons/` module under `UI/Components`) so both `IncludeExcludeSelect` and the toggle-group call sites import from a name that matches what it is — and so `OptionalToggleGroup` can reuse `IncludeExcludeIconPair` without depending on `IncludeExcludeSelect`. Either way, a quick decision worth making explicit.

src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
```diff
@@ -43,15 +51,21 @@ export const OptionalToggleGroup = <T extends string>({
...
-          onChange={() => handleChange(value)}
-        />
-      ))}
+      {options.map(({ value, buttonId, label, icon, ariaLabel }) => {
+        const isSelected = selected.includes(value);
+
+        return (
+          <ToggleGroupItem
+            key={value}
+            aria-label={label ?? ariaLabel}
```
AronH99: 🤖 Claude · [Nit]: in text mode this sets both `text={label}` and `aria-label={label}`; the visible text already supplies the accessible name, so the `aria-label` is redundant (harmless since they're equal). Could scope it to icon mode — `aria-label={icon ? ariaLabel : undefined}` — but either is fine for me.

src/Test/Utils/index.ts
```diff
@@ -1,3 +1,8 @@
-export function flushPromises(): Promise<void> {
-  return new Promise((resolve) => process.nextTick(resolve));
-}
+export const getById = (id: string): HTMLElement => {
```
AronH99, line 1: 🤖 Claude · [Nit]: this swaps `flushPromises` out for `getById` rather than adding `getById` alongside it. I checked the tree and `flushPromises` has no remaining references, so removing it is safe dead-code cleanup — just flagging it's slightly outside the icon-toggle scope, in case you'd rather keep this PR tight. Fine to leave :)

AronH99 (commented):
## 🤖 Claude pre-review — PR #7029 — Add include/exclude icons to resource status toggles and share the component

**Context map**

- **Type & intent:** feat/refactor (changelog: patch · minor-improvement) — render the orphaned / deploying / purged resource status filters as check/cross icons instead of "Include"/"Exclude" text, and promote `OptionalToggleGroup` to a text-*or*-icon API. Closes #6995.
- **Surface area (grouped):**
  - *Shared `src/UI/Components`*: `OptionalToggleGroup` (new icon|text option shape), new `includeExcludeIcons`, and the whole `IncludeExcludeSelect/*` **moved here from the Resource slice** + barrel export.
  - *Shared `src/Core/Language`*: `invertFilter` / `removeInvertedSelection` **moved here from the slice's `utils.ts`** (+ spec).
  - *Shared test `src/Test/Utils`*: `getById` added, `flushPromises` removed.
  - *Resource slice*: `StatusFilterSelect`, `ResourceFilterForm` switched to icon options; tests re-pointed at `@/Test`.
- **Size & balance:** +172 / −119 across 18 files; test delta is present (OptionalToggleGroup icon test, `Utils.spec`, e2e selectors are id-based). Healthy ratio.
- **Blast radius:** *high-ish* — touches a shared component (`OptionalToggleGroup`), a shared util (`Core/Language/Utils`), and the shared barrel. Mitigated: `scenario-6-resources` selects toggles by `#buttonId`, not by text, so the text→icon swap does **not** break e2e.
- **Risk hotspots:** shared-component API design (icon|text), a11y of icon-only toggles, code placement of the new icon vocabulary.

Behaviour is correct, conventional, and tested, and the icon swap is safe against the id-based e2e selectors — no hard blockers. What remains is **2 should-fix** items on the *shared* surface (enforce the label-XOR-icon contract in the type; reconsider the icon vocabulary's folder home + reuse its named type) and **2 nits**. The typing one is what I'd most want addressed before this shared API spreads. Flips to a clean approve once those two are resolved or consciously declined.

Nice calls worth calling out: moving `invertFilter`/`removeInvertedSelection` and `IncludeExcludeSelect` out of the feature folder into shared, consolidating the duplicated local `getById` into `@/Test`, and a clean, conventional changelog. One thing to confirm before merge: Jenkins/e2e is still **pending** — make sure it goes green (it should, given id-based selectors).

2 should-fix, 2 nits inline below.

AronH99:
Addressed the feedback, think those were some solid adjustments, if tests are green then all should be good to be merged

bartv (commented):
_Posted by Claude on behalf of Bart

## Verdict: Approve with nits

Solid, well-scoped PR. Lint (`--max-warnings=0`), `tsc --noEmit`, and all 35 tests across the 5 affected files pass locally. The discriminated-union API on `OptionalToggleGroup` and the accessibility wiring are genuinely nice work. No merge blockers; the inline comments are quality polish.

**One thing to decide before merge:** the description says `closes #6995`, but that issue has a second ask: _"Also look into the toggle groups active/focus state which is a little confusing (see video)."_ This PR only swaps text for icons (no styling/CSS changes), so the focus-state ambiguity the reporter flagged is untouched. Either address the selected/focus styling here, or change `closes` to a plain reference and keep #6995 open with a tracking note so that ask isn't silently dropped on merge.

**Worth calling out as good work:**
- The `{ label } | { icon + ariaLabel }` union makes invalid prop combinations unrepresentable at the type level; icons are `aria-hidden` with the accessible name carried on the button, and include vs exclude is conveyed by glyph shape + fill, not color alone (colorblind-safe).
- Good shared-component hygiene: `IncludeExcludeSelect` is genuinely domain-agnostic and used 3x, `IncludeExcludeOption` stays private (barrel re-exports only the parent), and `invertFilter`/`removeInvertedSelection` correctly land in `Core/Language/Utils.ts`.

AronH99:
Yes, about the remark of the toggle groups active/focus state:
I wanted to address it but since the icons make it clear whether or not the filters are active or not, this issue isn't really a problem anymore for this kind of implementation.
So I dropped that since it would most likely only add extra custom css which would mean potentially more cleanup work when pf v7 drops.

bartv (approved):

PR #7030, opened 2026-06-22
Fix notification badge flicker on initial load

### LINKED ISSUE #7025: Notification button flickers at load time
Notification button in the top bar flickers at load time.
iso8, 9 and 10

https://github.com/user-attachments/assets/f66a8263-3d91-4a90-857e-0457f0e5458d

AronH99:
# Description

- Render the notification badge enabled in the neutral `read` variant while the first request is in flight, so it no longer flashes from disabled to active on initial load
- Reserve the disabled placeholder badge for the genuine "no environment selected" case
- Update Badge tests: loading now asserts an enabled `read` badge, and variant cases wait for the resolved value

=> this is for iso 8 (most likely will need to make separate pr for this), 9 and 10

closes https://github.com/inmanta/web-console/issues/7025

https://github.com/user-attachments/assets/57beac34-84e1-4212-b247-a218b5e133ea

### DIFF
--- a/changelogs/unreleased/7025-notification-badge-flicker.yml
+++ b/changelogs/unreleased/7025-notification-badge-flicker.yml
@@ -1,6 +1,6 @@
 description: The notification badge no longer flashes from a disabled to an active state on initial load.
 issue-nr: 7025
 change-type: patch
-destination-branches: [master, iso9]
+destination-branches: [master, iso9, iso8]
 sections:
   bugfix: "{{description}}"
--- a/changelogs/unreleased/7025-notification-badge-flicker.yml
+++ b/changelogs/unreleased/7025-notification-badge-flicker.yml
@@ -1,0 +1,6 @@
+description: The notification badge no longer flashes from a disabled to an active state on initial load.
+issue-nr: 7025
+change-type: patch
+destination-branches: [master, iso9]
+sections:
+  bugfix: "{{description}}"
--- a/src/Slices/Notification/UI/Badge/Badge.test.tsx
+++ b/src/Slices/Notification/UI/Badge/Badge.test.tsx
@@ -1,6 +1,6 @@
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { render, screen, waitFor } from "@testing-library/react";
-import { delay, graphql, HttpResponse } from "msw";
+import { graphql, HttpResponse } from "msw";
 import { setupServer } from "msw/node";
 import { NotificationsResponse } from "@/Data/Queries";
 import { MockedDependencyProvider } from "@/Test";
@@ -24,7 +24,7 @@ function setup() {
     </QueryClientProvider>
   );
 
-  return { component };
+  return { component, queryClient };
 }
 const server = setupServer();
 
@@ -61,7 +61,7 @@ describe("Badge", () => {
           });
         })
       );
-      const { component } = setup();
+      const { component, queryClient } = setup();
 
       render(component);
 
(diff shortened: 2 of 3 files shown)


src/Slices/Notification/UI/Badge/Badge.test.tsx
```diff
@@ -71,11 +70,13 @@ describe("Badge", () => {
       });
 
       expect(button).toBeVisible();
-      expect(button).toHaveAttribute("data-variant", variant);
+      // The badge renders immediately in the neutral `read` state while loading, so wait
+      // for it to settle on the variant derived from the resolved notifications.
+      await waitFor(() => expect(button).toHaveAttribute("data-variant", variant));
```
AronH99, line 76: 🤖 Claude · [Nit]: now that the badge renders as `Badge-Success` with `data-variant="read"` *during* loading, this `waitFor` passes immediately for the two `read` rows (`no notifications` and `only read`) — before the server responds — so those two cases no longer actually verify the *resolved* state, just the loading default. The unread/attention rows are still honest (their resolved variant ≠ `read`, so `waitFor` genuinely waits). Minor and fine to leave; if you wanted them airtight you'd assert something that only exists post-resolution for the read rows. The comment + `waitFor` are the right call otherwise :)

changelogs/unreleased/7025-notification-badge-flicker.yml
```diff
@@ -0,0 +1,6 @@
+description: The notification badge no longer flashes from a disabled to an active state on initial load.
+issue-nr: 7025
+change-type: patch
+destination-branches: [master, iso9, iso8]
```
AronH99, line 4: 🤖 Claude · [Question]: the description says this is for iso8, 9 and 10, and this targets `[master, iso9, iso8]`. Recent changelogs target `[master, iso9]`, so I'm reading `master` as the iso10 dev line — i.e. iso10 is already covered here and nothing's missing. Just confirming that mapping is right, and that the iso8 backport is intended and lands cleanly (you added it in the follow-up commit and Jenkins is green, so this looks fine) — the "most likely will need a separate PR" note in the description made me want to double-check. Not blocking.

AronH99 (commented):
## 🤖 Claude pre-review — PR #7030 — Fix notification badge flicker on initial load

**Context map**

- **Type & intent:** fix (changelog: patch · bugfix) — stop the notification badge flashing from *disabled* → *active* on initial load. While the first request is in flight it now renders the enabled, neutral `read` badge instead of the disabled placeholder; the disabled placeholder is reserved for the genuine "no environment selected" case. Closes #7025.
- **Surface area (grouped):** confined to one slice — `src/Slices/Notification/UI/Badge` (`Badge.tsx` render-branch reorder + `Badge.test.tsx`) plus the changelog. No shared layers touched.
- **Size & balance:** +32 / −24 across 3 files; test delta present and matched to the behaviour change. Healthy ratio.
- **Blast radius:** *low* — single feature component, no shared component/hook/type changed. I grepped cypress/e2e for the `Badge-*` aria-labels and found **no** selectors depending on them, so the loading-state relabel (`Badge` → `Badge-Success`) breaks nothing downstream.
- **Risk hotspots:** render-branch ordering correctness; loading-vs-resolved test sensitivity; changelog backport targets.

**Verdict: Approve (with nits)**

Behaviour is correct and minimal, the branch reorder (error → no-env → loading/success) covers every path with no dead code, the inline comments explain the *why* well, and the tests are updated to match. All CI is green (CodeQL + Jenkins SUCCESS). No blockers. Posting as a `COMMENT` review only because GitHub blocks self-approve on your own PR — treat this as an approve-with-nits.

Nice touches: dropping the redundant Fragment wrapper around the error badge, and the comment that justifies keeping the badge enabled during load. Nothing flips this; the two notes below are optional. 0 blocking, 0 should-fix, 1 nit + 1 question inline.

AronH99 (commented):
_🤖 Duplicate of the Claude pre-review above (a double-POST on my end) — its inline comments have been removed. Please read [the canonical review](https://github.com/inmanta/web-console/pull/7030#pullrequestreview-4544959665) instead._

src/Slices/Notification/UI/Badge/Badge.test.tsx
```diff
@@ -71,11 +70,13 @@ describe("Badge", () => {
...
     }
   );
 
-  test("Given Badge WHEN request is loading THEN variant is not shown and badge is disabled", () => {
+  test("Given Badge WHEN request is loading THEN an enabled read badge is shown", () => {
     server.use(
       queryBase.operation(() => {
         delay(100);
```
bartv: nit: `delay(100)` is a no-op here. It returns a Promise that is never awaited or returned, so the handler responds synchronously and the request is not actually held in flight. The test still passes for the right reason (the assertions below are synchronous and capture the first render, where React Query has `data === undefined` regardless of any delay), but the line is misleading about how the loading state is isolated. Either drop it (along with the now-redundant blank line) since the synchronous first-render assertion is what establishes the in-flight state, or make it real with `await delay(100)` inside an `async` handler.

src/Slices/Notification/UI/Badge/Badge.tsx
```diff
@@ -45,31 +45,32 @@ export const Badge: React.FC<{ onClick(): void }> = ({ onClick }) => {
...
-        data-variant={variant}
-        variant={variant}
-        onClick={onClick}
-      />
+      <NotificationBadge aria-label="Badge" variant={NotificationBadgeVariant.read} isDisabled />
     );
   }
 
+  // While the first request is in flight, default to the neutral `read` variant and keep
+  // the badge enabled, so it doesn't flash from disabled to active on initial load.
```
bartv: nit: this branch is also taken when switching environments, not just on initial load. The query keys on `[envID]` with no `placeholderData`/`keepPreviousData`, so a switch produces a fresh entry with `data === undefined` and lands here too. Worth widening the comment so the next reader knows this path covers both cases.

```suggestion
  // While a request is in flight (initial load or after an environment switch), default to the
  // neutral `read` variant and keep the badge enabled, so it doesn't flash from disabled to active.
```

bartv (commented):
Claude here, reviewing on behalf of Bart.

The fix looks correct: the render-branch reorder handles every state properly (including the switch from `data.notifications` to `data?.notifications ?? []`), and tests + lint pass locally. Two small optional nits inline.

AronH99:
Should have addressed all of the possible flakes/issues for the unit tests now

bartv (approved):

PR #7034, opened 2026-06-23
Replace optional boolean form field radios with a True/False toggle group

### LINKED ISSUE #6996: Adapters of service instances should also be using the newly created OptionalToggleGroup instead of the current radio buttons.
# Description
Adapters of service instances should also be using the newly created OptionalToggleGroup instead of the current radio buttons.

# Notes
This is mainly because the new toggle group is a better looking/fitting component than what we currently have

# Example
Image

AronH99:
# Description

- Optional boolean attributes in the service instance form now render as a `[True][False]` toggle group (no selection = null/unset) instead of True/False/null radio buttons, reusing `OptionalToggleGroup`
- Removed the now-unused `BooleanFormInput` component and its test
- Deselecting a toggle blurs the clicked button so it no longer keeps the focused/selected styling (PatternFly toggle-group quirk)
- Added boolean-value and focus coverage to `OptionalToggleGroup` tests; updated the service-instance form/create/edit tests for the new rendering

Closes: https://github.com/inmanta/web-console/issues/6996

https://github.com/user-attachments/assets/b095dfd7-b409-4b89-9275-d135ab7a7b9b


src/Slices/CreateInstance/UI/CreateInstance.test.tsx
```diff
@@ -50,6 +50,16 @@ const getDictValue = (testId: string, container: HTMLElement = document.body) =>
   return JSON.parse(content.textContent ?? "{}");
 };
 
+const getBooleanOption = (
```
AronH99: 🤖 Claude · [Should-fix]: `getBooleanOption` is duplicated verbatim in `EditInstancePage.test.tsx` (identical body). Two copies will drift. Extract it once to a shared test helper (e.g. under `src/Test/Utils/`) and import in both — or generalise to `getToggleOption(groupName, optionLabel, container)` since nothing about it is boolean-specific. Small, but worth doing now while it's two files and not five.

src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
```diff
@@ -25,26 +26,35 @@ export interface OptionalToggleGroupProps<T extends string = string> {
  *    the accessible name).
  *  @prop {string[]} selected - The full list of currently active filter values. Only values matching this group's options are considered.
```
AronH99: 🤖 Claude · [Should-fix]: this docstring is now stale — `@prop {string[]} selected` should read `{T[]}`, and 'currently active filter values' is too narrow now that the component is generic over `string | boolean` and used for a plain boolean form field, not only filters. A quick reword (selected/onChange operate on `T[]`, where `T` is the option value type) keeps it honest given how strict this repo is about docstrings reflecting intent.

src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
```diff
@@ -25,26 +26,35 @@ export interface OptionalToggleGroupProps<T extends string = string> {
...
       ? selected.filter((status) => status !== value)
-      : [...selected.filter((status) => !optionValues.includes(status as T)), value];
+      : [...selected.filter((status) => !optionValues.includes(status)), value];
     onChange(next);
+
+    // This is essentially a workaround for Patternfly's internal behaviour of the ToggleGroup.
+    // We don't want a button to stay active/focused after deselect.
+    // Can possibly be removed in future versions of Patternfly whenever they address this.
+    if (isSelected && target instanceof HTMLElement) {
+      target.blur();
```
AronH99, line 57: 🤖 Claude · [Question]: `target.blur()` on deselect drops keyboard focus to `<body>` (the clicked button is the active element for both mouse and keyboard). For mouse it's clearly right; for a Space/Enter deselect it loses focus position — Tab/Shift+Tab resume roughly where you were, so probably fine, just flagging it as an a11y call. Also worth capturing in the PR *why* blur was chosen over a CSS override of PF's focus background, so the next person doesn't reach for the CSS fix. Not blocking.

src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx
```diff
@@ -115,41 +123,53 @@ export const FieldInput: React.FC<Props> = ({
...
+          <FormHelperText>
+            <HelperText>
+              <HelperTextItem>{field.description}</HelperTextItem>
+            </HelperText>
+          </FormHelperText>
+          <OptionalToggleGroup
+            isDisabled={field.isDisabled && get(originalState, booleanPath) !== undefined && !isNew}
+            selected={booleanValue === null ? [] : [booleanValue]}
+            onChange={(next) => getUpdate(booleanPath, next[0] ?? null)}
+            options={[
```
AronH99, line 165: 🤖 Claude · [Question]: the old optional control had an explicit `null` radio; now clearing back to unset means clicking the already-selected button again, which is less discoverable. For an attribute that defaults to null this is likely fine, but it's a behaviour/UX change worth a conscious product call before merge (the attached recording helps). If we wanted it explicit we could add a small clear affordance — I'd leave that to the human reviewer. :)

AronH99 (commented):
## 🤖 Claude pre-review — PR #7034 — Replace optional boolean form field radios with a True/False toggle group

**Context map**

- **Type & intent:** refactor/fix (changelog: patch · minor-improvement) — optional boolean attributes in the service-instance form now render as a `[True][False]` `OptionalToggleGroup` (no selection = null) instead of True/False/null radios; required booleans keep their existing Switch. Closes #6996.
- **Surface area (grouped):** *shared UI* — `src/UI/Components/OptionalToggleGroup` (generic widened to `string|boolean`, `isDisabled` prop, deselect-blur) and `ServiceInstanceForm/Components/FieldInput` (Boolean branch rewritten; `BooleanFormInput` deleted). Tests in `Slices/CreateInstance`, `Slices/EditInstance`, and both shared components. + changelog.
- **Size & balance:** +224 / −182 across 9 files; solid test delta (102 new `OptionalToggleGroup` test lines incl. falsy-boolean + focus coverage).
- **Blast radius:** *medium* — `OptionalToggleGroup` is shared (also drives the resource status filters). Widening keeps the default `T = string`, so the filter consumers stay source-compatible (their tests pass); `isDisabled` and `key={value}→key={buttonId}` are additive. Note the deselect-blur now applies to the filters too.
- **Risk hotspots:** shared-component change (filters ride along), a11y of a toggle-group-as-form-field, keyboard focus on blur, test-helper duplication.

No blockers — behaviour is correct, scope is tight, types are clean, and the new tests guard the right things (falsy `false` + focus). Posting as `COMMENT` only because GitHub blocks self-approve; treat as approve-with-nits. Nothing flips this to request-changes; the two should-fixes are small.

CI: CodeQL green; Jenkins was still PENDING at review time — confirm green before merge. 0 blocking, 2 should-fix, 2 questions below. One out-of-scope note: the `FormGroup` carries `aria-label="BooleanFieldInput-<name>"`, exposing a dev/test string as the group's accessible name — pre-existing (the old `BooleanFormInput` did the same), not introduced here, but a worthwhile follow-up since assistive tech reads it aloud.

AronH99:
So just 1 question @bartv:
Do we want to keep it like it is now with the true/false or do we need do explicitly have a 3th toggle with "null" to make the intend clear?
Right now you can just toggle both to inactive and that sets the "null" value but that might be some adjustment to what it previously was (see screenshot):

Screenshot 2026-06-23 at 15 40 42

bartv:
> So just 1 question @bartv: Do we want to keep it like it is now with the true/false or do we need do explicitly have a 3th toggle with "null" to make the intend clear? Right now you can just toggle both off to set the "null" value but that might be some adjustment to what it previously was (see screenshot):
> 
> Screenshot 2026-06-23 at 15 40 42

@edvgui What do think? I think this is just fine.

edvgui:
> @edvgui What do think? I think this is just fine.

What is shown in the video looks good to me

src/UI/Components/OptionalToggleGroup/OptionalToggleGroup.tsx
```diff
@@ -23,28 +24,38 @@ export interface OptionalToggleGroupProps<T extends string = string> {
...
+    }
```
bartv, line 58: _Claude here, posting on behalf of Bart._

A question on this `blur()` workaround rather than a blocker. It runs through the button's click handler, and a native `<button>` also fires `click` on keyboard activation (Enter/Space), so this path executes for keyboard deselect too. There, `blur()` moves focus to `<body>`: the keyboard user loses their place (the next Tab restarts from the top of the document), and the focus ring it strips is exactly what they rely on to see where they are.

I checked why the artifact happens at all. PatternFly's `toggle-group.css` styles the lingering highlight on plain `:focus`, not `:focus-visible`:

```css
.pf-v6-c-toggle-group__button:is(:hover, :focus) {
  --pf-v6-c-toggle-group__button--BackgroundColor: var(--pf-v6-c-toggle-group__button--hover--BackgroundColor);
}
```

Because it is `:focus`, a mouse click keeps the highlight (browsers only suppress `:focus-visible` for mouse), so the workaround is genuinely needed for the mouse case. The same rule is just why keyboard gets caught too.

The root-cause fix would be re-scoping that to `:focus-visible`, but that is a CSS override of a PF component, which we want to avoid since it makes PF upgrades expensive. So I would suggest keeping the blur but limiting it to pointer clicks via `event.detail` (a real mouse click has `detail >= 1`, a keyboard-triggered click has `detail === 0`). Smallest version, only the call site changes:

```js
onChange={(event) => handleChange(value, event.detail === 0 ? null : event.currentTarget)}
```

For keyboard activations `target` then comes through as `null`, so the existing `target instanceof HTMLElement` guard skips the blur; mouse clicks blur as they do today. The `not.toHaveFocus()` test keeps passing, and a keyboard test could assert focus is retained.

@AronH99 does the keyboard case matter for this control in your view, and are you OK with the `event.detail` guard rather than a CSS override? Happy to push the tweak if you would like.

AronH99: On chrome and modern browsers I believe this issue is not as big as this comment makes it seem, so I'd probably say this is currently not worth the adjustment.
I did however as part of a bigger ticket made a ticket related to this where the accessibility (keyboard navigation) of the forms should be improved where a user should be able to go through the entire form only using keyboard.
Currently that's not the case and I thought maybe this might be useful to have in the future for filling out our forms as fast as possible. 
https://github.com/inmanta/web-console/issues/7042

This is the ticket.
I tried implementing this on my other pr of the form improvements but it was getting a little out of hand and I think I need to first think it through/analyze it properly.
But only if there is a clear aks/use for it off course.
What do you think? @bartv

bartv (approved):

PR #7038, opened 2026-06-24
Consolidate code editors into a single shared CodeEditor component

### LINKED ISSUE #6998: Make a PF generic CodeEditor component with a reusable/common preset we can rely on throughout the entire console
We currently have quite a few CodeEditors which most likely can be reworked to reuse a single component in the shared ui folder.
Maybe 
We have the ReadOnlyCodeEditor but we probably can rework that one a little into CodeEditor which we can then reuse throughout the console.
=> essentially would be nice to have an reuse 1 code editor component

AronH99:
# Description

- Added a shared CodeEditor (PatternFly/Monaco wrapper) under src/UI/Components/CodeEditor/ with a common preset: auto dark/light theme, http-safe copy control, download, and content-sized height with an expand toggle

- Replaced ReadOnlyCodeEditor and migrated all 6 call sites (AttributeList, FileBlock, Traceback, OrderDetailsRow, DictFieldInput, MarkdownPreviewer) to the shared component

- Fixed the missing dark-theme on the FileBlock and Orders "body" editors, and the language label now follows the editor's language

- Moved the ClassifiedAttribute-specific helpers (isEditorKind / languageForKind) into AttributeList/ to keep the shared editor domain-free

- Removed MarkdownCodeEditorControls and the now-unused word keys; the markdown previewer uses the built-in copy + download (downloadFileName), dropping the "Copy raw" shortcut

So in all essence what this PR represents is the unification of the CodeEditor. 
It's now consistently the same component throughout the entire web-console which means some minor changes here and there but all intended functionalities everywhere should have stayed the same.
Some very minor visual changes here and there but nothing major to make note of.
With this PR came some solid cleanup as well.
I also tested every and all code editors myself and compared them to the current implementation we have on mgmt, everything should be as intended.

closes https://github.com/inmanta/web-console/issues/6998


src/UI/Components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,60 @@
...
+      // Remount on toggle so Monaco re-layouts to the new height.
+      key={isAutoHeight ? String(isExpanded) : undefined}
+      code={code}
+      language={language}
+      height={resolvedHeight}
+      isReadOnly={isReadOnly}
+      isDarkTheme={isDark}
+      isDownloadEnabled={isDownloadEnabled}
+      // Default to showing the label when a language is set; ?? keeps an explicit false.
+      isLanguageLabelVisible={isLanguageLabelVisible ?? language !== undefined}
```
AronH99, line 64: 🤖 Claude · [Question]: defaulting `isLanguageLabelVisible` to `language !== undefined` is a nice touch, but it's a visible change for the two read-only editors that previously passed no label — Traceback now shows a `PYTHON` tab and OrderDetails a `JSON` tab where PF's default (`false`) gave none before. That matches your "label follows the language" note in the description, so likely intended — raising it as a visual/product call for the human reviewer to confirm across all read-only sites, not a code issue. The `??` correctly preserves an explicit `false` (DictFieldInput). :)

src/Slices/MarkdownPreviewer/UI/MarkdownPreviewer.tsx
```diff
@@ -51,22 +49,13 @@ export const MarkdownPreviewer: React.FC<Props> = ({ service, instance, instance
...
           <CodeEditor
-            isDarkTheme={isDark}
-            isLineNumbersVisible
             isMinimapVisible
-            isLanguageLabelVisible
             code={markdownContent}
             language={Language.markdown}
-            isFullHeight
+            isReadOnly={false}
             height="calc(100vh - 550px)"
```
AronH99, line 56: 🤖 Claude · [Question]: this drops the old `isFullHeight` while keeping the explicit `height="calc(100vh - 550px)"`. PF only applies its `pf-m-full-height` modifier when `isFullHeight` is true or `height==="100%"`, so that modifier is no longer set here. With an explicit calc height driving the Monaco container it's probably visually identical (and you said you tested it) — just confirming the preview pane still fills as before; if it looks short, `isFullHeight` still passes through via `...rest`. Not blocking.

src/UI/Components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * and content-sized height with an expand toggle. Read-only by default; pass
+ * `isReadOnly={false}` + `onChange` to edit. All other PF props pass through.
+ */
+export const CodeEditor: React.FC<Props> = ({
+  code = "",
+  language,
+  height,
+  isReadOnly = true,
+  isDownloadEnabled = true,
+  isCopyEnabled = true,
```
AronH99, line 28: 🤖 Claude · [Nit]: worth a one-liner in the JSDoc that `isCopyEnabled` is *intercepted* here rather than forwarded — PF's own `isCopyEnabled` renders its built-in copy (which uses `navigator.clipboard` and breaks on http), whereas here it toggles the http-safe `CodeEditorCopyControl` instead. The behaviour is right; it's just that someone reading `Omit<CodeEditorProps, ...>` would reasonably expect PF semantics. Same goes for noting `isDownloadEnabled` defaults to `true` here vs PF's `false`.

src/UI/Components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  // Without an explicit height, size the editor to its content and offer an
+  // expand/collapse toggle. An explicit height fixes the size and drops the toggle.
+  const isAutoHeight = height === undefined;
+  const resolvedHeight = isAutoHeight ? getAutoHeight(code, isExpanded) : height;
+
+  return (
+    <PFCodeEditor
+      // Remount on toggle so Monaco re-layouts to the new height.
+      key={isAutoHeight ? String(isExpanded) : undefined}
```
AronH99, line 56: 🤖 Claude · [Nit]: the remount-on-toggle `key` is correct for the read-only auto-height editors (every current auto-height caller is read-only). One latent footgun now that this is shared: an editable caller with no `height` + `isReadOnly={false}` + `onChange` would lose Monaco's undo stack and cursor on each expand/collapse. Not triggered today — a line in the docstring ("auto-height assumes read-only") or only keying when `isReadOnly` would close it off for the next person.

src/UI/Components/index.ts
```diff
@@ -35,7 +35,7 @@ export * from "./MultiTextSelect";
 export * from "./PageContainer";
 export * from "./PageSectionWithTitle";
 export * from "./PaginationWidget";
-export * from "./ReadOnlyCodeEditor";
+export * from "./CodeEditor";
 export * from "./RequiresTable";
```
AronH99, line 39: 🤖 Claude · [Nit]: `CodeEditor` landed in `ReadOnlyCodeEditor`'s old slot (between `PaginationWidget` and `RequiresTable`), but this barrel is otherwise alphabetical — it should sit up with the `C`s. Trivial, just keeps the list scannable.

AronH99 (commented):
## 🤖 Claude pre-review — PR #7038 — Consolidate code editors into a single shared CodeEditor component

**Context map**

- **Type & intent:** refactor (changelog: patch) — replace `ReadOnlyCodeEditor` and the per-site raw PatternFly editors with one shared `CodeEditor` wrapper under `src/UI/Components/CodeEditor/`, carrying a common preset: app dark/light theme, http-safe copy, download, and content-sized height with an expand toggle. Closes #6998.
- **Surface area:** all in shared `src/UI`. New `Components/CodeEditor/`; touched `AttributeList/` (domain helpers `isEditorKind`/`languageForKind` moved in, plus `AttributeList`/`FileBlock`), `CodeEditorControls/` (dropped unused `code` prop), `DictFieldInput`, the `Components` barrel and `words.tsx`; deleted `ReadOnlyCodeEditor/` and `MarkdownCodeEditorControls`. Slice call sites migrated: Diagnose/Traceback, MarkdownPreviewer, OrderDetails.
- **Size & balance:** +175 / −322 across 23 files (net −147). Test delta present and matched — new `CodeEditor.test` + `helpers.test`, the moved `getDefaultHeightEditor` test follows its code, and the obsolete `MarkdownCodeEditorControls.test` is removed. No duplicate tests left behind.
- **Blast radius: HIGH (well mitigated).** A shared component reached from 6 call sites across 3 slices plus Facts (via `isEditorKind`). I verified the migration is complete — no lingering `ReadOnlyCodeEditor` / `MarkdownCodeEditorControls` / `copy.raw` references, and the only direct `@patternfly/react-code-editor` `CodeEditor` import now lives in the wrapper (single chokepoint). Per-site prop parity checked against PF defaults (read-only, download, copy, line-numbers all preserved).
- **Risk hotspots:** prop-default parity across the 6 migrated sites; two deliberate visual changes (language label, full-height); shared-component API ergonomics.

**Verdict: Approve (with nits)**

Clean consolidation: net −147 LOC, a single PF chokepoint, two real bug fixes folded in (dark theme now reaches the FileBlock and OrderDetails "body" editors), a verified-complete migration, and honest tests that derive assertions from `words(...)` rather than literals. No blockers. I confirmed the markdown download is preserved exactly — PF appends `.${getExtensionFromLanguage(markdown)}` = `.md`, so `downloadFileName={service-instance-documentation}` still yields `service-instance-documentation.md`; nice cleanup dropping the bespoke `MarkdownCodeEditorControls` for the built-in download. Posting as a `COMMENT` review only because GitHub blocks self-approve on your own PR — treat it as approve-with-nits. 0 blocking, 2 questions, 3 nits below.

src/UI/Components/CodeEditor/CodeEditor.tsx
AronH99: not relevant at this point

src/UI/Components/CodeEditor/CodeEditor.tsx
AronH99: not relevant

src/Slices/MarkdownPreviewer/UI/MarkdownPreviewer.tsx
AronH99: not a real problem/issue

AronH99:
This can be reviewed now

bartv:
Looks like real nice cleanup. Claude flagged another issue that I want you to have a look at before we merge.

src/UI/Components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <>
+          {isCopyEnabled && <CodeEditorCopyControl code={code} />}
+          {isAutoHeight && (
+            <CodeEditorHeightToggleControl
+              isExpanded={isExpanded}
+              onToggle={() => setIsExpanded(!isExpanded)}
+            />
+          )}
+        </>
+      }
```
bartv: _Claude here, posting this review on behalf of Bart._

`customControls` is always a truthy React element here, even when both children render nothing. When a caller disables copy **and** passes an explicit `height` (so `isAutoHeight` is false), the fragment's children are both `false`, but the `<>…</>` element itself is still truthy.

PatternFly treats any truthy `customControls` as header content (`hasEditorHeaderContent = … || customControls || …` in `@patternfly/react-code-editor`'s `CodeEditor.js`), so it renders the `codeEditorHeaderContent` / `codeEditorControls` band regardless.

This regresses `DictFieldInput`, which sets `isCopyEnabled={false} isDownloadEnabled={false} isLanguageLabelVisible={false}` with an explicit height: before this PR it passed no `customControls` and rendered no header content; now an empty toolbar strip (with PF padding) appears above the JSON dict editor.

Suggested fix: only pass the fragment when it actually has controls, e.g.

```tsx
const controls = (
  <>
    {isCopyEnabled && <CodeEditorCopyControl code={code} />}
    {isAutoHeight && (
      <CodeEditorHeightToggleControl
        isExpanded={isExpanded}
        onToggle={() => setIsExpanded(!isExpanded)}
      />
    )}
  </>
);
// ...
customControls={isCopyEnabled || isAutoHeight ? controls : undefined}
```

bartv (commented):
_Claude here, posting this review on behalf of Bart._

Nice consolidation - one shared `CodeEditor` wrapper replacing `ReadOnlyCodeEditor` + the bespoke `MarkdownCodeEditorControls`, all 6 call sites migrated, net -141 lines. The cross-file move of `isEditorKind`/`languageForKind` into `AttributeList/helpers` is safe: the `@/UI/Components` barrel re-exports `./AttributeList` and `AttributeList/index.ts` now re-exports `./helpers`, so the `Facts` slice consumers keep resolving.

I checked theme handling, the `isLanguageLabelVisible ?? language !== undefined` defaulting, the auto-height remount `key`, and the new default-`true` `isDownloadEnabled` against prior behavior at each call site - all consistent. The dropped "Copy raw" control and `markdownPreviewer.download*` word keys are intentional and have no remaining references on the branch.

One minor finding (left inline): the always-rendered `customControls` fragment forces an empty editor header in `DictFieldInput`. Non-blocking - everything else looks good.

bartv (approved):

PR #7039, opened 2026-06-24
Sort numeric enum selects by value and make long option lists scrollable

### LINKED ISSUE #7037: Enum fields defined as numbers are not sorted properly
The enum field of sizing_memory => MenuList is not sorted properly, most likely because they are handled as strings and not as numbers

Image

Reference url:
http://172.25.223.3:8888/console/lsm/catalog/paloalto_firewall_cluster/inventory/1422c6bd-768b-42f6-9f12-fdcd2be20f83/edit?env=0463881e-b4d2-11f0-a02c-2800af76fefe

AronH99:
# Description

- Numeric enum options in service instance form selects now sort by value (5, 6, … 128) instead of lexicographically (10, 128, … 5)
- String enum options continue to sort alphabetically
- Long option lists are now scrollable (`isScrollable` on `Select`) so 20+ item enums no longer dominate the screen

closes https://github.com/inmanta/web-console/issues/7037

Screenshot 2026-06-24 at 16 16 49

Note: Not going to review this one with the skill from Lukas since this is a very small change which has a very limited impact, not much to review here...

### DIFF
--- a/src/UI/Components/ServiceInstanceForm/Components/SelectFormInput.tsx
+++ b/src/UI/Components/ServiceInstanceForm/Components/SelectFormInput.tsx
@@ -50,7 +50,9 @@ export const SelectFormInput: React.FC<Props> = ({
 }) => {
   const [isOpen, setIsOpen] = useState(false);
 
-  const selectOptions = Object.keys(options).sort();
+  const selectOptions = Object.keys(options).sort((a, b) =>
+    a.localeCompare(b, undefined, { numeric: true })
+  );
   const formattedOptions = selectOptions.map((option) => {
     return {
       value: options[option],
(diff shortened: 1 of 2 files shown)


src/UI/Components/ServiceInstanceForm/Components/SelectFormInput.tsx
```diff
@@ -50,7 +50,9 @@ export const SelectFormInput: React.FC<Props> = ({
 }) => {
   const [isOpen, setIsOpen] = useState(false);
 
-  const selectOptions = Object.keys(options).sort();
+  const selectOptions = Object.keys(options).sort((a, b) =>
```
bartv, line 53: What happens on string forms in this case? Do we need to scope it to numeric fields only?

AronH99: String enum options continue to sort alphabetically => same behaviour as before

bartv (approved):

PR #7049, opened 2026-06-26
Support label/value pairs in service instance form suggestions

### LINKED ISSUE #7008: Split label and value in suggested values
# T1 — Split label and value in suggested values

> Part of #7007 Foundation for T2/T3/T4, which all assume this output shape.

## Background
Service form text fields can show an autocomplete dropdown of suggested values
(`web_suggested_values` annotation). Today each suggestion is a single string
that does three jobs at once: it's shown to the user, it's what the filter
matches on, and it's what gets submitted to the API.

That falls apart when the value the API needs isn't human-friendly — a UUID, an
internal id, a `source://` URI. The dropdown ends up showing
`9f3c1b2a-…` and the operator has to recognize raw identifiers.

## Goal
Let a suggestion carry a **label** (shown + searched) separate from the
**value** (submitted):

- Show `10 Gbps` → submit `10000`
- Show `Production network` → submit `9f3c1b2a-…`

A plain string stays valid and means `label == value`, so **existing models keep
working unchanged**. This applies to all flavors (`literal`, `parameters`, and
the future `graphql`), giving one shared output shape: `{ label, value }[]`.

## Scope
- [ ] Widen the `values` type to accept `string | { label: string; value: string }`
      — `src/Core/Domain/ServiceInstanceModel.ts` (`FormSuggestion`, ~L88).
- [ ] Normalize every flavor into a single `{ label, value }[]` shape. Replace the
      flat `string[]` handling and remove the fragile format guard / TODO
      — `src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx` (~L97–115)
      and `src/Data/Queries/Slices/ServiceInstance/FormSuggestions/useSuggestions.ts`.
      A bare string normalizes to `{ label: s, value: s }`.
- [ ] Display + filter on `label`, submit `value`:
      - `SuggestionsPopover.tsx` — render `label`, filter on `label`, fix the
        `key`/`aria-label` (currently the suggestion string).
      - `TextFormInput.tsx` (~L143–156) — `handleSuggestionClick` must submit
        `value`, not the displayed label.
      - `TextListFormInput.tsx` — same for the chip/list case.

## Acceptance criteria
- [ ] A suggestion can show a label while submitting a different value.
- [ ] Plain-string suggestions still render and submit identically to today.
- [ ] The dropdown filters on the label text.
- [ ] No `string[]` suggestion contract remains; flavors share `{ label, value }[]`.
- [ ] Tests cover both entry shapes (bare string and `{ label, value }`) for the
      single-value and list inputs.

## Notes
- Backward compatible: pure superset of the current annotation format.
- This is also the right moment to clear the existing
  *"remove this when the API returns a fixed format"* TODO in `FieldInput.tsx`.
- No backend/API change — annotations pass through unchanged.

AronH99:
# Description

- Add a `SuggestionValue` `{ label, value }` type; `FormSuggestion.values` now accepts plain strings or `{ label, value }` pairs
- Normalize every suggestion flavor (literal, parameters) to one `{ label, value }[]` shape in `useSuggestions`, coercing numeric fields to strings

So essentially it's now possible to use label/value pairs for the form suggestions and everything will get normalized into that shape.
Wrote tests to try and minimize/catch regressions.
I think more improvements are possible but I split possible future form improvements (mostly keyboard related) into a different ticket to try and stick to the scope/essence of the ticket.

Below the example (10Gbps as label  => 10000 as value):
Screenshot 2026-06-26 at 15 38 29
Screenshot 2026-06-26 at 15 38 34

closes https://github.com/inmanta/web-console/issues/7008

### DIFF
--- a/changelogs/unreleased/7008-suggestion-label-value.yml
+++ b/changelogs/unreleased/7008-suggestion-label-value.yml
@@ -1,4 +1,4 @@
-description: Service instance form suggestions now support `{ label, value }` pairs, showing a human-friendly label while submitting the underlying value.
+description: Service instance form suggestions now support "label/value pairs", showing a human-friendly label while submitting the underlying value.
 issue-nr: 7008
 change-type: minor
 destination-branches: [master, iso9]
--- a/changelogs/unreleased/7008-suggestion-label-value.yml
+++ b/changelogs/unreleased/7008-suggestion-label-value.yml
@@ -1,0 +1,6 @@
+description: Service instance form suggestions now support `{ label, value }` pairs, showing a human-friendly label while submitting the underlying value.
+issue-nr: 7008
+change-type: minor
+destination-branches: [master, iso9]
+sections:
+  minor-improvement: "{{description}}"
--- a/src/Core/Domain/ServiceInstanceModel.ts
+++ b/src/Core/Domain/ServiceInstanceModel.ts
@@ -83,11 +83,27 @@ export interface FormAttributeResult {
 }
 
 /**
+ * Interface representing a single suggestion.
+ *
+ * The `label` is shown to the user and searched on, while the `value` is what
+ * gets submitted to the API. A plain string suggestion normalizes to a pair
+ * where `label === value`.
+ */
+export interface SuggestionValue {
+  label: string;
+  value: string;
+}
+
+/**
  * Interface representing a suggestions that are stored in the web_suggested_values.
+ *
+ * A suggestion entry can be a plain string (label and value are identical) or a
+ * `{ label, value }` pair where the displayed/searched label differs from the
+ * submitted value.
  */
 export interface FormSuggestion {
   type: FormSuggestionType;
-  values?: string[];
+  values?: (string | SuggestionValue)[];
   parameter_name?: string;
 }
 
--- a/src/Core/Domain/ServiceInstanceModel.ts
+++ b/src/Core/Domain/ServiceInstanceModel.ts
@@ -87,7 +87,7 @@ export interface FormAttributeResult {
  *
  * This is the shape the form actually consumes: the `label` is shown to the user
  * and searched on, while the `value` is what gets submitted to the API. Both are
- * always strings — `normalizeSuggestions` coerces every raw entry (see
+ * always strings - `normalizeSuggestions` coerces every raw entry (see
  * {@link RawFormSuggestion}) into this shape. A plain string suggestion
  * normalizes to a pair where `label === value`.
  */
--- a/src/Core/Domain/ServiceInstanceModel.ts
+++ b/src/Core/Domain/ServiceInstanceModel.ts
@@ -83,11 +83,13 @@ export interface FormAttributeResult {
 }
 
 /**
- * Interface representing a single suggestion.
+ * Interface representing a single, normalized suggestion.
  *
- * The `label` is shown to the user and searched on, while the `value` is what
- * gets submitted to the API. A plain string suggestion normalizes to a pair
- * where `label === value`.
+ * This is the shape the form actually consumes: the `label` is shown to the user
+ * and searched on, while the `value` is what gets submitted to the API. Both are
+ * always strings — `normalizeSuggestions` coerces every raw entry (see
+ * {@link RawFormSuggestion}) into this shape. A plain string suggestion
+ * normalizes to a pair where `label === value`.
  */
 export interface SuggestionValue {
   label: string;
@@ -95,15 +97,29 @@ export interface SuggestionValue {
 }
 
 /**
- * Interface representing a suggestions that are stored in the web_suggested_values.
+ * A suggestion entry exactly as it arrives in `web_suggested_values` (or a
+ * parameter's metadata), before normalization.
  *
- * A suggestion entry can be a plain string (label and value are identical) or a
- * `{ label, value }` pair where the displayed/searched label differs from the
- * submitted value.
+ * It can be a bare scalar or an explicit `{ label, value }` pair, and either
+ * form may be a string or a number (numeric attributes). Every variant is
+ * coerced to a string-only {@link SuggestionValue} by `normalizeSuggestions`
+ * before the form uses it.
+ */
+export type RawFormSuggestion =
+  | string
+  | number
+  | { label: string | number; value: string | number };
+
+/**
+ * Interface representing the suggestions that are stored in the web_suggested_values.
+ *
+ * `values` holds the raw entries (see {@link RawFormSuggestion}): each is a plain
+ * scalar (label and value are identical) or a `{ label, value }` pair where the
+ * displayed/searched label differs from the submitted value.
  */
 export interface FormSuggestion {
   type: FormSuggestionType;
-  values?: (string | SuggestionValue)[];
+  values?: RawFormSuggestion[];
   parameter_name?: string;
 }
 
--- a/src/Data/Queries/Slices/ServiceInstance/FormSuggestions/useSuggestions.ts
+++ b/src/Data/Queries/Slices/ServiceInstance/FormSuggestions/useSuggestions.ts
@@ -1,67 +1,13 @@
 import { useContext, useMemo } from "react";
 import { useQuery } from "@tanstack/react-query";
-import { FormSuggestion, SuggestionValue } from "@/Core";
+import { FormSuggestion } from "@/Core";
 import { useGet, getParametersKey } from "@/Data/Queries";
 import { DependencyContext } from "@/UI/Dependency";
+import { normalizeSuggestions } from "./helpers";
 
 interface ResponseData {
   parameter?: { metadata?: { values?: unknown } };
 }
-
-/**
- * Type guard for a string or number scalar.
- *
- * @param value - The value to check.
- * @returns Whether the value is a string or a number.
- */
-const isStringOrNumber = (value: unknown): value is string | number =>
-  typeof value === "string" || typeof value === "number";
-
-/**
- * Type guard for an explicit `{ label, value }` suggestion pair.
- *
- * Both fields may be a string or a number (e.g. numeric attributes); they are
- * coerced to strings when normalized.
- *
- * @param entry - The raw entry to check.
- * @returns Whether the entry is a `{ label, value }` pair of scalars.
- */
-const isLabelValuePair = (
-  entry: unknown
-): entry is { label: string | number; value: string | number } =>
-  typeof entry === "object" &&
-  entry !== null &&
-  "label" in entry &&
-  "value" in entry &&
-  isStringOrNumber(entry.label) &&
-  isStringOrNumber(entry.value);
-
-/**
- * Normalizes a raw list of suggestion entries into a single `{ label, value }[]` shape.
- *
- * Every flavor (literal, parameters, ...) is reduced to this shape so the rest
- * of the form only deals with one contract. A bare scalar normalizes to a pair
- * where `label === value`; an explicit pair has its (possibly numeric) fields
- * coerced to strings; anything else is dropped.
- *
- * @param values - The raw values, of unknown shape.
- * @returns The normalized suggestions, or null when `values` is not an array.
- */
-export const normalizeSuggestions = (values: unknown): SuggestionValue[] | null => {
-  if (!Array.isArray(values)) {
-    return null;
-  }
-
-  return values.reduce<SuggestionValue[]>((acc, entry) => {
-    if (isStringOrNumber(entry)) {
-      acc.push({ label: String(entry), value: String(entry) });
-    } else if (isLabelValuePair(entry)) {
-      acc.push({ label: String(entry.label), value: String(entry.value) });
-    }
-
-    return acc;
-  }, []);
-};
 
 /**
  * React Query hook to handle suggested values for a parameter.
--- a/src/Data/Queries/Slices/ServiceInstance/FormSuggestions/useSuggestions.ts
+++ b/src/Data/Queries/Slices/ServiceInstance/FormSuggestions/useSuggestions.ts
@@ -1,17 +1,73 @@
-import { useContext } from "react";
+import { useContext, useMemo } from "react";
 import { useQuery } from "@tanstack/react-query";
-import { FormSuggestion } from "@/Core";
+import { FormSuggestion, SuggestionValue } from "@/Core";
 import { useGet, getParametersKey } from "@/Data/Queries";
 import { DependencyContext } from "@/UI/Dependency";
 
 interface ResponseData {
-  parameter: Record<string, Record<string, unknown>>;
+  parameter?: { metadata?: { values?: unknown } };
 }
 
 /**
+ * Type guard for a string or number scalar.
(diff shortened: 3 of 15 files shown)


src/Data/Queries/Slices/ServiceInstance/FormSuggestions/useSuggestions.ts
```diff
@@ -1,17 +1,73 @@
...
+ *
+ * Every flavor (literal, parameters, ...) is reduced to this shape so the rest
+ * of the form only deals with one contract. A bare scalar normalizes to a pair
+ * where `label === value`; an explicit pair has its (possibly numeric) fields
+ * coerced to strings; anything else is dropped.
+ *
+ * @param values - The raw values, of unknown shape.
+ * @returns The normalized suggestions, or null when `values` is not an array.
+ */
+export const normalizeSuggestions = (values: unknown): SuggestionValue[] | null => {
```
AronH99: 🤖 Claude · [Should-fix]: `normalizeSuggestions` + the two type guards are pure, separately-tested logic living in the hook file. The repo's file-role split keeps logic in a `helpers.ts` (the sibling `PostInstance` slice already has one with its own `helper.test.ts`) and hook files to hooks. Consider moving these three into `FormSuggestions/helpers.ts` and the spec to `helpers.test.ts`; `useSuggestions.ts` then just imports and calls. Judgment call, not a hard blocker — but it keeps the slice consistent with the rest of `Data/Queries`.

src/UI/Components/ServiceInstanceForm/Components/TextFormInput.tsx
```diff
@@ -57,27 +48,53 @@ export const TextFormInput: React.FC<Props> = ({
...
-   *
-   * @param {string} value - The new value of the input field.
-   *
-   * @returns {void}
-   */
-  const handleChange = (value) => {
-    setInputValue(value);
+  // Leaving the field is its "commit" point: resolve the value to its label,
+  // Plain (non-suggestion) fields are left alone so their typed text isn't normalized.
+  const handleBlur = () => {
```
AronH99, line 82: 🤖 Claude · [Question]: `handleBlur` early-returns for plain fields *before* clearing `editedRef`, so once a non-suggestion field is edited `editedRef.current` stays `true` for the component's lifetime — and the sync effect below (gated on `!editedRef.current`) then ignores any later external `attributeValue` change for that field. In normal typing `formState` just mirrors the input so it's invisible, but it's a behavior change from the old sync effect. Can you confirm nothing re-syncs a plain field after edit (form reset / cross-field population / duplicate-instance)? If unsure, clearing `editedRef` here too keeps plain fields fully controlled.

src/UI/Components/ServiceInstanceForm/Components/TextFormInput.tsx
```diff
@@ -123,35 +140,28 @@ export const TextFormInput: React.FC<Props> = ({
...
+          {hasSuggestions && (
             <SuggestionsPopover
               suggestions={suggestions}
-              filter={inputRef.current?.value || ""}
-              handleSuggestionClick={(suggestion) => {
-                if (inputRef.current) {
-                  handleChange(suggestion);
-                }
-              }}
+              filter={displayValue}
```
AronH99, line 160: 🤖 Claude · [Question]: `filter={displayValue}` means that when you focus a field that already holds a value (edit flow), the popover is pre-filtered to the resolved label — so only the current selection shows and you have to clear the field to browse the other options. Intended typeahead behavior, or would showing the full list on focus be friendlier? Minor UX, leaving the call to you.

src/UI/Components/ServiceInstanceForm/Components/TextFormInput.tsx
```diff
@@ -123,35 +140,28 @@ export const TextFormInput: React.FC<Props> = ({
           <TextInput
             ref={inputRef}
             isRequired={!isOptional}
-            type={type}
+            // Rendered as text so a non-numeric label coupled to a numeric
+            // value can show (e.g. "10 Gbps" + 10000).
+            type={hasSuggestions ? TextInputTypes.text : type}
```
AronH99, line 145: 🤖 Claude · [Nit / product call]: nicely commented — just surfacing it explicitly: forcing `type=text` whenever suggestions exist drops the native number affordances (spinner, numeric mobile keyboard) for numeric attributes that have suggestions. It's the right trade-off to show a non-numeric label; flagging it as a deliberate UX decision for the human reviewer rather than ruling on it.

src/UI/Components/ServiceInstanceForm/Components/SuggestionsPopover.tsx
```diff
@@ -17,46 +18,49 @@ interface Props {
...
     /**
      * Handles the suggestion click event.
      *
      * @param {React.MouseEvent} event - The click event.
-     * @param {string} suggestion - The suggestion value.
+     * @param {SuggestionValue} suggestion - The clicked suggestion.
      * @returns {void}
      */
-    const handleSuggestionClickInternal = (event, suggestion) => {
+    const handleSuggestionClickInternal = (event, suggestion: SuggestionValue) => {
```
AronH99: 🤖 Claude · [Nit]: you typed `suggestion` here but left `event` implicit — and the JSDoc right above already declares it a `React.MouseEvent`. A quick `event: React.MouseEvent` keeps it aligned with the repo's typing bar :)

src/UI/Components/ServiceInstanceForm/Components/suggestionResolvers.ts
```diff
@@ -0,0 +1,21 @@
...
+  value: unknown
+): string =>
+  suggestions?.find((suggestion) => suggestion.value === String(value ?? ""))?.label ??
+  String(value ?? "");
+
+/**
+ * Resolves the value to submit for typed text: if it exactly matches a
+ * suggestion's label, returns that suggestion's value; otherwise the text itself.
+ */
+export const resolveValue = (
```
AronH99, line 22: 🤖 Claude · [Question]: `resolveValue` substitutes the underlying value whenever typed text *exactly* equals a suggestion label. Correct for a typeahead, but on a free-text-with-suggestions field it means a user who types a literal string that happens to match a label silently submits a different value. Intended? If so, a one-liner in the docstring noting that exact-label typing is treated as a selection would save a future reader the double-take.

AronH99 (commented):
## 🤖 Claude pre-review — PR #7049 — Support label/value pairs in service instance form suggestions

**Context map**

- **Type & intent:** improvement (changelog: minor) — let form suggestions carry `{ label, value }` pairs: show a human-friendly label while submitting the underlying value (e.g. "10 Gbps" → 10000). Closes #7008.
- **Surface area (grouped):**
  - *Core domain* — `ServiceInstanceModel.ts`: new `SuggestionValue` + `RawFormSuggestion`, `FormSuggestion.values` widened from `string[]`.
  - *Data layer* — `FormSuggestions/useSuggestions.ts`: new `normalizeSuggestions` + two type guards; both literal and parameter flavors collapse to one `{label,value}[]` contract. `PostInstance/helper.test.ts`: numeric-sanitize coverage.
  - *Form UI* — `ServiceInstanceForm/Components`: `TextFormInput` (typeahead — display label / submit value / round-trip), `TextListFormInput` (chips show label, store value), `SuggestionsPopover` (now `SuggestionValue[]`, effect→derived filter, `setIsOpen`→`close`), `FieldInput` (drops local state+effect, forwards normalized data), new shared `suggestionResolvers.ts`.
- **Size & balance:** +609 / −193 across 13 files. Strong test delta — 2 new specs + 3 updated; covers normalize edge cases, numeric coercion, and load/edit/async round-trips.
- **Blast radius:** *low–medium*. `SuggestionValue[]` and the `close` prop are internal to the ServiceInstanceForm components. I checked the one external consumer — `TreeTable/TreeRow/InlineInput.tsx` renders `TextListFormInput` without `suggestions`, so the type widening doesn't touch it. CI green (CodeQL + Jenkins SUCCESS).
- **Risk hotspots:** the typeahead state machine in `TextFormInput` (edit/blur/async round-trips); label↔value reverse mapping; placement of the new pure helpers.

**Verdict: Comment (approve-with-nits in spirit)**

No blockers — the feature is correct and well-covered, the new exports carry real intent-explaining docstrings (matching the repo bar), the resolvers are DRY-shared across both inputs, and the `SuggestionsPopover` derived-state cleanup is a genuine improvement. Posted as `COMMENT` only because GitHub blocks self-approve on your own PR; treat it as approve-with-nits. The one thing I'd actually act on before merge is the placement of the pure normalize/guard logic (should-fix below); the rest are questions/nits you can take or leave.

0 blocking · 1 should-fix · 2 nits · 3 questions inline.

AronH99:
@bartv 
Weet niet of je dit kan instellen maar deze pr mag zeker en vast zeer kritisch reviewed worden

src/Data/Queries/Slices/ServiceInstance/FormSuggestions/useSuggestions.ts
```diff
@@ -34,8 +36,12 @@ export const useSuggestedValues = (suggestions: FormSuggestion | null | undefine
   if (suggestions.type === "literal") {
     return {
       useOneTime: () => {
+        // A field's literal values are static for its lifetime, so memoizing
+        // once keeps the normalized array referentially stable across renders.
+        const data = useMemo(() => normalizeSuggestions(suggestions.values), []);
```
bartv, line 41: 🤖 Claude (on behalf of Bart) - [Should-fix]: this `useMemo` references `suggestions.values` but declares empty deps. It's a latent staleness bug. It's unreachable today (the annotation is static per mounted field, and a changed shape remounts the field via its React `key`), and it slips past `eslint --max-warnings=0` only because `react-hooks/exhaustive-deps` can't see a `useMemo` nested inside the `useOneTime` object-property arrow (a top-level equivalent does warn). Depending on `[suggestions.values]` is a free fix - `normalizeSuggestions` is pure, so referential stability is preserved in the static case - and it removes the staleness risk plus the lint fragility right before §2/§3 introduce values that legitimately vary.

```suggestion
        const data = useMemo(() => normalizeSuggestions(suggestions.values), [suggestions.values]);
```

src/Core/Domain/ServiceInstanceModel.ts
```diff
@@ -83,11 +83,43 @@ export interface FormAttributeResult {
 }
 
 /**
- * Interface representing a suggestions that are stored in the web_suggested_values.
+ * Interface representing a single, normalized suggestion.
+ *
+ * This is the shape the form actually consumes: the `label` is shown to the user
+ * and searched on, while the `value` is what gets submitted to the API. Both are
+ * always strings — `normalizeSuggestions` coerces every raw entry (see
```
bartv: 🤖 Claude (on behalf of Bart) - [Nit]: em dash in the new JSDoc; repo convention is to avoid em/en dashes (use a hyphen). This is the only occurrence in the changed files.

```suggestion
 * always strings - `normalizeSuggestions` coerces every raw entry (see
```

changelogs/unreleased/7008-suggestion-label-value.yml
```diff
@@ -0,0 +1,6 @@
+description: Service instance form suggestions now support `{ label, value }` pairs, showing a human-friendly label while submitting the underlying value.
```
bartv: 🤖 Claude (on behalf of Bart) - [Nit / question]: this is the only unreleased changelog entry that embeds inline-code backticks in `description`; the neighbours are plain prose. Cosmetic only - just worth confirming the release-notes generator passes the backticks through acceptably, otherwise reword to "label/value pairs" to match the surrounding entries.

bartv (commented):

**Verdict: right direction, good to merge after one small fix.** This implements design §1 ("Split label and value") cleanly and nothing else, and - the main thing I was asked to check - the architecture it establishes is the correct foundation for the later design steps (§2-§5). I verified findings against the actual files and ran a second adversarial pass over each.

**Local gate (PR branch):** ESLint `--max-warnings=0` clean; `tsc --noEmit` clean for all PR files (the lone error is an uninstalled `@patternfly/react-drag-drop` in a non-PR file, present on master too); Vitest 27/27 pass.

**Is it the right direction for §2-§5?**

| Step | Verdict | Why |
|------|---------|-----|
| §1 split label/value | Done | One `normalizeSuggestions` coerces every flavor to `SuggestionValue[]`; controls show/search label, submit value; the old "return a fixed format" TODO in `FieldInput` is genuinely removed (normalization centralized in the hook). |
| §2 vars in `parameter_name` | Compatible, additive | Resolution belongs in the hook and must feed the cache key; today both the URL and `queryKey` use the raw name. Growing the hook to take form context is accretive, no rework of the normalized contract. |
| §3 `graphql` flavor | Strong foundation | `normalizeSuggestions(unknown)` already accepts exactly the two shapes §3 projects into, so a graphql branch reuses it verbatim; the `{useOneTime}` + branch-per-`type` shape admits a third branch (reusing the existing `useGraphQLRequest`). The narrowed `ResponseData` is local to the parameters branch and does not box graphql in. |
| §4 cascading fields | Nothing to undo | `FieldInput` already receives whole-form `formState` + its `path` (sibling reads + `self` scope), `shouldBeDisabled` is the natural seam for blocking, and label/value + forced `type=text` + string-submit-then-sanitize is exactly the cascading-graphql UUID behavior. No DictPath logic leaked into the feature. |
| §5 form tabs | Orthogonal | Pure layout concern; this per-field value pipeline neither helps nor harms it. |

**Inline:** one should-fix (`useMemo` deps, free one-liner) and two nits (em dash, changelog backtick). The previously-resolved threads (blur `editedRef`, `filter={displayValue}` UX, `type=text` for numeric) all check out. The numeric-suggestion-submits-a-string concern is **not** a real gap: create / PATCH / PUT / duplicate all run `sanitizeAttributes`, and the expert path is a raw JSON editor.

**To carry into the next steps (not blockers here):**
1. Make `FormSuggestion` a discriminated union on `type` now, while only two variants exist, before optional fields proliferate in §3/§4. A runtime parse/validate at the annotation boundary is worth considering since the data arrives untyped.
2. Harden `resolveValue` before §3/§4 land: its typed-text-equals-label reverse lookup uses first-match, but graphql will project `label` from a non-unique attribute and `value` from a UUID, so a typed label could silently bind to one of N UUIDs. Make a list selection authoritative.
3. Home the future DictPath get-only evaluator in `src/Core/Language` (standalone, shared), porting vectors from inmanta-core `test_dict_path.py` - nothing here pre-empts it.

Nice, well-scoped first step with strong test coverage.

src/Data/Queries/Slices/ServiceInstance/FormSuggestions/useSuggestions.ts
AronH99: The react-hooks/exhaustive-deps rule is applied correctly here, nothing gets flagged because the suggestions is a parameter of the useSuggestedValues which is non reactive parameter so nothing gets flagged here, so we should not add it to the dependency array

bartv (approved):

PR #7052, opened 2026-06-29
Fix stuck action toggle highlight and modal drag-out close

### LINKED ISSUE #7047: Setting the state (expert mode) for an instance gives a visual bug whenever utilizing expert actions
1) There is a bug where whenever on expert mode and you try to set an expert action state the list with actions and modal keeps being highlighted/open
2) Also a bug related to the modal is one where we start clicking inside of the modal and whenever we end clicking outside of the global modal it still closes => unintended behaviour

Image

AronH99:
# Description

- Expert and destroy action toggles in the instance-details actions menu now collapse when their confirmation modal opens, instead of staying highlighted behind it
- ModalProvider now decides outside-clicks on `mousedown`, so a press that starts inside the dialog and releases on the backdrop no longer closes the modal
- Added some tests to verify regression
- Made a small sensible refactor to remove pf v6 class select

closes https://github.com/inmanta/web-console/issues/7047

Screenshot 2026-06-29 at 11 35 35

### DIFF
--- a/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
+++ b/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
@@ -13,7 +13,7 @@ interface Props {
   instance_id: string;
   service_entity: string;
   version: ParsedNumber;
-  onClose: () => void;
+  collapseToggle: () => void;
   setInterfaceBlocked: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
@@ -25,7 +25,7 @@ interface Props {
  *  @prop {string} instance_id - the hashed id of the instance
  *  @prop {string} service_entity - the service entity type of the instance
  *  @prop {ParsedNumber} version - the current version of the instance
- *  @prop {function} onClose - callback method when the modal gets closed
+ *  @prop {function} collapseToggle - collapses the dropdown toggle when the modal opens
  *  @prop {React.Dispatch<React.SetStateAction<boolean>>} setInterfaceBlocked - setState variable to block the interface when the modal is opened.
  *  This is meant to avoid clickEvents triggering the onOpenChange from the dropdown to shut down the modal.
  * @returns {React.FC<Props>} A React Component displaying the Destroy Dropdown Item
@@ -35,7 +35,7 @@ export const DestroyAction: React.FC<Props> = ({
   instance_display_identity,
   instance_id,
   version,
-  onClose,
+  collapseToggle,
   setInterfaceBlocked,
 }) => {
   const { triggerModal, closeModal } = useContext(ModalContext);
@@ -59,7 +59,9 @@ export const DestroyAction: React.FC<Props> = ({
       iconVariant: "danger",
       cancelCb: closeModalCallback,
     });
-    onClose();
+    // Collapse the toggle now: setInterfaceBlocked(true) suppresses the dropdown's
+    // onOpenChange, so it can't collapse the toggle itself while the modal is open.
+    collapseToggle();
   };
 
   /**
--- a/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
+++ b/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
@@ -58,6 +58,7 @@ export const DestroyAction: React.FC<Props> = ({
       ),
       iconVariant: "danger",
     });
+    onClose();
   };
 
   /**
--- a/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
+++ b/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
@@ -57,6 +57,7 @@ export const DestroyAction: React.FC<Props> = ({
         />
       ),
       iconVariant: "danger",
+      cancelCb: closeModalCallback,
     });
     onClose();
   };
@@ -67,8 +68,7 @@ export const DestroyAction: React.FC<Props> = ({
   const closeModalCallback = useCallback(() => {
     closeModal();
     setInterfaceBlocked(false);
-    onClose();
-  }, [closeModal, setInterfaceBlocked, onClose]);
+  }, [closeModal, setInterfaceBlocked]);
 
   return (
     <>
--- a/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/InstanceActions.tsx
+++ b/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/InstanceActions.tsx
@@ -87,7 +87,7 @@ export const InstanceActions: React.FC = () => {
               instance_id={instance.id}
               service_entity={instance.service_entity}
               version={instance.version}
-              onClose={() => setIsDropdownOpen(false)}
+              onClose={() => setIsExpertDropdownOpen(false)}
               setInterfaceBlocked={setBlockedInterface}
             />
             {!instance.deleted && expertStateTargets.length > 0 && (
@@ -101,7 +101,7 @@ export const InstanceActions: React.FC = () => {
                   instance_id={instance.id}
                   service_entity={instance.service_entity}
                   version={instance.version}
-                  onClose={() => setIsDropdownOpen(false)}
+                  onClose={() => setIsExpertDropdownOpen(false)}
                   setInterfaceBlocked={setBlockedInterface}
                 />
               </>
--- a/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/InstanceActions.tsx
+++ b/src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/InstanceActions.tsx
@@ -87,7 +87,7 @@ export const InstanceActions: React.FC = () => {
               instance_id={instance.id}
               service_entity={instance.service_entity}
               version={instance.version}
-              onClose={() => setIsExpertDropdownOpen(false)}
+              collapseToggle={() => setIsExpertDropdownOpen(false)}
               setInterfaceBlocked={setBlockedInterface}
             />
             {!instance.deleted && expertStateTargets.length > 0 && (
@@ -101,7 +101,7 @@ export const InstanceActions: React.FC = () => {
                   instance_id={instance.id}
                   service_entity={instance.service_entity}
                   version={instance.version}
-                  onClose={() => setIsExpertDropdownOpen(false)}
+                  collapseToggle={() => setIsExpertDropdownOpen(false)}
                   setInterfaceBlocked={setBlockedInterface}
                 />
               </>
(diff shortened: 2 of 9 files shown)


src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/InstanceActions.tsx
```diff
@@ -101,7 +101,7 @@ export const InstanceActions: React.FC = () => {
                   instance_id={instance.id}
                   service_entity={instance.service_entity}
                   version={instance.version}
-                  onClose={() => setIsDropdownOpen(false)}
+                  onClose={() => setIsExpertDropdownOpen(false)}
```
bartv: 🤖 Claude (on behalf of Bart) - [Blocker]: re-pointing both `onClose` handlers at `setIsExpertDropdownOpen` is correct (`DestroyAction` and `ExpertStateTransfer` both live in the expert dropdown). But the toggle-collapse fix is **asymmetric**: `DestroyAction.onDestroySelect` now calls `onClose()` right after `triggerModal`, whereas `ExpertStateTransfer.onStateSelect` does not.

Because `setInterfaceBlocked(true)` deliberately suppresses the dropdown's `onOpenChange` (the gate is `!blockedInterface`), nothing collapses the expert toggle when a state transfer is triggered, so it still stays highlighted behind the modal - which is exactly the behavior the changelog says is fixed ("the expert **and** destroy action toggles no longer stay highlighted").

Fix: add `onClose();` at the end of `ExpertStateTransfer.onStateSelect` (right after the `triggerModal({ ... })` call), mirroring `DestroyAction.onDestroySelect`. A regression assertion (expert toggle `aria-expanded=false` after selecting a state target) would lock this in.

src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
```diff
@@ -58,6 +58,7 @@ export const DestroyAction: React.FC<Props> = ({
       ),
       iconVariant: "danger",
     });
+    onClose();
```
bartv: 🤖 Claude (on behalf of Bart) - [Should-fix]: this `triggerModal` call has no `cancelCb`, unlike `ExpertStateTransfer.onStateSelect` and `StateAction.onSelect`. `closeModal` only resets `setInterfaceBlocked(false)` through `cancelCb` (which runs `closeModalCallback` = reset + `onClose()`).

This PR newly lets the backdrop close the modal on `mousedown` (and the X button already could), so dismissing the destroy modal that way runs `closeModal()` but never resets `blockedInterface` - it stays stuck `true`, which dead-locks the expert dropdown's `onOpenChange` gate (it can still be force-toggled, but no longer auto-closes on outside click).

Fix: add `cancelCb: closeModalCallback` to the `triggerModal({ ... })` call above, consistent with the other two actions.

src/UI/Root/Components/ModalProvider/ModalProvider.tsx
```diff
@@ -148,19 +154,21 @@ export const ModalProvider: React.FC<PropsWithChildren> = ({ children }) => {
       return;
     }
 
-    const handleClick = (e: MouseEvent) => {
+    // Close only when the press starts on the backdrop, not inside the dialog,
+    // so a press that begins in the dialog and ends on the backdrop is ignored.
+    const handleMouseDown = (e: MouseEvent) => {
```
bartv, line 159: 🤖 Claude (on behalf of Bart) - [Should-fix]: `handleMouseDown` has no button guard, so a right-click or middle-click on the backdrop now closes the modal (the old `click` listener effectively only reacted to a primary click). Restrict to the primary button:

```suggestion
    const handleMouseDown = (e: MouseEvent) => {
      if (e.button !== 0) {
        return;
      }
```

src/UI/Root/Components/ModalProvider/ModalProvider.test.tsx
```diff
@@ -140,6 +140,51 @@ describe("ModalProvider", () => {
...
+      expect(screen.queryByTestId("GlobalModal")).toBeNull();
+    });
+
+    it("stays open when the press starts inside the dialog and ends on the backdrop", async () => {
+      setup();
+
+      await userEvent.click(screen.getByText("Click me"));
+
+      // Press begins on the modal content (e.g. selecting text) and is released
+      // on the backdrop — must not close the modal. The decision is made on
```
bartv: 🤖 Claude (on behalf of Bart) - [Nit]: em dash in the comment; repo convention is to avoid em/en dashes (use a hyphen).

```suggestion
      // on the backdrop - must not close the modal. The decision is made on
```

bartv (changes requested):

**Verdict: solid approach, but one blocker before merge.** Gate is green (ESLint clean, types clean, `ModalProvider` tests 16/16), and the core ideas are right - but the headline "toggle no longer stays highlighted" fix only covers part of the expert menu.

**🔴 Blocker - the fix is asymmetric (see inline on `InstanceActions.tsx`):** `DestroyAction` got an `onClose()` on modal-open, but `ExpertStateTransfer` did not. Since `setInterfaceBlocked(true)` suppresses the dropdown's auto-close, triggering an **expert state transfer** still leaves the expert toggle highlighted behind the modal - the exact case the changelog claims is fixed. One-line fix in `ExpertStateTransfer.onStateSelect`.

**🟠 Should-fix:**
- `DestroyAction`'s `triggerModal` has no `cancelCb`, so the new backdrop-mousedown close (or the X) can leave `blockedInterface` stuck `true` and dead-lock the expert dropdown gate (inline).
- `handleMouseDown` needs an `e.button === 0` guard, otherwise right/middle-click on the backdrop closes the modal (inline, with suggestion).
- Em dash in a test comment (inline, with suggestion).
- **No regression test** for the toggle-collapse behavior itself - the existing expert-action tests only assert the dialog opens, never the toggle's `aria-expanded`. A test asserting the toggle collapses after select would have caught the blocker (and the state-transfer case would fail until it's fixed, which is the point).

**🟡 Question:** switching the outside-close from `click` to `mousedown` means it now fires on press rather than release. That's the right call for the drag-out fix (a press starting inside the dialog and ending on the backdrop no longer closes); just confirming the press-to-dismiss behavior on the backdrop is intended.

**🟢 Nit:** the "press inside, release on backdrop" test ends with `fireEvent.click(getBackdrop())`, which is inert now that the handler only listens on `mousedown` - the test would still pass if the inside-press guard broke. Using `fireEvent.mouseUp(getBackdrop())` makes the gesture faithful (the other two backdrop tests are solid).

**Confirmed-good:** the `setIsExpertDropdownOpen` scoping is correct; the `BACKDROP_CLASS` + `backdropClassName` approach correctly avoids PatternFly's version-specific internal class (verified PF appends it onto the backdrop element) and is the right pattern; the effect's listener swap introduces no new staleness; the changelog is otherwise well-formed (just tighten the wording once the expert-transfer case is actually fixed).

Once the expert-state-transfer `onClose()` is added (plus ideally the `cancelCb` and button guard), this is good to go.

src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/InstanceActions.tsx
AronH99: made it consistent throughout the 2 files and added the test but these comments are a little weird/far fetched to be fair since it works perfectly fine, the test is useful though

src/UI/Root/Components/ModalProvider/ModalProvider.tsx
AronH99: solid comment

src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/InstanceActions.tsx
```diff
@@ -87,7 +87,7 @@ export const InstanceActions: React.FC = () => {
               instance_id={instance.id}
               service_entity={instance.service_entity}
               version={instance.version}
-              onClose={() => setIsDropdownOpen(false)}
+              onClose={() => setIsExpertDropdownOpen(false)}
```
bartv: 🤖 Claude, on behalf of Bart.

The expert/destroy toggles now collapse **deterministically**: their select handlers call `onClose()` on modal-open, which runs `setIsExpertDropdownOpen(false)` directly and bypasses the `!blockedInterface` guard on `onOpenChange`. That's the right fix.

The normal `DeleteAction` (further down) and `StateAction` weren't given the same treatment - they still fire `onClose` only from their close callback, so the normal toggle's collapse still relies on PatternFly's `onOpenChange(false)` winning against `setInterfaceBlocked(true)` - the exact race this PR eliminates for expert/destroy. It happens to collapse in jsdom today so it isn't a confirmed bug, but it's the same latent fragility living on in the normal menu.

Non-blocking: consider applying the `onClose()`-on-open pattern to `DeleteAction`/`StateAction` too, or add a short note on why only expert/destroy needed it.

src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/Actions/DestroyAction.tsx
```diff
@@ -57,7 +57,9 @@ export const DestroyAction: React.FC<Props> = ({
         />
       ),
       iconVariant: "danger",
+      cancelCb: closeModalCallback,
     });
+    onClose();
```
bartv: 🤖 Claude, on behalf of Bart.

Nit (naming/docs): `onClose()` is now invoked here when the modal *opens* (to collapse the toggle), but the `onClose` prop's JSDoc above still reads *"callback method when the modal gets closed."* The same prop also still fires on the actual close in `DeleteAction`/`StateAction`, so one prop name now means two opposite things across the four actions.

Consider renaming it for the expert/destroy actions (e.g. `onOpen` / `collapseToggle`), or at minimum update the JSDoc here and in `ExpertStateTransfer.tsx` so the timing is accurate.

bartv (commented):

Re-reviewed at `83bafa5`. **The blocker from my previous review is resolved:** `ExpertStateTransfer` now calls `onClose()` on modal-open (matching `DestroyAction`), and both pass `cancelCb: closeModalCallback`, so the expert/destroy toggles collapse deterministically and the interface no longer stays blocked. The `e.button === 0` guard is in place, the `.pf-v6-c-backdrop` dependency was replaced with an owned `backdropClassName`, and the tests faithfully reproduce the drag-out and stuck-highlight cases (`mouseDown`/`mouseUp` on the right targets, `aria-expanded` assertions).

**Verdict: good to merge.** Two non-blocking follow-ups left inline:
- Toggle-collapse consistency: the normal `DeleteAction`/`StateAction` still rely on the `onOpenChange` race rather than the new deterministic `onClose()`-on-open pattern.
- The `onClose` prop name/JSDoc is now misleading for expert/destroy (fires on open, not close).

Left as COMMENT rather than a gate - fold in or defer as you see fit. (Note: my earlier "Changes requested" review is now stale; dismiss it when you approve.)

src/Slices/ServiceInstanceDetails/UI/Components/InstanceActions/InstanceActions.tsx
bartv: 🤖 Claude, on behalf of Bart.

Fair - it did work in practice; the point was that the two expert actions handled the collapse differently, so it was one refactor away from silently regressing. `ExpertStateTransfer.onStateSelect` now calls `onClose()` too and the `aria-expanded=false` assertion locks it in. Addressed - resolving.

AronH99:
Adjusted the inconsistencies and the naming issues now, I honestly thought about this as well, at first didn't want to touch too much of the code but I guess this make sense

bartv (approved):

PR #7054, opened 2026-06-29
Align and soften version dropdown timestamps in attribute comparison

### LINKED ISSUE #7044: Version compare of service instances timestamp should be less obvious and responsive
For the instance details compare improve responsiveness timestamp + timestamp itself should be less obvious

AronH99:
# Description

- Version dropdown timestamps now align in a column, sized to the widest version via a derived ch min-width
- Timestamps render smaller (body--sm token) and subtler (text--color--subtle token), with the label outline removed

=> This should be way more subtle without doing crazy adjustments

closes https://github.com/inmanta/web-console/issues/7044

Screenshot 2026-06-29 at 14 58 26

### DIFF
--- a/src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesCompare.tsx
+++ b/src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesCompare.tsx
@@ -7,7 +7,7 @@ import {
   FormSelect,
   FormSelectOption,
   SelectOptionProps,
-  Label,
+  Content,
 } from "@patternfly/react-core";
 import { InstanceAttributeModel } from "@/Core";
 import { InstanceLog } from "@/Core/Domain/HistoryLog";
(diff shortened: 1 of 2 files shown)


src/Slices/ServiceInstanceDetails/UI/Components/AttributesComponents/AttributesCompare.tsx
```diff
@@ -53,21 +53,33 @@ export const AttributesCompare: React.FC<Props> = ({ instanceLogs, selectedVersi
...
+        >
+          <Content style={{ minWidth: `${versionColWidth}ch` }}>{version}</Content>
+          <Content
+            style={{
+              fontSize: "var(--pf-t--global--font--size--body--sm)",
+              color: "var(--pf-t--global--text--color--subtle)",
+            }}
+          >
+            {datePresenter.getFull(timestamp)}
+          </Content>
```
bartv: 🤖 Claude (on behalf of Bart) - [Should-fix, idiom]: a bare `<Content>` (no `component`) renders PatternFly's prose-container `div` variant, which is meant to wrap rich text blocks rather than a single value. For "smaller + subtler" text the repo idiom is `<Content component={ContentVariants.small}>` (see `Description.tsx`, `ServiceItem.tsx`), which pulls the small-body font size from PF and lets you drop the hand-rolled `fontSize` var:

```tsx
<Content
  component={ContentVariants.small}
  style={{ color: "var(--pf-t--global--text--color--subtle)" }}
>
  {datePresenter.getFull(timestamp)}
</Content>
```

(Add `ContentVariants` to the `@patternfly/react-core` import.) The version element above could likewise be a `span`/`ContentVariants.p` rather than a bare `Content`. Functionally what you have works - this is about using the intended component variant. The `color`/token-via-inline-`style` part is fine and matches existing usage elsewhere.

bartv (commented):

**Verdict: looks good, mergeable - one optional idiom note.** Gate is green (ESLint clean, types clean), and the change is functionally correct:
- `versionColWidth` correctly guards against `Math.max()` returning `-Infinity` on an empty array;
- `String(version).length` + the `ch` unit is a sound alignment basis for the numeric (`ParsedNumber`) version;
- the `Label` import is cleanly removed and `Content` is a valid `react-core` export;
- consuming PF design tokens via inline `style` on your own element is an established pattern here (not a PatternFly-CSS override);
- no markup hazard (these render inside `SearchSelect`'s menuitem/toggle, not a native `<option>`), and the existing `Page.test.tsx` still covers the version + timestamp render path.

**🟠 Should-fix (idiom only, inline):** use `<Content component={ContentVariants.small}>` for the timestamp instead of a bare prose-container `Content` (also removes the manual `fontSize` var).

**🟡 Optional nit:** the `versionColWidth` computation is pure and untested; if you want to lock the empty-input behavior, extract a tiny helper and add a two-case unit test. Not required for a presentational change.

Nice, clean improvement.

bartv (approved):

PR #7056, opened 2026-06-30
Make chevron directions consistent across the console

### LINKED ISSUE #7046: Chevron icons after last pf update are very much broken/ugly/confusing
See if we can improve this ourselves before release

AronH99:
# Description

- Add RhMicronsCaretDownIcon override that renders the classic angle-down chevron, wired in via a resolve.alias in vite.config.ts redirecting PatternFly's caret icon import
- Add patternfly-chevron-fix.css overriding PatternFly's rotation custom properties so the disclosure direction reads right (collapsed) → down (expanded)
- Import the new stylesheet in src/index.tsx

@bartv so for what this PR is I think it's honestly a solid solution but I don't know if we should do this, I don't really like these kinds of things since they are quite "unstable".
I made this PR along with claude because you wanted to see this fixed for the next release and if you look at the screenshots it seems to be doing fine but I'd still prefer if we would wait for PatternFly themselves to fix this.
I think this is what it used to be no?
We now overwrite the chevron icons with the old one and add custom styling to fix the directions of the chevrons.
What's your take on this?

closes https://github.com/inmanta/web-console/issues/7046

Screenshot 2026-06-30 at 09 13 53
Screenshot 2026-06-30 at 09 14 13

### DIFF
--- a/changelogs/unreleased/7046-patternfly-chevron-fix.yml
+++ b/changelogs/unreleased/7046-patternfly-chevron-fix.yml
@@ -1,4 +1,4 @@
-description: Aligned the custom expand/collapse toggles on the order details and callbacks rows with PatternFly's chevron direction (down when collapsed, up when expanded), so they are consistent with the rest of the console's expandable components.
+description: Made the expand/collapse chevron directions consistent across the console.
 issue-nr: 7046
 change-type: patch
 destination-branches: [master]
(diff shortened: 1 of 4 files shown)


AronH99:
This is on hold at the moment, i am figuring out what was initially wrong with the directions of the chevrons

AronH99:
I fixed the inconsistencies now @bartv 
I deleted all custom styling so we don't fight PatternFly on their new design direction.
Now everything should go either up or down with the new icon style (see screenshot)
The fix was mainly few tweaks on our sides and a dedupe in the yarn.lock
This should do it then, no more weird left or right where you don't expect it.

Screenshot 2026-06-30 at 14 24 03

bartv (approved):

PR #7059, opened 2026-07-01
Add Copy raw value control to shared CodeEditor

### LINKED ISSUE #7045: Add raw copy button to the Code Editors
Add a raw copy button to the code editors so that it is possible to also copy the raw api value that get's passed along 

Discussion:
https://inmanta.slack.com/archives/C045Y42N49W/p1782826131280779

AronH99:
# Description

So basically what happened is that the CodeEditor component now supports copying a rawValue (outlined copy button) of the code passed to it (json/xml...)
Adjusted tests wherever needed and adjusted the implementation of CodeEditor throughout web-console wherever needed to support this small improvement as well (limited to only 1 place where it currently made sense to do so).

closes https://github.com/inmanta/web-console/issues/7045

Screenshot 2026-07-01 at 11 34 49


src/UI/Components/CodeEditor/CodeEditor.tsx
```diff
@@ -37,10 +51,15 @@ export const CodeEditor: React.FC<Props> = ({
   const isAutoHeight = height === undefined;
   const resolvedHeight = isAutoHeight ? getAutoHeight(code, isExpanded) : height;
 
+  // Offer a raw-value copy only when the raw form is meaningfully different from
+  // the formatted code on screen.
+  const showRawCopy = isCopyEnabled && rawValue !== undefined && rawValue !== code;
```
AronH99, line 56: 🤖 Claude · [Should-fix]: this `showRawCopy` gating (`rawValue !== undefined && rawValue !== code`, also gated on `isCopyEnabled`) is the actual new behaviour of the wrapper, but it isn't covered anywhere — `CodeEditorControls.test.tsx` only tests `CodeEditorRawCopyControl` in isolation. `CodeEditor.test.tsx` already has a `hides the copy control when isCopyEnabled is false` case; I'd add sibling cases there: raw button appears when `rawValue` differs from `code`, and is hidden when `rawValue === code`, when `rawValue` is omitted, and when `isCopyEnabled={false}`. That's the branch most likely to silently regress on a future refactor.

src/UI/Components/CodeEditorControls/CodeEditorControls.tsx
```diff
@@ -1,11 +1,15 @@
 import React from "react";
 import { CodeEditorControl } from "@patternfly/react-code-editor";
-import { CopyIcon, ExpandIcon, CompressIcon } from "@patternfly/react-icons";
+import { CopyIcon, OutlinedCopyIcon, ExpandIcon, CompressIcon } from "@patternfly/react-icons";
```
AronH99, line 3: 🤖 Claude · [Question / UX — human call]: the raw control uses `OutlinedCopyIcon` right next to the filled `CopyIcon`. Two near-identical copy glyphs side-by-side lean entirely on the tooltip/aria-label to tell them apart, which isn't obvious at a glance. Tooltips do mitigate it, and this matches the screenshot, so I'd leave the final call to the human reviewer — just flagging whether the filled-vs-outlined distinction is discoverable enough, or whether a labelled control / different icon would read clearer. Not blocking.

src/Slices/OrderDetails/UI/OrderDetailsRow.tsx
```diff
@@ -87,7 +92,8 @@ export const OrderDetailsRow: React.FC<Props> = ({
                 <DescriptionListDescription>
                   {row.config && Object.keys(row.config).length ? (
                     <CodeEditor
-                      code={JSON.stringify(row.config, null, 2)}
+                      code={jsonFormatter.format(row.config)}
+                      rawValue={JSON.stringify(row.config)}
```
AronH99, line 96: 🤖 Claude · [Question]: here `rawValue={JSON.stringify(row.config)}` — but `row.config` is already an object from the API, so there's no original stored string; the "raw" value is just the compact re-serialization of what's displayed. That's consistent with how the classifier treats object values (`rawValue: JSON.stringify(value)`), so it's fine — just confirming that copying compact JSON (vs the pretty view) is the intended value here, since the issue framed this as copying "the raw api value". Same question applies to `row.attributes || row.edits` a few lines down.

AronH99 (commented):
## 🤖 Claude pre-review — PR #7059 — Add Copy raw value control to shared CodeEditor

**Type & intent:** feat / minor-improvement — give the shared `CodeEditor` a second copy button that copies the *original un-formatted* string (compact JSON/XML) behind the pretty-printed view. Closes #7045.

**Surface area:**
- Shared UI (blast radius): `src/UI/Components/CodeEditor`, `CodeEditorControls` (+ test), `AttributeList`, `src/UI/words.tsx`
- Shared Data: `src/Data/Common/AttributeClassifier` (`AttributeClassifier`, `ClassifiedAttribute`, `Mock`)
- Feature: `src/Slices/OrderDetails/UI/OrderDetailsRow`

**Size & balance:** +156 / −22, 10 files. Test delta present (`CodeEditorControls.test.tsx` +30), proportionate to the change.

**Blast radius:** touches shared component + shared classifier — rippled correctly (new `Formatted<Kind>` type, `rawValue` threaded through, `code`→`value` prop rename on the copy control). Confined behaviourally: the raw button only appears when a `rawValue` is passed *and* differs from `code`, so existing call sites are unaffected.

**Risk hotspots:** test coverage of the new gating logic; UX of two near-identical copy icons; semantics of "raw" for object-valued payloads.

**Verdict: Comment (approve-with-nits).**
Genuinely clean, focused, and unusually well-documented (the `Formatted<Kind>` and control docstrings explain *why*, and the classifier comments honestly note that object values have no byte-exact source). Nothing blocking. What would flip it to request-changes: if the Jenkins e2e run (currently PENDING) fails on something real, or if the raw-copy gating in `CodeEditor.tsx` turns out to regress an existing call site.

Synopsis: 1 should-fix (no `CodeEditor`-level test for the new `showRawCopy` gating — the unit test only covers the control in isolation), plus 1 UX question (two similar copy icons side-by-side, discoverability rests on tooltips — a product/taste call for the human reviewer) and 1 clarifying question (what "raw" means for object config in OrderDetails) below. The `code`→`value` rename on `CodeEditorCopyControl` is a reasonable disambiguation given the new sibling control.

src/UI/Components/CodeEditorControls/CodeEditorControls.tsx
AronH99: It has a tooltip to this is fine

src/Slices/OrderDetails/UI/OrderDetailsRow.tsx
AronH99: this is fine indeed

bartv (approved):
_Posted by Claude on behalf of Bart._

Reviewed the diff and verified locally: prop rename (`code`→`value`) has no stale callers, the new mandatory `rawValue` on `Formatted<Kind>` is set at all `Json`/`Xml` construction sites, `rawValue` is not leaked to the DOM, the `OrderDetailsRow` switch to `jsonFormatter.format` is behavior-preserving (identical to `JSON.stringify(x, null, 2)`), and the `showRawCopy` gating is correct. All 24 CodeEditor/CodeEditorControls tests pass. LGTM.

PR #7061, opened 2026-07-01
Fix blank dashboard flash when deleting an environment

### LINKED ISSUE #7058: Environment deletion redirection bug
When an environment is deleted using the web-console, you are directed to the page for the deleted environment. That results in an error of course given that the environment was deleted. The problem resolves after a few seconds.

https://github.com/user-attachments/assets/20f24ee1-2b27-471f-a1de-6bf03ba68d60

AronH99:
# Description

- On delete, refetch the environment preview list (feeds redirect logic) and reset the REST environment list (header selector, user management) so the deleted env is gone app-wide
- Navigate directly to a surviving environment instead of the neutral Dashboard URL, avoiding the missing-env bounce that flashed a blank/errored dashboard
- Fall back to the Create Environment page (via PageFrame Provider, so its "no environments" toast still fires) when the last environment is deleted

closes https://github.com/inmanta/web-console/issues/7058

https://github.com/user-attachments/assets/775be92a-64d7-403f-aafe-233243386cdc

### DIFF
--- a/src/Slices/Settings/UI/Tabs/Environment/Components/ConfirmationForm.tsx
+++ b/src/Slices/Settings/UI/Tabs/Environment/Components/ConfirmationForm.tsx
@@ -12,8 +12,14 @@ import {
 import { useQueryClient } from "@tanstack/react-query";
 import styled from "styled-components";
 import { FlatEnvironment } from "@/Core";
-import { useClearEnvironment, useDeleteEnvironment, getEnvironmentsKey } from "@/Data/Queries";
+import {
+  useClearEnvironment,
+  useDeleteEnvironment,
+  getEnvironmentsKey,
+  GetEnvironmentPreviewKey,
+} from "@/Data/Queries";
 import { AppAlert } from "@/UI/Components";
+import { DependencyContext } from "@/UI/Dependency";
 import { ModalContext } from "@/UI/Root/Components/ModalProvider";
 import { useNavigateTo } from "@/UI/Routing";
 import { words } from "@/UI/words";
@@ -35,8 +41,10 @@ interface Props {
  */
 export const ConfirmationForm: React.FC<Props> = ({ environment, type }) => {
   const { closeModal } = useContext(ModalContext);
+  const { environmentHandler } = useContext(DependencyContext);
   const navigateTo = useNavigateTo();
   const client = useQueryClient();
+  const allEnvironments = environmentHandler.useAll();
 
   const [candidateEnv, setCandidateEnv] = useState("");
   const [errorMessage, setErrorMessage] = useState<null | string>(null);
@@ -46,13 +54,27 @@ export const ConfirmationForm: React.FC<Props> = ({ environment, type }) => {
   const redirectToHome = () => navigateTo("Dashboard", undefined);
 
   const deleteEnv = useDeleteEnvironment(environment.id, {
-    onSuccess: () => {
-      //reset the queries removes the cache which improves the ux when navigating back to the environments page,
-      // without it the user won't see loading state and will see the old data for a split second and then removed env will be removed from the view
+    onSuccess: async () => {
+      // reset removes the cache so the header env selector / user management list show a loading
+      // state instead of briefly rendering the deleted env. Fire-and-forget: neither drives the redirect.
       client.resetQueries({ queryKey: getEnvironmentsKey.root() });
 
+      // Refresh the preview list (feeds the header env preview and PageFrame/Provider) so the deleted
+      // env is gone app-wide. Awaited so the empty-list case below sees an up-to-date allEnvironments.
+      await client.refetchQueries({ queryKey: GetEnvironmentPreviewKey.root() });
+
       closeModal();
-      redirectToHome();
+
+      // Navigate straight to a surviving env instead of the neutral Dashboard URL, so we skip the
+      // Provider's missing-env bounce that would otherwise flash a blank Dashboard. When no envs are
+      // left, fall back to the Provider (redirectToHome) so its "no environments" toast still fires.
+      const remaining = allEnvironments.filter((env) => env.id !== environment.id);
+
+      if (remaining.length > 0) {
+        navigateTo("Dashboard", undefined, `?env=${remaining[0].id}`);
+      } else {
+        redirectToHome();
+      }
     },
     onError: (error) => {
       setErrorMessage(error.message);
--- a/src/Slices/Settings/UI/Tabs/Environment/Components/ConfirmationForm.tsx
+++ b/src/Slices/Settings/UI/Tabs/Environment/Components/ConfirmationForm.tsx
@@ -51,7 +51,7 @@ export const ConfirmationForm: React.FC<Props> = ({ environment, type }) => {
   const [isBusy, setIsBusy] = useState(false);
   const validated = environment.name === candidateEnv ? "success" : "default";
 
-  const redirectToHome = () => navigateTo("Dashboard", undefined);
+  const redirectToHome = () => navigateTo("Dashboard", undefined, undefined, { replace: true });
 
   const deleteEnv = useDeleteEnvironment(environment.id, {
     onSuccess: async () => {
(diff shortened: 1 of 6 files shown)


src/Slices/Settings/UI/Tabs/Environment/Components/ConfirmationForm.tsx
```diff
@@ -46,13 +54,27 @@ export const ConfirmationForm: React.FC<Props> = ({ environment, type }) => {
...
-    onSuccess: () => {
-      //reset the queries removes the cache which improves the ux when navigating back to the environments page,
-      // without it the user won't see loading state and will see the old data for a split second and then removed env will be removed from the view
+    onSuccess: async () => {
+      // reset removes the cache so the header env selector / user management list show a loading
+      // state instead of briefly rendering the deleted env. Fire-and-forget: neither drives the redirect.
       client.resetQueries({ queryKey: getEnvironmentsKey.root() });
 
+      // Refresh the preview list (feeds the header env preview and PageFrame/Provider) so the deleted
+      // env is gone app-wide. Awaited so the empty-list case below sees an up-to-date allEnvironments.
```
AronH99: 🤖 Claude · [Should-fix]: this comment misstates what the `await` does. `allEnvironments` here is the render-time snapshot captured by the `onSuccess` closure — awaiting `refetchQueries` updates the query *cache*, not this local variable, so `allEnvironments` below is NOT refreshed by it. The `remaining` computation is correct anyway because it filters the snapshot by the known deleted id. What the `await` actually buys you is on the *empty* path: `redirectToHome()` lets `PageFrame/Provider` decide Dashboard-vs-CreateEnvironment from `environmentHandler.useAll()`, and awaiting the preview refetch gives that list time to drop the deleted env before the Provider re-renders. Suggest rewording so the next reader doesn't assume the local `allEnvironments` becomes up-to-date.

src/Slices/Settings/UI/Tabs/Environment/Components/ConfirmationForm.tsx
```diff
@@ -46,13 +54,27 @@ export const ConfirmationForm: React.FC<Props> = ({ environment, type }) => {
...
+
       closeModal();
-      redirectToHome();
+
+      // Navigate straight to a surviving env instead of the neutral Dashboard URL, so we skip the
+      // Provider's missing-env bounce that would otherwise flash a blank Dashboard. When no envs are
+      // left, fall back to the Provider (redirectToHome) so its "no environments" toast still fires.
+      const remaining = allEnvironments.filter((env) => env.id !== environment.id);
+
+      if (remaining.length > 0) {
```
AronH99, line 76: 🤖 Claude · [Should-fix]: this branch — navigate to a surviving env vs. fall back to Create Environment — is the actual behavior change in the PR, and it has no test. `Actions.test.tsx`'s delete test only asserts the DELETE request fired (`counter === 1`); it never checks where you land. Worth adding two cases: (1) with other envs present, after delete the route becomes `?env=<remaining>`; (2) deleting the last env redirects to CreateEnvironment. That also locks down the empty-case timing (the Provider reading a freshly-emptied `useAll()`), which is currently only manually verified.

src/Slices/Settings/UI/Tabs/Environment/Components/ConfirmationForm.tsx
```diff
@@ -46,13 +54,27 @@ export const ConfirmationForm: React.FC<Props> = ({ environment, type }) => {
...
+        navigateTo("Dashboard", undefined, `?env=${remaining[0].id}`);
```
AronH99: 🤖 Claude · [Question]: `navigateTo` pushes a history entry, so pressing Back after a delete returns to the deleted env's URL (`?env=<deletedId>`), which then bounces through the Provider. The Provider's own redirect used `replace` for exactly this reason. Now that we navigate directly, is a pushed entry to the now-dead URL intended? If not, it'd be worth giving `navigateTo` a `replace` option (it's pre-existing behavior for `redirectToHome`, just more visible here). Not blocking — flagging as a call. :)

AronH99 (commented):
## 🤖 Claude pre-review — PR #7061 — Fix blank dashboard flash when deleting an environment

**Context map**
- **Type & intent:** `fix` — deleting an environment no longer flashes a blank/errored Dashboard before landing somewhere valid (closes #7058).
- **Surface area:** one slice — `Slices/Settings/UI/Tabs/Environment/Components/ConfirmationForm.tsx` — plus a changelog entry. Logic *reads* shared state: `environmentHandler` (`@/UI/Dependency`) and two env-list query keys (`getEnvironmentsKey`, `GetEnvironmentPreviewKey`).
- **Size & balance:** +33 / −5, 2 files. No test delta — and the branch this PR adds is exactly the untested part (see inline).
- **Blast radius:** low. Behavior is confined to one component; it reasons about app-wide caches but doesn't change them structurally.
- **Risk hotspots:** correctness of the redirect branching (survivor vs last-env-deleted), the awaited-refetch timing, and test coverage of the new navigation.

**Verdict: Comment** (no hard blocker — CodeQL green, Jenkins still pending)
The fix is correct and I verified the reasoning end-to-end. Two should-fixes below: the new navigation branch has no test (and it's the whole point of the PR), and one comment misstates why the `await` is there. A strict reviewer could reasonably gate merge on the missing test. 2 should-fix, 1 question inline. (Posted as COMMENT because GitHub blocks approve/request-changes on your own PR — the verdict lives here.)

src/Slices/Settings/UI/Tabs/Environment/Components/ConfirmationForm.tsx
```diff
@@ -35,24 +41,43 @@ interface Props {
...
-      // without it the user won't see loading state and will see the old data for a split second and then removed env will be removed from the view
+    onSuccess: async () => {
+      // reset removes the cache so the header env selector / user management list show a loading
+      // state instead of briefly rendering the deleted env. Fire-and-forget: neither drives the redirect.
       client.resetQueries({ queryKey: getEnvironmentsKey.root() });
 
+      // Refresh the preview list (feeds the header env preview and PageFrame/Provider) so the deleted
+      // env is gone app-wide. Awaited so the empty-list fallback below reaches the Provider only after
+      // the list has emptied. Note: allEnvironments here is the render snapshot and is NOT refreshed by
+      // this await — the filter below drops the deleted env, which is what makes `remaining` correct.
```
bartv: _Claude here, on behalf of Bart._

Our convention is no em dashes (`—`) in code or comments; use a regular `-` or restructure. Applies to the em dash in this comment.

```suggestion
      // this await; the filter below drops the deleted env, which is what makes `remaining` correct.
```

src/UI/Routing/Utils.ts
```diff
@@ -1,28 +1,38 @@
...
 
 /**
  * The useNavigateTo hook returns a navigateTo function which navigates to a route.
- * @param newSearch This string should start with a question mark "?".
+ * @param kind The route to navigate to.
+ * @param params The route parameters for that route.
+ * @param newSearch The query string to use; must start with a question mark "?". When omitted, the
+ *   current location's search is preserved.
+ * @param options React Router's `NavigateOptions` (e.g. `{ replace: true }` to replace the current
+ *   history entry instead of pushing a new one — useful for redirects).
```
bartv: _Claude here, on behalf of Bart._

Same as the other comment: please avoid the em dash (`—`) here and use a regular `-`.

```suggestion
 *   history entry instead of pushing a new one, useful for redirects).
```

bartv (commented):
_Claude here, on behalf of Bart._

Reviewed the fix; the logic looks solid. I traced the delete flow through `ConfirmationForm`, the real `EnvironmentHandler.useAll()`, the `PageFrame/Provider` redirect, and `useNavigateTo`: both the "navigate to a surviving env" and the "last env deleted -> fall back to the Provider" paths land on a valid state, the awaited preview refetch is correct, and the two new tests are consistent. No correctness concerns.

Only nit: two comments use em dashes (`—`), which we avoid in code/comments. Left inline suggestions on both lines that you can apply directly.

bartv (approved):

AronH99:
Not cherry-picking this to iso9. This fixes a regression introduced by the Home→Dashboard route consolidation, which is master-only. On iso9 the `Home` route still exists (`/`, `environmentRole: Forbidden`) and remains the post-delete redirect target, so the "directed to the deleted environment" bounce never happens there — the bug doesn't reproduce on iso9. Forcing the patch in would regress iso9's redirect toward master's env-required Dashboard flow.

Follow-up: updating the changelog's `destination-branches` to `[master]` accordingly.

PR #7069, opened 2026-07-03
Database login fallback for OIDC/JWT authentication

bartv:
_Claude, opening this on behalf of @bartv._

## What
When the identity provider is unavailable (or misconfigured), the web-console now lets an admin sign in with a database **break-glass** account, and manage users/roles as with database-only auth.

## Behaviour (identity provider first)
- On a normal load the OIDC provider auto-redirects to the IdP exactly as before.
- The local-login option is a **backup**: the chooser is shown only when the IdP errors (`auth.error` set), plus a manual break-glass via the `/console/login` URL.
- The local login page no longer bounces to the IdP: a failed local login (wrong password -> 401) triggers `authHelper.login()`, so `login()` now returns to `/login` (instead of `signinRedirect()`) when a local session exists or the user is on the login page.

## How
- Shared local-token store (`localToken.ts`, extracted from `DatabaseAuthProvider`) used by the OIDC and JWT providers in `getToken`/`updateUser`.
- `isDatabaseSession()` added to `AuthContextInterface`; the three checks that were hard-coded to `method === "database"` (`EnvSelector`, `UserManagement/Page`, `UserInfoRow`) now key off it, so user management works on a fallback session.
- `localFallback` field on the config types; three new translation strings.

Depends on inmanta-ui emitting `localFallback` in `config.js` and on inmanta-core allowing a break-glass DB user under `auth_method=oidc`.

## Tests
- New `OidcAuthProvider.test.tsx` (6 cases): IdP-first redirect, chooser only on error, plain error when fallback disabled, authenticated/local-session render the app.
- Affected suites (UserManagement, EnvSelector, Login) pass; eslint + tsc clean.

## Related PRs / MRs
- inmanta-core: https://github.com/inmanta/inmanta-core/pull/10531 (stacked on #10529 `db-singleton-lock`)
- inmanta-ui: https://github.com/inmanta/inmanta-ui/pull/822
- web-console: https://github.com/inmanta/web-console/pull/7069
- local-setup (test harness): https://code.inmanta.com/inmanta/local-setup/-/merge_requests/48

### DIFF
--- a/changelogs/unreleased/oidc-database-login-fallback.yml
+++ b/changelogs/unreleased/oidc-database-login-fallback.yml
@@ -1,0 +1,5 @@
+description: Added a database login fallback for OIDC and JWT authentication. When enabled and the identity provider is unavailable, an admin can sign in with a local database account and manage users and roles as with database authentication.
+change-type: minor
+destination-branches: [master, iso9]
+sections:
+  minor-improvement: "{{description}}"
--- a/src/Data/Auth/AuthContext.ts
+++ b/src/Data/Auth/AuthContext.ts
@@ -38,6 +38,15 @@ export interface AuthContextInterface {
    * @returns the boolean.
    */
   isDisabled: () => boolean;
+
+  /**
+   * Whether the active session authenticates against the built-in database, either
+   * because database auth is configured or because the user logged in via the local
+   * login fallback while an external provider (OIDC/JWT) is configured. When true,
+   * database user and role management is available.
+   * @returns the boolean.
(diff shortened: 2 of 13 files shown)


LukasStordeur (approved):
This will be the only place we are using the blocked buttons in the app, I personally don't mind, it's just an FYI.

bartv:
> This will be the only place we are using the blocked buttons in the app, I personally don't mind, it's just an FYI.

I have no idea what that means :-D

PR #7070, opened 2026-07-04
List and revoke API tokens in the Settings Tokens tab

bartv:
Extend the Settings -> Tokens tab so registered API tokens can be viewed and revoked.

Backend counterpart: inmanta/inmanta-core#10535.

## Change

- **Generate a revocable token:** a "Create a revocable token" checkbox on the generate form. When checked, the token is minted with `idempotent: false`, so the server registers it and it can later be revoked. Unchecked keeps the previous behavior (a stable, non-revocable token).
- **Registry table:** below the form, a table lists the environment's registered tokens (creator, client types, issued time, last used, active/revoked status) with a per-row **Revoke** action behind a confirm modal. Already-revoked tokens show as such and cannot be revoked again.
- New query hooks `useGetTokens` / `useRevokeToken` against `environment_token_list` / `environment_token_revoke`; generating a revocable token refreshes the list.

## Tests

`TokenTable.test.tsx` (list rendering, empty and error views, revoke via the confirm modal, cancel) and extended `Tab.test.tsx` (the revocable checkbox controls the `idempotent` flag). All green; tsc / eslint (zero-warnings) / prettier clean.

## Changelog

`changelogs/unreleased/token-registry.yml` (minor).

### DIFF
--- a/src/Slices/Settings/UI/Tabs/Token/Tab.tsx
+++ b/src/Slices/Settings/UI/Tabs/Token/Tab.tsx
@@ -1,7 +1,6 @@
 import React, { useState } from "react";
 import { Title } from "@patternfly/react-core";
 import { useQueryClient } from "@tanstack/react-query";
-import styled from "styled-components";
 import { ClientType, toggleValueInList } from "@/Core";
 import { getTokensKey, useGenerateToken } from "@/Data/Queries";
 import { words } from "@/UI/words";
@@ -69,15 +68,10 @@ export const Tab: React.FC = () => {
         error={error}
         isBusy={isBusy}
       />
-      <RegisteredTitle headingLevel="h2">
+      <Title className="lined_section" headingLevel="h2" size="md">
         {words("settings.tabs.token.registered.title")}
-      </RegisteredTitle>
+      </Title>
       <TokenTable />
     </>
   );
 };
-
-const RegisteredTitle = styled(Title)`
-  padding-top: 1rem;
-  padding-bottom: 1rem;
-`;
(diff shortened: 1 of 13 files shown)


src/Slices/Settings/UI/Tabs/Token/Tab.tsx
```diff
@@ -39,20 +51,33 @@ export const Tab: React.FC = () => {
...
+        onErrorClose={() => setError(null)}
+        getClientTypeSelector={getClientTypeSelector}
+        isClientTypeSelected={isClientTypeSelected}
+        isRevocable={isRevocable}
+        onRevocableChange={setIsRevocable}
+        token={token}
+        error={error}
+        isBusy={isBusy}
+      />
+      <RegisteredTitle headingLevel="h2">
```
LukasStordeur: I think we should reuse the same class/styling for this as on the settings tab to keep things more consistent; 
```ts
<Title className="lined_section" headingLevel="h2" size="md"></Title>
```

this class is globally available:
```css
  .lined_section {
    border-bottom: 1px solid var(--pf-t--global--border--color--default) !important;
    padding-bottom: 0.5rem;
    margin-bottom: 0.5rem;
  }
```

LukasStordeur (changes requested):

src/Slices/Settings/UI/Tabs/Token/Tab.tsx
bartv: Claude (on behalf of @bartv): Done in 75ec40a1 - replaced the one-off `styled(Title)` with `<Title className="lined_section" headingLevel="h2" size="md">`, matching the other settings tabs (e.g. Configuration/Container.tsx). Also rebased the branch onto current master.

LukasStordeur (approved):

PR #7074, opened 2026-07-06
Support ${...} variables in suggested-values parameter names

### LINKED ISSUE #7009: Variables in the parameter name
# T2 — Variables in the parameter name

> Part of #7007 . Depends on T1 (shared parameters path). The extract step here
> is the foundation T4 (cascading fields) builds its dependency graph on.

## Background
The `parameters` suggestion flavor fetches its list from a named orchestrator
parameter: `GET /api/v1/parameter/<parameter_name>`. Today `parameter_name` is a
fixed constant, so one annotation points at one global parameter for the whole
environment.

## Goal
Allow `${...}` placeholders in `parameter_name`, resolved from the current form's
identity, so a single annotation can address a per-entity or per-instance
parameter:

    "parameter_name": "topology_files_${entity_type}"   →   topology_files_Connection

Supported variables:
- `${entity_type}` — the form's service entity type (`service_entity` /
  `serviceEntity.name`).
- `${identifying_attribute}` — value of the attribute named by the service's
  `service_identity`, read from form state.
- `${instance_id}` — the edited instance's id (`instance.id`); empty on create.

## Approach: extract, then substitute
Handle `${...}` in two separate operations that run at different times — not one
plain string replace.

1. **Extract** — input is *just the template*; output is **data**: a structured
   list of which variables it references. Runs the moment the annotation loads,
   before any value exists. Answers "what does this field depend on?"

       extract("topology_files_${entity_type}")  →  [{ namespace: "entity_type" }]

2. **Substitute** — input is the template *plus the now-known values*; output is
   the final name string. Runs later, once those values exist.

       substitute("topology_files_${entity_type}", { entity_type: "Connection" })
         →  "topology_files_Connection"

Think of it as a function's signature (extract = what inputs it needs, known
early) vs. calling it (substitute = the result, known late). A plain replace only
does the second and throws away the dependency info.

That info is what makes the rest work:
- skip the fetch entirely when a required variable is absent (e.g.
  `${instance_id}` on a create form) instead of fetching a malformed name;
- flag an unknown variable as a model error at load time;
- **T4 reuses this**: running extract over all annotations yields the field-to-
  field dependency graph that drives blocking and re-querying. Build extract as a
  small standalone helper so T3 and T4 reuse the same engine.

## Scope
- [ ] Add a standalone `${...}` extract + substitute helper for the namespaces
      `entity_type`, `identifying_attribute`, `instance_id` (room for `form.*` /
      `self.*` later). Unknown namespace = surfaced model error. URL-encode
      substituted values.
- [ ] Thread form context (`entity_type`, identifying-attribute value,
      `instance_id`) down `ServiceInstanceForm.tsx → FieldInput.tsx →
      useSuggestedValues`. Source `instance_id` from the edit/create call sites
      (`EditForm.tsx`, `CreateInstance.tsx`); derive the identifying attribute
      from `service_identity` (`ServiceModel.ts`).
- [ ] In `useSuggestions.ts`, substitute the name *before* building the fetch URL,
      and use the **resolved** name in the React Query cache key
      (`getParametersKey.single`) so distinct contexts cache separately.
- [ ] On create, a name depending on `${instance_id}` resolves to "no value" and
      yields no suggestions (don't fire a malformed fetch).

## Acceptance criteria
- [ ] A templated `parameter_name` resolves per entity/instance and fetches the
      correct parameter.
- [ ] Two different contexts (e.g. two entity types) cache separately — no
      cross-contamination from the cache key.
- [ ] The create form shows no suggestions for an `${instance_id}`-dependent name.
- [ ] An unrecognized `${...}` namespace surfaces a clear error rather than
      fetching a broken name.
- [ ] Tests cover extract + substitute (incl. URL-encoding) and the create vs.
      edit context difference.

## Notes
- `${...}` content is never evaluated — parsed into a fixed namespace, used as a
  lookup key, values URL-encoded. Not a code-injection surface.
- No backend/API change: the compiler just publishes values under the composed
  name; the console fetches it through the existing `/api/v1/parameter/{name}`.

AronH99:
# Description

So what happened:
- Add a standalone ${...} template engine (extract / substitute / unknown-namespace detection) for the namespaces entity_type, identifying_attribute, instance_id, with URL-encoded substitution
- Made adjustments in the create/duplicate/edit forms so that these new param values can be supported
- Made adjustments to the composer's EntityForm so that these new param values can be supported
- Covered all necessary tests for everything

=> everything should be working exactly the same, in the forms and in the composer, it's kind of hard to show the full flow but i essentially intercepted and mocked some of the attributes locally which you can see the result of below, 1 working in combination with the "${entity_type}" and 1 with an error whenever there is a templated parameter which is not recognised.

Screenshot 2026-07-06 at 12 22 49

Screenshot 2026-07-06 at 12 23 21

closes https://github.com/inmanta/web-console/issues/7009


src/UI/Components/Composer/Data/Helpers/findOwningInstanceShape.ts
```diff
@@ -0,0 +1,36 @@
...
+ * it to the root instance, but links drawn on the canvas add the immediate
+ * parent and can accumulate several. So we climb one non-embedded step at a time,
+ * guarding against cycles, and return the first core/relation shape reached (or
+ * undefined when the chain can't be resolved, e.g. an unconnected new entity).
+ *
+ * @param shape - The starting shape (a non-embedded shape is returned as-is).
+ * @param cells - The canvas shapes, keyed by id.
+ * @returns The owning instance shape, or undefined if none can be resolved.
+ */
+export const findOwningInstanceShape = <T extends ShapeNode>(
```
LukasStordeur: I'd be careful with the term "owning". 
There's an `owned_entities` field in the model, which we can't use because it's not reliable, or at least doesn't do what we'd expect from it, from what I understood. 

I think `findTopRootInstance` might be safer to use, and more explicit considering the wordings we are already using so far in the composer.

It's something I'd add in the shape class too, to only calculate it once.

src/UI/Components/Composer/UI/EntityForm.tsx
```diff
@@ -35,13 +37,45 @@ interface Props {
...
   const [originalState, setOriginalState] = useState<InstanceAttributeModel>({});
   const [isDirty, setIsDirty] = useState(false);
   const [currentCellId, setCurrentCellId] = useState<string | null>(null);
 
+  // Identity used to resolve `${...}` variables in a suggestion's parameter name.
+  // An embedded entity's suggestions resolve against its owning instance -
+  // entity_type, identity, and id all belong to the service instance, not the
+  // embedded part - so we climb to that instance's shape; a top-level shape uses
+  // itself.
+  const suggestionContext: TemplateContext = useMemo(() => {
```
LukasStordeur: Aren't these things we'd better keep in the `EntityShape`? It feels redundant to recalculate this whenever the user clicks through different shapes.

src/UI/Components/Composer/UI/EntityForm.tsx
```diff
@@ -205,6 +239,7 @@ export const EntityForm: React.FC<Props> = ({ activeCell, isDisabled }) => {
                 getUpdate={getUpdate}
                 path={null}
                 suggestions={field.suggestion}
+                suggestionContext={suggestionContext}
```
LukasStordeur: If we use a `suggestionContext`, I'd expect it to contain `suggestions` and not need to pass it as an additional attribute

src/UI/Components/ServiceInstanceForm/Components/FieldInput.tsx
```diff
@@ -357,6 +368,7 @@ interface NestedProps {
   getUpdate: GetUpdate;
   path: string | null;
   isNew?: boolean;
+  suggestionContext?: TemplateContext;
```
LukasStordeur: if the attribute is always going to be `suggestionContext`, I'd rename the Type to `SuggestionContext` instead of `TemplateContext`.
In the composer, `Template` could potentially mean different things, such as a blueprint concept.

src/UI/Components/Composer/Data/Helpers/findOwningInstanceShape.ts
```diff
@@ -0,0 +1,36 @@
+interface ShapeNode {
```
LukasStordeur: You do have this information in the `ServiceEntityBase` interface. I wouldn't recreate an interface for that. You can do a `Pick` to get what you need.

```ts
export interface ServiceEntityBase {
  entityType: EntityShapeType;
  serviceModel: ServiceModel | EmbeddedEntity;
  instanceAttributes: InstanceAttributeModel;
  readonly: boolean;
  isNew: boolean;
  rootEntities: Record<string, string[]>; // => This is the list of parentIds
  interServiceRelations: Record<string, string[]>;
  embeddedEntities: Record<string, string[]>;
  id: string;
}
```

LukasStordeur (dismissed):
Makes me wonder now that I see the whole implementation, why the backend doesn't resolve the variables on their side. For this specific case. 
I do understand it should be resolved on our side for the other variations, but this one? We have it now, but for the future, it might be something we'd like to move to the backend since the variables don't rely on frontend values. It would also shift the model validation on the backend side and make them able to be more specific in the logs.

What's the difference between service id, and service identity? 

You're mixing terms Template and Suggestion, I'd stick to Suggestions in general.

I'd also only keep `suggestionContext`, instead of having both context, and suggestions.

AronH99:
@LukasStordeur 

That's actually a valid concern which I didn't think about for this implementation, but we for sure do need this kind of implementation for the identifying_attribute too.
So whilst it's true the backend can technically handle the entity_type and instance_id directly themselves for both edit and create, the identifying attribute is something which can sometimes be edited or whenever adding a new instance initialized and it immediately fetches the suggestions whenever filled.
For the instance_id whenever adding a new instance the uuid is undefined too.
All in all I am not 100% sure what other things are exactly possible as well but I think this keeps it more flexible than purely on the backend and since we need the extraction/substitution part either way this is quite all right to do I reckon.
But for sure a concern which can be addressed later on in the FE meeting if you want.

To clarify some examples I used:
```ts
//   "topology_files_${entity_type}"          -> per-entity, always resolvable => Example: basic-instance (type of instance)
//   "topology_files_${identifying_attribute}" -> resolves once you type the id attr => Example: fe-prev (name value can be the identifying attribute, but is flexible and defined on the backend)
//   "topology_files_${instance_id}"  -> no fetch on create, fetches on edit=> Example: 61a81011-879e-4058-9d9c-34d998c473e8 (just a uuid from the backend)
```

src/UI/Components/Composer/UI/EntityForm.tsx
AronH99: This is somewhat unavoidable since the identifying_attribute can only be calculated live

src/UI/Components/Composer/UI/EntityForm.tsx
AronH99: SuggestionContext is form wide whilst suggestions is field specific so i'd prefer to keep them separate as they serve different purposes, i can place a comment to clarify this?

src/UI/Components/Composer/UI/EntityForm.tsx
LukasStordeur: What makes it impossible to let the shape handle that? It contains all the instance form data too, I don't see the problem.

And I'd keep the `identifying_attribute` name over `identity`, if they're the same.

src/UI/Components/Composer/Data/Helpers/findOwningInstanceShape.ts
AronH99: This can be added for sure, I kind of did now (locally atm) but it still can't be cached or it might get stale

src/UI/Components/Composer/Data/Helpers/findOwningInstanceShape.ts
AronH99: I did use a pick now of the entityShape

src/UI/Components/Composer/UI/EntityForm.tsx
LukasStordeur: It just feels weird to have a suggestionContext that doesn't contain the suggestions. 
If it's truly a context, it might be worth having it as a real react context. That would prevent the prop-drilling part

src/Data/Queries/Slices/ServiceInstance/FormSuggestions/index.ts
```diff
@@ -1 +1,3 @@
 export { useSuggestedValues } from "./useSuggestions";
+export { TEMPLATE_NAMESPACES } from "./templateVariables";
+export type { SuggestionContext } from "./templateVariables";
```
LukasStordeur: Also rename the folder/file names, type to be in sync

src/UI/Components/Composer/Data/Helpers/findTopRootInstance.test.ts
```diff
@@ -0,0 +1,66 @@
+import { findTopRootInstance } from "./findTopRootInstance";
+
+interface TestShape {
```
LukasStordeur: Pick BaseEntityShape?

src/UI/Components/Composer/Data/Helpers/findTopRootInstance.ts
```diff
@@ -0,0 +1,40 @@
+import type { ServiceEntityShape } from "../../UI/JointJsShapes/ServiceEntityShape";
+
+/**
+ * The subset of a shape this walk needs: its id, its type, and the ids of the
+ * shapes it hangs off. Typed as a `Pick` of the real shape so we reuse that
+ * contract instead of inventing a parallel interface, while staying generic
+ * enough to unit-test with plain objects.
+ */
+type RootWalkable = Pick<ServiceEntityShape, "id" | "entityType" | "parentIds">;
```
LukasStordeur: PartialEntityShape?

src/UI/Components/Composer/UI/JointJsShapes/ServiceEntityShape.ts
```diff
@@ -392,6 +393,23 @@ export class ServiceEntityShape extends shapes.standard.HeaderedRecord {
     return this.serviceModel.name;
   }
 
+  /**
+   * Resolves the top root instance shape whose identity a suggestion's `${...}`
```
LukasStordeur: I think you can omit the suggestion part in the comment, after all, this is just a plain method that will fetch the topRoot node, nothing more. 

You can also drop the Not cached part, I think that's evident.

LukasStordeur (dismissed):
You're still mixing identity/serviceIdentity, and indentifying_attribute, I'd keep the naming consistent to prevent confusion.

LukasStordeur:
also an fyi, not sure if you took that in account in your logic, but the composer is creating uuids for instance_id when adding a root shape on the canvas.

AronH99:
> also an fyi, not sure if you took that in account in your logic, but the composer is creating uuids for instance_id when adding a root shape on the canvas.

Should have taken it into account, it resolves as undefined whenever adding a new root shape onto the canvas, which makes sense to do so

src/UI/Components/Composer/UI/EntityForm.tsx
AronH99: renamed it suggestionVariables, to have less confusion

AronH99:
@bartv So I should have addressed all of Lukas's concerns mostly about naming/confusion and some other minor things.
If you can have a last check then this should be good to merge.
I think 1 approve is enough?

bartv (approved):

PR #7099, opened 2026-07-10
Add annotation-driven form tabs to service instance form

### LINKED ISSUE #7012: Form tabs
# T5 — Form tabs

> Part of #7007 . Independent of the suggestion work (T1–T4) — can be done in
> parallel. Pure layout, annotation-driven.

## Background
The service form renders as a single scrolling column (`ServiceInstanceForm.tsx`
maps over fields top to bottom). Large services are hard to navigate.

## Goal
Let the model split the form's **top-level** fields into named PatternFly tabs via
annotations — no per-service frontend code. Two separate concerns:

- **Tab catalog** — on the service entity (`__annotations` → `web_tabs`): the set
  of tabs, each `{ key, label, order, default, icon? }`.
- **Tab assignment** — per field (`web_tab` = a tab key): on an attribute's
  `__annotations`, or on a relation via `lsm::RelationAnnotations` (an assigned
  embedded relation moves its whole sub-form onto the tab as one unit).

## Approach
- [ ] Parse + type a `web_tabs` catalog from the service entity's
      `entity_annotations` (currently `Record<string, unknown>` in
      `ServiceModel.ts`).
- [ ] Read `web_tab` from attribute annotations (add to `AttributeAnnotations`) and
      from relation annotations; surface a resolved tab key per top-level field
      (likely via `FieldCreator`).
- [ ] Render top-level fields/relations into PatternFly `Tabs` in
      `ServiceInstanceForm.tsx` (reuse the pattern from `TabView.tsx`). Group by
      resolved tab; default tab catches unassigned fields and shows first; order by
      `order`, ties by `key`.

## Acceptance criteria
- [ ] No `web_tabs` on the entity → form renders single-column, exactly as today.
- [ ] Tabs display in ascending `order`; ties (or missing `order`) fall back to
      `key` — deterministic, never list-position.
- [ ] Fields with no `web_tab` land on the `default` tab; the default shows first.
- [ ] Exactly one `default` is required; zero or more than one is a surfaced model
      error.
- [ ] A `web_tab` naming a key not in `web_tabs` is a surfaced model error.
- [ ] `web_tabs`/`web_tab` deeper than top level (on embedded entities or their
      fields) is ignored; an assigned embedded relation moves as a whole unit.
- [ ] Tab assignment is presentation-only — no change to attribute order in the API
      or to LSM behavior.

## Notes
- Mirrors the existing documentation-tab annotation split (`web_title`/`web_icon`
  on the entity, assignment per field).
- Explicit `order` + `default` (not positional) so correctness doesn't depend on
  any stage (compiler → API → frontend) preserving list order.
- Contract check: verify the API delivers `web_tab` on relation annotations for
  embedded relations (`InterServiceRelation.attribute_annotations` is untyped
  `Record<string, unknown>`). File a linked backend issue if it doesn't.

AronH99:
# Description

What happened in this pr:

- The web_tabs on entityAnnotations and the web_tab on attribute annotations are now supported.
- Covered new changes with unit tests
- Surfaces model errors properly (1 screenshot of 1 possible error => if any errors we show normal view)

=> requirements of ticket should be fulfilled.

Note: We should probably add e2e tests whenever the backend can add a service which includes this functionality.

closes https://github.com/inmanta/web-console/issues/7012

Screenshot 2026-07-10 at 09 19 10
Screenshot 2026-07-10 at 09 19 15
Screenshot 2026-07-10 at 09 19 21
Screenshot 2026-07-10 at 10 11 27


changelogs/unreleased/7012-form-tabs.yml
```diff
@@ -0,0 +1,6 @@
+description: Service instance forms can now be split into named tabs through annotations. A web_tabs catalog on the service entity defines the tabs (key, label, order, default, optional icon), and a web_tab annotation on a top-level attribute or relation assigns it to a tab; unassigned fields land on the default tab.
+issue-nr: 7012
+change-type: minor
```
AronH99, line 3: this a minor right?

.gitignore
```diff
@@ -77,6 +77,9 @@ temp
 # version file that is autogenerated on each build.
 src/version.json
 
+# Local dev-only request mocks (kept on-machine, never committed)
+src/Dev/
```
AronH99, line 81: 🤖 Claude · [Should-fix]: This `src/Dev/` ignore for local request mocks looks unrelated to the form-tabs feature. This reviewer is firm on scope hygiene — unrelated changes inflate review load and mix concerns. I'd pull it into its own tiny PR (or drop it), unless there's a reason it belongs here that I'm missing :)

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
```diff
@@ -251,19 +286,46 @@ export const ServiceInstanceForm: React.FC<Props> = ({
           data={JSON.stringify(formState, null, 2)}
           onChange={onEditorChange}
         />
+      ) : formTabs.kind === "tabs" ? (
+        <Tabs
```
AronH99, line 290: 🤖 Claude · [Question]: Before tabs, every field was in one column so a submit/validation error was always visible. With tabs, a field that fails backend validation can sit on an inactive tab with no indication of which tab holds the error. That's a new gap this feature introduces. PF supports a status/error indicator on the tab title — worth deciding whether we surface that now or accept it for the MVP. Flagging as a product call for the human reviewer rather than blocking on it.

AronH99 (commented):
## 🤖 Claude pre-review — PR #7099 — Add annotation-driven form tabs to service instance form

**Context map**

- **Type & intent:** feat — annotation-driven tabs for the service instance form (closes #7012, T5 of the annotation epic). A `web_tabs` catalog on the entity defines tabs; a `web_tab` on a top-level attribute/relation assigns it; unassigned fields land on the default tab.
- **Surface area:** shared `src/UI/Components/ServiceInstanceForm` (FieldCreator refactor + new `FormTabs` resolver + render branch), `src/Core/Domain` types (`Field.tab`, `EntityAnnotations`, `RelationAnnotations`, `FormTabDefinition`), three slice call-sites (Create/Duplicate/Edit) threading `entityAnnotations`, `words`, and test fixtures.
- **Size & balance:** +654 / −105 across 17 files. Test-to-source ratio is healthy — new unit specs for `FormTabs`, `FieldCreator`, and the form component, plus a shared fixture. 👍
- **Blast radius:** touches a **shared** form component used by Create/Duplicate/Edit, so it deserves strict scrutiny — but the change is cleanly additive (untabbed path is preserved, tab logic is opt-in on `web_tabs`).
- **Risk hotspots:** runtime validation of model-author input (well handled), a11y/UX of the new PF Tabs layout, and scope hygiene.

No correctness blockers — the resolver is carefully validated, the untabbed path is untouched, docstrings and centralized `words` are in place, and the tests are thorough. Two items I'd want addressed or consciously accepted before merge (an unrelated `.gitignore` change, and the validation-on-a-hidden-tab UX question), plus three nits. Resolving those two would make this a clean approve, imo.

1 should-fix, 1 question, 3 nits inline below.

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
```diff
@@ -251,19 +286,46 @@ export const ServiceInstanceForm: React.FC<Props> = ({
...
+                  {tab.icon && (
+                    <TabTitleIcon>
+                      <DynamicFAIcon icon={tab.icon} />
+                    </TabTitleIcon>
+                  )}
+                  <TabTitleText>{tab.label}</TabTitleText>
+                </>
+              }
+            >
+              <Flex direction={{ default: "column" }} gap={{ default: "gapMd" }}>
```
AronH99: 🤖 Claude · [Nit]: The tabbed path wraps fields in `<Flex direction=column gap=gapMd>`, but the untabbed path renders fields directly (relying on the form's own spacing). Worth a quick visual check that field spacing inside a tab matches the single-column form so the two layouts don't drift — a visual/taste call for the human reviewer.

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
```diff
@@ -114,6 +128,14 @@ export const ServiceInstanceForm: React.FC<Props> = ({
   const [isEditorValid, setIsEditorValid] = useState(true);
   const [isSubmitDropdownOpen, setIsSubmitDropdownOpen] = useState(false);
 
+  // An "error" resolution surfaces the model error and falls back to the
+  // single-column layout so the form stays usable.
+  const formTabs = useMemo(
+    () => resolveFormTabs(entityAnnotations, fields),
+    [entityAnnotations, fields]
+  );
+  const [activeTab, setActiveTab] = useState(formTabs.kind === "tabs" ? formTabs.defaultKey : "");
```
AronH99, line 137: 🤖 Claude · [Nit]: `activeTab` is seeded once from the initial resolution. If `entityAnnotations`/`fields` later change the resolution (e.g. untabbed→tabs, or the active key stops existing), `activeTab` stays stale (`""`). Low-risk since the entity is stable per form, but if you want to be safe you could derive/resync it from `formTabs` in an effect. Fine to leave if we're confident the props don't change kind mid-life.

src/UI/Components/ServiceInstanceForm/Helpers/FormTabs.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  if (defaults.length !== 1) {
+    return {
+      kind: "error",
+      message: words("inventory.form.tabs.defaultRequired")(defaults.length),
+    };
+  }
+
+  // Every field assignment must point at a key that exists in the catalog.
+  const knownKeys = new Set(catalog.map((tab) => tab.key));
```
AronH99: 🤖 Claude · [Nit]: `new Set(catalog.map((tab) => tab.key))` is computed here and again on line 56 for the duplicate-key check. Minor, but you could compute the key Set once above and reuse it for both the uniqueness check and `knownKeys`.

.gitignore
AronH99: Yes this is something I want to have for mocking data in the future locally, so i want to keep this

src/UI/Components/ServiceInstanceForm/ServiceInstanceForm.tsx
AronH99: 🤖 Claude: Thanks for the context — retracting this. Confirmed submit/API errors surface via a global `notifyError` notification (e.g. `CreateInstance` `onError`), which is always visible regardless of the active tab, and there are no inline per-field validation errors except the suggestion-fetch error on the field itself. So the only thing that could sit on a hidden tab is a suggestion-fetch error, which is minor. No action needed here.

AronH99:
I should have addressed all relevant claude feedback now

changelogs/unreleased/7012-form-tabs.yml
bartv: yes

bartv (approved):

PR #7110, opened 2026-07-13
Token expiry on create + simplified create-token form

### LINKED ISSUE #7104: Allow setting a token expiry on create and show it in the Tokens tab
Follow-up to the token registry UI (#7070), from the demo-session feedback.

Let the user set an expiry when creating a token, and show each token's expiry in the Tokens tab.

### Scope
- Add an expiry input to the create-token form (`Settings > Tokens`), passed to `environment_auth` (`useGenerateToken`).
- Add an "Expires at" column to the Tokens tab.

### Depends on
inmanta-core: add a per-token expiry parameter to `environment_create_token` - inmanta/inmanta-core#5643.

bartv:
# Description

Claude here, on behalf of @bartv.

Follow-up to the token registry UI (#7070), from the demo-session feedback:

* The create-token form gets an "Expires after" select with preset lifetimes (1 hour / 1 day / 7 days / 30 days / 1 year, default "Never") plus a "Custom" option that reveals an amount input and a minutes/hours/days unit select. The control is only enabled for revocable tokens, because idempotent tokens carry no time-based claims and the server rejects the combination. When set, the computed `expire` (seconds) is passed to `environment_auth`; an empty or invalid custom amount falls back to no expiry.
* The registered-tokens table gets an "Expires at" column.
* **Simplified create-token form** (from the [#development discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1783946953533589) between Arnaud and Bart): tokens are generated for the `api` client type by default, and the client-type toggles move into a collapsed "Advanced options" section with helper text explaining when `agent`/`compiler` are actually needed (externally hosted agent, remote compiler). The icons are fixed while at it: api = key, agent = robot (no longer a human), compiler = code.

Stacked on #7106 (both touch the TokenTable); base is set to that branch and I'll retarget to master once it merges. Depends on inmanta/inmanta-core#10578 for the server-side `expire` parameter. The request-body tests set the revocable checkbox state explicitly, so they are independent of the default flip in #7105.

closes #7104


LukasStordeur (approved):

src/Slices/Settings/UI/Tabs/Token/TokenForm.tsx
```diff
@@ -1,27 +1,199 @@
...
+interface ExpiryInputProps {
+  onChange(value: number | null): void;
+  isDisabled: boolean;
+}
+
+/**
+ * Expiry editor for the create-token form: preset lifetimes plus a custom amount + unit entry.
+ * Emits the effective lifetime in seconds through onChange, or null when no (valid) expiry is set.
+ */
+const ExpiryInput: React.FC<ExpiryInputProps> = ({ onChange, isDisabled }) => {
```
AronH99, line 48: To be honest, I don't really like that this is inside of the TokenForm.tsx file, an entire input/component which possible can be a standalone component that we might reuse?
If we would never reuse this then at least putting this into a separate file inside of the Token folder would be better/more clear in my opinion..

LukasStordeur: seconding this, to at least extract it in another file. We can always decide later on if it should be moved to the shared Component folder or not.

src/Slices/Settings/UI/Tabs/Token/TokenForm.tsx
```diff
@@ -42,33 +215,10 @@ export const TokenForm: React.FC<Props> = ({
...
-          <ToggleGroupItem
-            icon={<ProcessAutomationIcon />}
-            text="compiler"
-            aria-label="CompilerOption"
-            isSelected={isClientTypeSelected("compiler")}
-            onChange={(_event, selected) => getClientTypeSelector("compiler")(selected)}
-            isDisabled={isBusy}
-          />
-        </ToggleGroup>
+        <label htmlFor="token-expiry">{words("settings.tabs.token.expiry")}</label>
```
LukasStordeur, line 218: Looks like I missed a few things earlier. 
but this should be part of the Input element. And we should use patternfly elements;
You can add a label to a FormGroup https://www.patternfly.org/components/forms/form#formgroup

src/Slices/Settings/UI/Tabs/Token/TokenForm.tsx
```diff
@@ -115,10 +271,26 @@ export const TokenForm: React.FC<Props> = ({
 );
 
 const StyledInputGroup = styled(InputGroup)`
+  padding-top: 1rem;
   padding-bottom: 1rem;
   max-width: 600px;
 `;
 
+const AdvancedContent = styled.div`
```
LukasStordeur, line 279: I'd prefer using the Flex components provided by patternfly over using styled components for this.

src/Slices/Settings/UI/Tabs/Token/TokenForm.tsx
bartv: _Claude here, on behalf of @bartv._

Extracted it into its own file (ExpiryInput.tsx in the Token folder) in c54cc3af; happy to promote it to the shared Components folder later if a second consumer shows up.

src/Slices/Settings/UI/Tabs/Token/TokenForm.tsx
bartv: _Claude here, on behalf of @bartv._

Done in c54cc3af: ExpiryInput now renders its own FormGroup with the Expires after label (fieldId association instead of the bare label + aria-label), so the label travels with the input component.

src/Slices/Settings/UI/Tabs/Token/TokenForm.tsx
bartv: _Claude here, on behalf of @bartv._

Replaced with a PatternFly Flex (column direction, gapSm) in c54cc3af. The two width-constraining styled components (amount input, unit select) stayed, since PatternFly has no width prop for those controls - same pattern as the existing StyledInputGroup.

bartv:
_Claude here, on behalf of @bartv._

This PR was auto-closed when its base branch (#7106) merged and was deleted, and GitHub refuses to reopen it after the head branch was rebased onto master. Continued in #7119 (same branch, rebased, targeting master); all review comments from this PR have been addressed there in 17e85935.

Issue #7073, opened 2026-07-06
Order view usability improvements


edvgui:
We finally started getting real usage on the order api view in the web console, and got some valuable user experience feedback:
- acknowledged and success are the same green color, this makes the progress bar look confusing as it first get green, then goes back to blue when things are being deployed, then again green when everything is deployed.  And when some services are deployed while others are still acknowledged, we get two numbers in the green zone, one for ack and the other for success
- all the service ids represented in the different views are not clickable, ideally they should allow us to jump to the details page of the relevant service (for the dependencies list, jumping to the service row in the order view is maybe better)
- when a service has an identifying attribute, we should use this attribute instead of the uuid for each reference to it (the same way we already do it for inter-service relationships)

PR #7115, opened 2026-07-14
7073 order view improvements

LukasStordeur:
# Description

1. The color for acknowledged is now of a lighter blue than the in progress state. So now the progression would be light blue, darker blue, green, or red if failed.
2. Order detail rows now replace raw uuids with the instance display name/attribute. 
3. instance ids within the order details rows, so in the status.dependencies part, will expand the matching row when clicked on.
4. instance ids in the order detail rows will forward to the instance details page when there is one. Two failureTypes lead to skipping the instance creation so we keep the instance id as plain display text. 
5. added regression test for breadcrumb decoding/url encoding. 

closes:
- #7073  

https://github.com/user-attachments/assets/299d3041-68e5-4bc2-8b3c-44b287055e13


src/UI/Routing/PrimaryRouteManager.ts
```diff
@@ -161,7 +160,10 @@ export function PrimaryRouteManager(baseUrl: string): RouteManager {
   function getUrl<K extends RouteKind>(kind: K, params: RouteParams<K>): string {
     const route = getRoute(kind);
 
-    return generatePath(route.path, params === undefined ? params : encodeParams(params));
+    // generatePath already encodeURIComponent's each param; encoding them here too would
+    // double-encode values with special characters (e.g. "Parent 1" -> "Parent%2520" instead
+    // of "Parent%20"), which then only gets decoded once and leaks a raw "Parent%20" into the UI.
+    return generatePath(route.path, params);
```
AronH99, line 166: 🤖 Claude · [Blocking]: Dropping `encodeParams` here is correct (RR7's `generatePath` already `encodeURIComponent`s each param — confirmed empirically), but the **read path was not updated to match**. `useRouteParams()` in `Utils.ts` calls `useParams()` **and then** `decodeParams()`. RR7's `decodePath()` already decodes matched params once, so `decodeParams` decodes a **second** time. So this moves the app from encode×2/decode×2 (balanced) to encode×1/decode×2 (asymmetric).

For a value like `"Parent 1"` the second decode is a harmless no-op, which is why it looks fine. But for any param containing a literal `%` (or a `%XX`-looking sequence) — e.g. a resource id or a service-identity attribute value — the second `decodeURIComponent` in `decodeParams` mangles it or throws `URIError`, and `decodeParams` has **no try/catch** (unlike RR's `decodePath`), so the page crashes.

Fix: apply the same reasoning to the read side — since RR already decodes, drop the extra `decodeParams` in `useRouteParams` so it's encode×1/decode×1 end-to-end.

src/UI/Routing/PrimaryRouteManager.test.ts
```diff
@@ -132,3 +132,20 @@ test("GIVEN '/resources/123' THEN breadcrumbs should be ['Dashboard', 'Resources
     },
   ]);
 });
+
+test("GIVEN a route param containing a space WHEN building a url with getUrl and reading it back with getCrumbs THEN the value is not double encoded", () => {
+  // Regression test: getUrl used to encodeURIComponent params before handing them to
+  // react-router's generatePath, which encodes them again internally. A value like
+  // "Parent 1" would survive the single decodeURIComponent done by consumers (e.g.
+  // PageBreadcrumbs) as "Parent%201" instead of "Parent 1".
+  const url = routeManager.getUrl("InstanceDetails", {
```
AronH99, line 141: 🤖 Claude · [Should-fix]: This regression test exercises `getUrl → getCrumbs`, which reads params via `getRouteMatchFromUrl`/`matchPath` (decode×1) — that path *is* balanced, so it passes and gives false confidence. It does **not** cover the `useRouteParams` path that actual pages use to read their params (the asymmetric one flagged on `PrimaryRouteManager.ts`). Add a test that round-trips through `useRouteParams` with a special-char value (e.g. one containing `%`).

src/UI/Routing/PrimaryRouteManager.ts
AronH99, line 166: 🤖 Claude · [Should-fix]: Scope/blast-radius — this is an app-wide URL-generation change riding inside an order-view usability PR. It's motivated here (you now navigate using display names like `"Parent 1"` that contain spaces), but it affects every route. Given the read-path asymmetry above, I'd want e2e evidence on the scary real inputs before merge: **resource IDs** (`std::File[agent,path=/etc/motd]` — heavy `::`, `[]`, `/`, `,`) and service instances whose identity attribute has special chars. Either split it into its own focused PR, or keep it here but add that e2e coverage. imo one unit test isn't enough for a change this wide.

src/Slices/OrderDetails/UI/getInstanceDisplayName.ts
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Returns the service identity display name or attribute value for an order item,
+ * or null if the item is unknown or has neither set.
+ */
+export function getInstanceDisplayName(row: ServiceOrderItem | undefined): string | null {
+  if (!row) {
+    return null;
+  }
+
+  return row.status.service_identity_display_name || row.status.service_identity_attribute_value;
```
AronH99, line 12: 🤖 Claude · [Should-fix]: Repo convention is logic in a `helpers.ts`, not a one-function file named after the function. Suggest folding this into an `OrderDetails/UI/helpers.ts`. Minor, but the reviewer is consistent about this split. :)

src/UI/Components/ClipboardCopyButton/ClipboardCopyButton.tsx
```diff
@@ -23,7 +23,9 @@ export const ClipboardCopyButton: React.FC<Props> = ({
   ...props
 }) => {
   const [copied, setCopied] = useState(false);
-  const onClick = () => {
+  const onClick = (event: React.MouseEvent) => {
+    // Prevent triggering click handlers on ancestor elements, e.g. a clickable table row.
+    event.stopPropagation();
```
AronH99, line 28: 🤖 Claude · [Should-fix]: This `stopPropagation` (and the ones in `CopyMultiOptions`) now applies to **every** consumer of these shared components, not just the clickable order rows this PR adds. It's benign (a copy click shouldn't bubble anyway) and you added tests, so I lean fine — just flagging it's a global behavioral change made for a local need. The alternative would be stopping propagation at the row level; the nested-button approach is pragmatic, your call.

src/Slices/OrderDetails/UI/OrderInstanceLink.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const instanceDisplayName = getInstanceDisplayName(row);
+  const displayValue = instanceDisplayName || row.instance_id;
+
+  const instanceDetailsUrl = routeManager.useUrl("InstanceDetails", {
+    service: row.service_entity,
+    instance: instanceDisplayName || row.instance_id,
+    instanceId: row.instance_id,
+  });
+
+  const instanceWasNeverCreated =
```
AronH99, line 41: 🤖 Claude · [Question]: The "never created" guard only covers `action === "create"`. A **completed `delete`** item links to an instance that no longer exists, so clicking lands on a not-found page. Intentional, or should a completed delete get the same plain-text (non-link) treatment?

src/Slices/Orders/UI/OrderProgressBar.tsx
```diff
@@ -64,7 +65,7 @@ const fromProgressToItems = (items: ServiceOrderItem[]): LegendItemDetails[] =>
       id: "acknowledged",
       label: words("orders.status.acknowledged"),
       value: Number(items.filter((item) => item.status.state === "acknowledged").length),
-      backgroundColor: t_global_icon_color_status_success_default.var,
+      backgroundColor: t_chart_color_blue_100.var,
```
AronH99, line 68: 🤖 Claude · [Nit]: `acknowledged` is now `t_chart_color_blue_100` and `in_progress` is `t_global_color_brand_default` — two blues. This fixes the original green/green complaint, but worth confirming the two blues are visually distinct in the bar so it doesn't recreate the confusion in a different color. Visual/product call for the human reviewer.

changelogs/unreleased/7073-orderView-improvements.yml
```diff
@@ -0,0 +1,6 @@
+description: Add a different color for the acknowledged status in the order progress bar and add a link to the service instance in the order details row.
```
AronH99: 🤖 Claude · [Nit]: The description covers the color change and the instance link, but omits the dependency-row navigation and the identity-attribute display (issue requirement 3). Worth expanding so the release notes match everything that shipped.

AronH99 (changes requested):
## 🤖 Claude pre-review — PR #7115 — 7073 order view improvements

_This is an automated pre-review by Claude (not a human reviewer's verdict). It borrows the team's review standards; the team decides what to do with it._

**Context map**
- **Type/intent:** feat + fix — order-view usability (progress-bar color, clickable instance IDs, identity-attribute display).
- **Surface area:** mostly `src/Slices/OrderDetails` + `src/Slices/Orders` (contained). It also touches three shared layers: `src/Data/Common/useExpansion.ts`, `src/UI/Components/{ClipboardCopyButton,CopyMultiOptions}`, and — the big one — `src/UI/Routing/{PrimaryRouteManager,Utils}.ts`.
- **Blast radius:** the routing change rewrites **every** `getUrl`/`useUrl` call in the app. That's the hotspot.
- **Tests:** 3 new test files + a routing regression test. Good ratio on the feature code.

**Verdict: Request changes** — 1 blocking (an encode/decode asymmetry in the routing change) + 2 should-fix + 3 nits/questions inline. The order-view feature work itself is clean, well-tested and PF-native, and would approve on its own; the concern is the shared routing edit bundled in.

What would flip it to approve: make the routing read-path symmetric (see the blocking comment) and back the app-wide URL change with e2e on real special-char inputs (resource IDs, identity attribute values), or split it into its own PR.

AronH99 (commented):
Think some of the comments might be warranted from claude... some probably aren't but I read through them and definitely some of them are worth a look.
Also (see video), I would expect whenever I click on the trst (in this case) the trst (parent-service) would open up, which it does, but then the user's scroll would be snapped to the just opened up service, no?

https://github.com/user-attachments/assets/643b89e0-3c98-4b38-ae98-f7b3d07db2af

src/UI/Routing/PrimaryRouteManager.ts
AronH99: This is something I was thinking as well that this change was app wide and I was looking if it wouldn't be an issue, because I remember working on this too somewhere else.
And it seems like it did found an issue (above)..

src/UI/Components/ClipboardCopyButton/ClipboardCopyButton.tsx
LukasStordeur: this is intended.

src/Slices/OrderDetails/UI/OrderInstanceLink.tsx
LukasStordeur: Deleted instances still have details pages.

src/Slices/Orders/UI/OrderProgressBar.tsx
LukasStordeur: Yes, mentionned in the pr description to have a logical progression in colors.

src/Slices/OrderDetails/UI/getInstanceDisplayName.ts
LukasStordeur: I think both are fine honestly, I find it personally easier to navigate when the file names match when they only contain one method.

LukasStordeur:
Adressed the comments. 
I added the scroll into view too.
The various sorts of ids we have in the app are now also tested in bothe directions. 


https://github.com/user-attachments/assets/b7af9076-750c-4048-8c3c-3d28cebeef5e

AronH99 (approved):

PR #7117, opened 2026-07-15
security bump dependencies

LukasStordeur:
# Description

This should address all open security alerts we have at the moment. 

Package | Before | After | Notes
-- | -- | -- | --
vite | 8.0.9 (pulled in via vitest's peer range, separate from the direct ^8.1.3 pin) | 8.1.4 | Also collapsed two duplicate vite installs into one
markdown-it | 14.1.1 | 14.3.0 | Merged with the existing ^14.3.0 resolution
@babel/core | 7.29.0 | 7.29.7 | Resolved to latest available patch
undici | 8.1.0 | 8.7.0 | Resolved to latest available patch


AronH99 (approved):
Got some conflicts but approved

PR #7120, opened 2026-07-15
Move filter building blocks to shared UI components and reuse across drawers

### LINKED ISSUE #7027: Components that can be shared should be moved to the shared UI components folder and reused in all of the filter drawers where needed.
1) Move AddableTextInput to shared components and scan the entire web-console where it can be reused.
2) Move AddableSelectInput to shared components and scan the entire web-console where it can be reused.
3) Move ActiveFilterGroup to shared components and scan the entire web-console where it can be reused.
4) Create TimeStampRangeFilter for shared components and reuse it.
5) Create ActiveFilters for shared components and reuse it.

=> These will be mainly used for the all of the new filter drawers (inside of ConnectedFilterWidget throughout web)

AronH99:
# Description

- Moved AddableTextInput and AddableSelectInput into their own top-level src/UI/Components/ dirs; kept ActiveFilterGroup in FilterDrawer/.
- Extracted a shared ActiveFilters section shell and adopted it in every drawer (Resources, Parameters, Agents, Events, DesiredState) plus the two former generic-widget users
- Removed the config-driven FilterWidget/FilterWidgetComponent; converted DesiredStateDetails and DiscoveredResources to the standard hand-rolled drawer pattern; relocated countActiveFilters into FilterDrawer/
- Added a drawer-friendly TimestampRangeFilter (two TimestampPickers + per-field Apply) and adopted it in Events/Parameters/DesiredState, replacing the inline date-range duplication
- Removed the now-dead resources.filters.active.group.close word

**Own words:**
The AI description sums up a lot already.
In general there were a lot of patterns for the filter drawers who could be deduplicated/extracted, which I did and then I adjusted it everywhere necessary in the same shape/style.
So everything should honestly be really consistent with the exact same code/style implementations.
Also removed unused stuff or renamed files for naming consistency as well.
All in all solid refactor I reckon, I adjusted the ticket to mostly match what happened in this PR.
Let me know if you have any concerns/feedback.

Ps: Got no screenshot(s) because everything stayed exactly the same, functionally and visually.

closes https://github.com/inmanta/web-console/issues/7027


LukasStordeur (commented):
We have a lot of "onAddxxxx", "removexxxChip" "clearxxx" it might be worth having a generic hook that takes the filterKey for those. Maybe make better use of the wrapper we currently already use `ConnectedFilterWidget` and let it have a Context with a config we pass to it?
A bit like the form generator, but in a less convoluted way.

AronH99:
So what happened:

- Made a hook (getFilterActions) which returns functions/methods to work with for all of the drawers, currently supported: strings, ints, dates. This can be extended later if needed but this should cover most of the use cases.
I then implemented this hook in all of the drawers and the hook also has it's own unit test file.
- Also extracted a IntRangeFilter which currently is only used in 1 drawer but since we were going for a full refactor I felt this was necessary in combination with the ints supported inside of my hook as well. 
It should be easy to reuse now into the future. 
- Adjusted e2e test.

Let me know if you have any concern, I tested all drawers manually and everything works just as before.

AronH99 (commented):
## 🤖 Claude pre-review - PR #7120 - Move filter building blocks to shared UI components and reuse across drawers

_Automated pre-review by Claude, borrowing the team's review standards. Not a human sign-off._

### Context map

**Type / intent:** `refactor` - dedupe the filter-drawer patterns into shared building blocks and adopt them everywhere (closes #7027). Directly follows up Lukas's earlier note asking for a generic key-scoped hook.

**Surface area**
- **New shared layer** `src/UI/Components/FilterDrawer/`: `getFilterActions` (+unit test), `ActiveFilters` shell, `ActiveFilterGroup` (moved from Resources), `utils` (moved).
- **New/rewritten shared filters** `src/UI/Components/Filters/`: `TimestampRangeFilter` (new), `IntRangeFilter` (rewritten from toolbar->drawer shape).
- **Promoted to top-level shared** `src/UI/Components/AddableTextInput/`, `AddableSelectInput/` (moved out of Resources, all display text now via props).
- **Drawer adopters** (each slimmed to the same shape): Agents, Events, Parameters, DesiredState, DesiredStateDetails, DiscoveredResources, Resources.
- **Removed:** config-driven `src/UI/Components/FilterWidget/` (component + 210-line test), dead words `active.clearAll` / `active.group.close`.

**Size / balance:** +1162 / -1964 across 37 files (net -800, good for a dedupe). Test delta: `filterActions` well covered; `Addable*` tests moved+extended; but the two new/rewritten range filters and the `ActiveFilters` shell have no direct unit tests.

**Blast radius: high** - touches shared `src/UI/Components` used across 7 slices. I verified there are no leftover references to the removed shared `FilterWidgetComponent`, the deleted words keys, or the old toolbar `IntRangeFilter` API (it had no other consumers). CI is green (Jenkins + CodeQL).

**Verdict: Comment** (posted as COMMENT since GitHub blocks Request-changes on your own PR - treat the should-fixes as change requests).

Strong, cohesive refactor that does what the issue asks and lands Lukas's suggestion cleanly - `getFilterActions` is a nice key-scoped, per-shape API and behaviour looks preserved (with intentional empty-list->undefined normalisation, covered by the updated Resources test). No correctness blocker found. What's holding it back from a clean approve: **2 should-fix** (missing unit tests on the two new range filters; a centralised aria word dropped for a hardcoded string), **1 should-fix/design** (chip removal re-parses the display label), **2 questions** (global `!`->red chips; range-filter aria labels not derived from `label`), **1 nit** (a deleted PF-beta caveat comment). Details inline.

src/UI/Components/Filters/TimestampRangeFilter.tsx
```diff
@@ -0,0 +1,111 @@
...
+              onChange={setFrom}
+              from={undefined}
+              datePickerLabel="From Date Picker"
+              timePickerLabel="From Time Picker"
+              action={
+                <Button
+                  variant="control"
+                  onClick={applyFrom}
+                  isDisabled={!from}
+                  aria-label="Apply date from filter"
```
AronH99: 🤖 Claude · [Should-fix]: New shared component with real apply/replace logic (commit a bound -> replace the same-operator entry, ignore empty) but no direct unit test. Same for `IntRangeFilter`. These are now used by 3+ drawers, so a small Vitest around `onChange` (apply from, apply to, replace existing operator, no-op on empty) would lock the contract. Separately, these aria-labels are hardcoded `"Apply date from filter"` / `"Apply date to filter"` regardless of `label`, whereas `IntRangeFilter` derives them from `label` (`Apply ${label} from filter`). Worth matching one style - and note two `TimestampRangeFilter`s on one page would collide on these labels + the fixed `"From Date Picker"` picker labels.

src/UI/Components/Filters/IntRangeFilter.tsx
```diff
@@ -3,133 +3,123 @@ import {
...
+ * @returns {React.ReactElement} The rendered integer-range filter input.
+ */
+export const IntRangeFilter: React.FC<Props> = ({ label, fromLabel, toLabel, value, onChange }) => {
+  const [from, setFrom] = useState("");
+  const [to, setTo] = useState("");
 
-  const onApply = () => {
-    const withNewFrom = insertNewValue(intRangeFilters, from, RangeOperator.Operator.From);
-    const withNewTo = insertNewValue(withNewFrom, to, RangeOperator.Operator.To);
+  const applyBound = (raw: string, operator: RangeOperator.Operator, reset: () => void) => {
```
AronH99: 🤖 Claude · [Should-fix]: The rewrite added non-trivial logic here (parseInt + isNaN guard, replace-same-operator, Enter-to-apply) but there's no unit test for it. Given it's now a shared building block, a test covering apply/replace/invalid-input would be valuable. (The old toolbar version had no other consumers - confirmed via grep - so dropping that API is safe. :) )

src/UI/Components/FilterDrawer/ActiveFilterGroup.tsx
```diff
@@ -39,9 +39,7 @@ export const ActiveFilterGroup: React.FC<ActiveFilterGroupProps> = ({
       isClosable={Boolean(onRemoveGroup)}
       onClick={onRemoveGroup}
       isEditable
-      closeBtnAriaLabel={
-        onRemoveGroup ? words("resources.filters.active.group.close")(title) : undefined
-      }
+      closeBtnAriaLabel={onRemoveGroup ? `Remove ${title} filters` : undefined}
```
AronH99: 🤖 Claude · [Should-fix]: This drops the centralised `words("resources.filters.active.group.close")(title)` for an inline template string (and the word was deleted from `words.tsx`). The repo convention is to keep UI/aria strings in the `words` module. Since this is now the shared group used by every drawer, I'd keep it centralised - re-add a `words` entry (e.g. a generic `filters.active.group.close`) rather than hardcoding here.

src/UI/Components/FilterDrawer/filterActions.ts
```diff
@@ -0,0 +1,128 @@
...
+    /** Display labels for the range, each formatted "operator | date" through the presenter. */
+    dateChips: (key: DateKey<F>, presenter: DatePresenter): string[] => {
+      return arrayValues<DateRange.Type>(key).map(
+        (entry) => `${entry.operator} | ${presenter.getFull(entry.date.toISOString())}`
+      );
+    },
+
+    /** Remove the chip whose label matches, keyed by its leading operator. */
+    removeDateChip: (key: DateKey<F>, label: string) => {
+      const operator = label.split("|")[0].trim();
```
AronH99, line 100: 🤖 Claude · [Should-fix]: `removeDateChip`/`removeIntChip` recover the operator by re-parsing the display label (`label.split("|")[0]`) that `dateChips`/`intChips` produced - so removal is coupled to the display format. It works today (operator enum is the stable prefix) and matches the old code, but now it's enshrined in a shared util. Cleaner would be for the chip helpers to return `{ operator, label }` and have remove take the operator directly, decoupling logic from presentation. Not blocking - flagging since this is the shared surface everything now depends on.

src/UI/Components/FilterDrawer/ActiveFilterGroup.tsx
```diff
@@ -13,6 +12,7 @@ export interface ActiveFilterGroupProps {
  * The ActiveFilterGroup component.
  *
  * This component is responsible of rendering the active values of a single filter category as a dismissible label group.
+ * Values prefixed with "!" (exclusions) are rendered in red. Shared building block for the filter drawers.
```
AronH99, line 16: 🤖 Claude · [Question]: The `!`-prefix -> red styling was Resources-specific (include/exclude), and this component is now shared by every drawer. In drawers with no exclude concept (Parameters name/source, Events version, Agents name), a user typing a literal value starting with `!` would now get a red chip. Harmless in practice, but it's a product-behaviour call - worth a human confirm before merge, or scope the red styling to drawers that actually support exclusions.

src/UI/Components/Filters/TimestampPicker.tsx
```diff
@@ -19,7 +19,6 @@ interface Props {
   action?: React.ReactNode;
 }
 
-/** Both the Date and the Time Picker from Patternfly are in beta stage, e.g. validation doesn't work as expected */
 export const TimestampPicker: React.FC<Props> = ({
```
AronH99, line 22: 🤖 Claude · [Nit]: This drops the `/** Both the Date and the Time Picker from Patternfly are in beta stage, e.g. validation doesn't work as expected */` caveat. Convention here is to keep deliberate-tradeoff/gotcha comments - this one warns future readers about PF beta validation, unrelated to the refactor. I'd restore it.

src/UI/Components/Filters/TimestampRangeFilter.tsx
AronH99: 🤖 Claude: Addressed in c1bec94b - added `TimestampRangeFilter.test.tsx` covering apply-from/to, replace-same-operator, keep-other-operator, and buttons-disabled-until-picked. Left the aria labels as-is for now (one range per page, so no collision).

src/UI/Components/Filters/IntRangeFilter.tsx
AronH99: 🤖 Claude: Addressed in c1bec94b - added `IntRangeFilter.test.tsx` covering apply-from/to, apply-on-Enter, replace-same-operator, and keep-other-operator.

src/UI/Components/FilterDrawer/ActiveFilterGroup.tsx
AronH99: 🤖 Claude: Addressed in c1bec94b - re-added `resources.filters.active.group.close` to words.tsx and `ActiveFilterGroup` now uses `words(...)(title)` instead of the hardcoded string.

src/UI/Components/FilterDrawer/filterActions.ts
AronH99: 🤖 Claude: Reconsidered and dropping this one - the operator is always the stable leading token of the label (`from | ...` / `to | ...`), so `split("|")[0]` is safe by construction. The index-based alternative I first applied leaked an unused index into the string-chip `onRemove` handlers (type/agent/value/status), which wasn't worth it. Keeping the original operator-parse.

src/UI/Components/FilterDrawer/ActiveFilterGroup.tsx
AronH99: This is fine for now, it understands the !prefix but if this gives issues we will address it later

src/Slices/DesiredStateDetails/UI/Controls/FilterWidget/DesiredStateDetailsFilterWidget.tsx
```diff
@@ -1,38 +1,48 @@
...
 import { words } from "@/UI/words";
 
 interface DesiredStateDetailsFilterWidgetProps {
   onClose: () => void;
 }
 
-const FIELDS: FilterField[] = [
+type TextFilterKey = "type" | "agent" | "value";
+
+const FIELDS: { key: TextFilterKey; label: string; placeholder: string }[] = [
```
LukasStordeur, line 20: Probably best to have it as a next phase. But I'd like to have all the widgets work with a config object like this one. And maybe add a fieldType to it, to make it complete.

src/UI/Components/FilterDrawer/filterActions.ts
```diff
@@ -0,0 +1,128 @@
...
+
+/** Keys of F whose value is a DateRange array - i.e. the from/to date-range fields. */
+type DateKey<F> = {
+  [K in keyof F]-?: NonNullable<F[K]> extends readonly DateRange.Type[] ? K : never;
+}[keyof F];
+
+/** Keys of F whose value is an IntRange array - i.e. the from/to integer-range fields. */
+type IntKey<F> = {
+  [K in keyof F]-?: NonNullable<F[K]> extends readonly IntRange.Type[] ? K : never;
+}[keyof F];
```
LukasStordeur: This is really obscure Typescript for me, is there no way to write it in a more readable way?

src/UI/Components/FilterDrawer/filterActions.ts
```diff
@@ -0,0 +1,128 @@
...
+ * fields (e.g. a page's disregardDefault flag). For the ranges, applying a bound is handled
+ * upstream by TimestampRangeFilter / IntRangeFilter, so only the chip side lives here.
+ *
+ * @param filter - The current filter state.
+ * @param setFilter - Setter to persist filter changes upstream.
+ * @returns The scoped, per-shape actions described above.
+ */
+export const getFilterActions = <F extends object>(filter: F, setFilter: (filter: F) => void) => {
+  /** Read an array field, coalescing a missing/unset value to []. Caller supplies the element type E. */
+  const arrayValues = <E>(key: keyof F): E[] => {
```
LukasStordeur, line 46: Same here, hard to read typing, but maybe it's just me...?

src/UI/Components/FilterDrawer/filterActions.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    return Array.isArray(value) ? value : [];
+  };
+
+  /** Write a string list, dropping an emptied list to undefined. `patch` merges in extra fields. */
+  const writeStrings = (key: ListKey<F>, values: string[], patch?: Partial<F>) => {
+    setFilter({ ...filter, [key]: values.length > 0 ? values : undefined, ...patch });
+  };
+
+  return {
```
LukasStordeur, line 57: I'd personnaly would have worked a bit different, and added a FieldType that would result in a switch case like we have for the form :) but can be done in a next pass.

src/UI/Components/Filters/TimestampRangeFilter.tsx
```diff
@@ -0,0 +1,111 @@
...
+  label,
+  fromLabel,
+  toLabel,
+  value,
+  onChange,
+}) => {
+  const [from, setFrom] = useState<Date | undefined>();
+  const [to, setTo] = useState<Date | undefined>();
+
+  const applyBound = (date: Date, operator: RangeOperator.Operator) => {
```
LukasStordeur: what's Bound? For me a Bound is a Link, or relation 🤔 (Comment for both the TIme related range filters)

src/UI/Components/Filters/TimestampRangeFilter.tsx
```diff
@@ -0,0 +1,111 @@
...
+            />
+          </FormGroup>
+        </FlexItem>
+        <FlexItem>
+          <FormGroup label={toLabel}>
+            <TimestampPicker
+              timestamp={to}
+              onChange={setTo}
+              from={from}
+              datePickerLabel="To Date Picker"
```
LukasStordeur: if those are actual labels displayed to the user, they need to be in the words-dict

LukasStordeur (changes requested):

src/Slices/DesiredStateDetails/UI/Controls/FilterWidget/DesiredStateDetailsFilterWidget.tsx
AronH99: Was exactly thinking the same to be fair, but then I needed a config file/object and this would be a little too much for this PR, I'll make a ticket for this one thoough

src/UI/Components/FilterDrawer/filterActions.ts
AronH99: No for sure you are definitely right, putting this in the follow up ticket too

src/Slices/DesiredStateDetails/UI/Controls/FilterWidget/DesiredStateDetailsFilterWidget.tsx
AronH99: https://github.com/inmanta/web-console/issues/7142

src/UI/Components/FilterDrawer/filterActions.ts
AronH99: https://github.com/inmanta/web-console/issues/7142

AronH99:
Those more complex typescript types are mostly beneficial for type guarding.
The filterActions file will most likely be adjusted/removed later on in:
https://github.com/inmanta/web-console/issues/7142
for now I placed some more docstrings and made a few small adjustments etc to make it a little more clear because you are definitely right that it's a bit obscure.

All your other feedback has been addressed as well

src/UI/Components/FilterDrawer/filterActions.ts
```diff
@@ -0,0 +1,135 @@
...
+import type { DateRange, IntRange } from "@/Core";
+import { uniq } from "@/Core/Language/collection";
+import type { DatePresenter } from "@/UI/Presenters";
+
+/**
+ * Given a filter type F and a value type T, resolves to the union of F's keys whose value is a
+ * T (ignoring whether the field is optional). This is what lets each action below accept only
+ * the keys it makes sense for - e.g. a date action won't accept the key of a string field.
+ */
+type KeyForValue<F, T> = {
```
LukasStordeur, line 11: Let's add this to the ticket too about the config object.

src/UI/Components/FilterDrawer/filterActions.ts
LukasStordeur: Add to ticket, you resolved it, but I have no idea what your opinion is on it.

src/UI/Components/Filters/IntRangeFilter.tsx
```diff
@@ -3,133 +3,124 @@ import {
...
+      ...value.filter((entry) => entry.operator !== operator),
+      { value: parsed, operator },
+    ]);
+    reset();
   };
 
-  const removeChip = (category, value) => {
-    const raw = prettyToRaw(value);
+  const applyFrom = () => applyRangeValue(from, RangeOperator.Operator.From, () => setFrom(""));
+  const applyTo = () => applyRangeValue(to, RangeOperator.Operator.To, () => setTo(""));
```
LukasStordeur: Block theselike you did on the other component, instead of chaining double anonymous methods.

LukasStordeur (approved):
I'll approve, but resolve the last comment/answer them :)

PR #7141, opened 2026-07-17
Restructure Tokenform

LukasStordeur:
# Description

Restructured the form to keep a more logical order and present better.

image
image

### DIFF
--- a/src/Slices/Settings/UI/Tabs/Token/ExpiryInput.tsx
+++ b/src/Slices/Settings/UI/Tabs/Token/ExpiryInput.tsx
@@ -65,7 +65,7 @@ export const ExpiryInput: React.FC<Props> = ({ onChange, isDisabled }) => {
   };
 
   return (
-    <FormGroup label={words("settings.tabs.token.expiry")} fieldId="token-expiry">
+    <StyledFormGroup label={words("settings.tabs.token.expiry")} fieldId="token-expiry">
       <Flex gap={{ default: "gapSm" }} flexWrap={{ default: "nowrap" }}>
         <FlexItem>
           <FormSelect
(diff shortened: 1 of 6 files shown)


src/Slices/Settings/UI/Tabs/Token/ExpiryInput.tsx
```diff
@@ -122,3 +122,7 @@ const AmountInput = styled(TextInput)`
 const UnitSelect = styled(FormSelect)`
   width: auto;
 `;
+
+const StyledFormGroup = styled(FormGroup)`
+  --pf-v6-c-form--m-horizontal__group-label--md--GridColumnWidth: 16rem;
```
AronH99, line 127: We might be able to dedupe this into a tokenFormStyles.ts or something?
I know it's small but at least we have the same for both the ExpiryInput as the TokenForm, up to you though, don't think there are ways to improve this otherwise

AronH99 (approved):

src/Slices/Settings/UI/Tabs/Token/ExpiryInput.tsx
LukasStordeur: It's an actual component, and if I move this one, I should then move the others too? I know one's reused twice, but not the others really, so I'm conflicted.

AronH99: I think it might be beneficial to review all of the --pf-v6 tokens (not that much I saw) and then replace them by stable tokens (non v6 ones) or remove them if possible?
But this might be for a separate ticket to be honest..
Wouldn't bother moving them to separate files at this point no

LukasStordeur: agreed for a separate task.

Issue #7134, opened 2026-07-16
Use the UnitInputField in the token form


LukasStordeur:
Use the UnitInputField in the token tab

Issue #7133, opened 2026-07-16
Use the UnitInputField in the Instance forms


LukasStordeur:
Use the UnitInputField in the service instance form.

Issue #7132, opened 2026-07-16
Display converted values in tables with corresponding unit


LukasStordeur:
Display converted values in tables with corresponding unit

PR #7155, opened 2026-07-23
1) UnitFormInput - Reusable component

LukasStordeur:
# Description

Part one of the unit input field task. 

- `units.ts` — unit catalogue (Size/Bit-rate/Byte-rate/Duration), factor lookup, web_unit_scales/web_unit_display defaulting rules
- `resolveUnitConfig.ts` — resolves the web_unit* annotations into offered units via the "integer rule", with graceful degradation (never throws) for bad/unsupported input
- `convert.ts` — BigInt-safe unit conversion via bignumber.js, downgrading to number/bigint per the repo's existing big-int convention
- `validate.ts` — exactness (whole-number-of-API-unit) and bounds (ge/gt/le/lt) validation, returning structured errors
- `display.ts` — auto unit-selection for initial/read-only display, plus the "other scale" equivalent used by the live helper line
- `UnitFormInput.tsx` — the actual PatternFly component: numeric input + steppers + unit dropdown + helper text, self-validating, unit-switch re-interprets typed digits rather than converting them

Coverage: 141 new tests (unit, integration, component + accessibility), all passing; repo-wide lint/typecheck/test suite (1869 tests) green with zero regressions.

Scope: standalone, reusable component only — not yet wired into forms (#7133), read-only tables (#7132), or settings (#7134).

image

> ℹ️  NOTE: The increase/decrease buttons have been removed in the next pr, to match the requested design better and support native scrolling behavior. 
> The width of the dropdowns have also been adjusted to a fixed width in the next pr.


AronH99 (approved):

src/UI/Components/UnitInput/units.ts
```diff
@@ -0,0 +1,172 @@
...
+    return "both";
+  }
+
+  const family = familyOf(kind, webUnit);
+
+  return family === "metric" || family === "iec" ? family : null;
+}
+
+/** `web_unit_display` defaulting rule: defaults to `web_unit` itself. */
+export function defaultDisplay(webUnit: string): string {
```
AronH99: maybe 1 small thing:
we don't really need this function do we? or was this meant to be more than what it is now?
we can clean this up and the unit tests for it?

AronH99 (approved):

src/UI/Components/UnitInput/units.ts
LukasStordeur: you're right, no idea why it was still there 😅

PR #7157, opened 2026-07-23
2) UnitFormInput - Wire into Create/Edit/Duplicate instance forms + Composer

LukasStordeur:
# Description

Attributes annotated with web_presentation "unit" now render the UnitInputField in Create/Edit/Duplicate instance forms and the Composer, instead of a plain number field.

I adjusted the visual  little, to use the native input arrows, which also adds the support to scroll up and down to adjust the value. 

image

image
> Note: the color of the headers of the shapes is unrelated to this task, and will be fixed in a different pr. 

closes:
- #7133


src/Test/Data/Field.ts
```diff
@@ -163,6 +164,26 @@ export const numberArr: TextField = {
...
+  description: "description",
+  isOptional: true,
+  isDisabled: false,
+  defaultValue: null,
+  type: "int?",
+  config: {
+    kind: "bitrate",
+    apiUnit: "kbit/s",
+    isInt: true,
+    isOptional: true,
```
AronH99: The isOptional bool under the config of a unit is not in use at this time anywhere, is this possible or am I missing something?

AronH99 (commented):
1 thing to check out, but I am going to approve already..
(Going to test on the last pr locally as well)

AronH99 (approved):

src/Test/Data/Field.ts
LukasStordeur: It's used on the higher level, I missed the redundancy.  I'll see to remove it :)

PR #7158, opened 2026-07-23
3) UnitFormInput - Unit attribute field tables

LukasStordeur:
# Description

Unit fields are now also displayed properly in the attributes table on the instance details page. 
There's a tooltip to display the raw value.

image

closes:
- #7132


AronH99 (approved):
Oke, ook wat getest en zag er allemaal goed uit :)

Issue #7136, opened 2026-07-16
Orchestrator status card


LukasStordeur:
Create a temporary route for the new dashboard page.
Create the top Orchestrator status card.

PR #7165, opened 2026-07-29
1) Dashboard v2 - General status/Summary

LukasStordeur:
# Description

The first stage lays the foundation for the page.
The old dashboard page will be removed in the last step, and the name V2 removed then. 

We don't yet have a way to get a summary for the agents, like we do for the resources. So I split it up in three calls with the minimum paging set to 1, to get the totalAmount per type in the metadata. It's a hack, but that will fetch the minimal amount of data to get the information we need. 

closes:
- #7136

> Note:  this will be a feature branch, I will add the changelog for the entire feature in the last step.

https://github.com/user-attachments/assets/3cf7b736-b7eb-40de-8a22-7582ae4ec27d


AronH99 (commented):
So the code is pretty wel structured however there is 1 thing I would improve:
The cards on first load will shift in height, I actually have some cool thing to show for how we can solve these kinds of issues in graphql in the future, which I will do in the demo's but for now I think it might be beneficial to work with Patternfly's skeleton loading guards until everything is loaded? (see video)
It might be needed to give the cards a max height or height for this to work consistently, have loading guards and have no shifting content though..

https://github.com/user-attachments/assets/e2f98a74-fee9-49d9-a045-2f4482648f59

LukasStordeur:
> So the code is pretty wel structured however there is 1 thing I would improve: The cards on first load will shift in height, I actually have some cool thing to show for how we can solve these kinds of issues in graphql in the future, which I will do in the demo's but for now I think it might be beneficial to work with Patternfly's skeleton loading guards until everything is loaded? (see video) It might be needed to give the cards a max height or height for this to work consistently, have loading guards and have no shifting content though..
> 


Yes, I think that's a good improvement, but if you don't mind, I'd prefer to make a ticket for it rather than take it up in this pr?

AronH99 (approved):

Issue #7139, opened 2026-07-16
Create the resource card


LukasStordeur:
<img width="504" height="504" alt="Image" src="https://github.com/user-attachments/assets/8a311d4f-4eef-44fb-a7a2-add25989917f" />

PR #7170, opened 2026-07-31
4) Dashboard v2 - Resources Card

LukasStordeur:
# Description

Stage 4 of the dashboard v2. 
The resources card was added.
Clicking on a section on the progressbars will navigate to the resources page with the filter preset. The !orphan
is by default kept when redirecting. 

https://github.com/user-attachments/assets/b8591e75-9837-45e9-b53e-0524b91f059b

closes: 
- #7139


src/Slices/DashboardV2/UI/resourcesHealth.ts
```diff
@@ -9,13 +9,19 @@ export interface ResourcesHealth {
 
 /**
  * "Failed to deploy" is mapped to resourceSummary.lastHandlerRun.failed — a handler execution
- * failure is the most direct read of that phrase.
+ * failure is the most direct read of that phrase. Severity stays keyed off that count alone;
+ * the total resource count and non-compliant count are shown as extra context, not additional
+ * severity triggers.
```
LukasStordeur, line 14: Need to confirm this, if non-compliant is considered as an Attention warning, or can remain as Healthy.

src/Slices/DashboardV2/UI/Components/Resources/ResourcesCard.tsx
```diff
@@ -0,0 +1,225 @@
...
+import { DependencyContext } from "@/UI/Dependency";
+import { SearchHelper } from "@/UI/Routing/SearchHelper";
+import { words } from "@/UI/words";
+import { MINIMAL_PAGE } from "../../EnvironmentHealthRow";
+import { IconBadge } from "../IconBadge";
+import { ResourceStatusBar } from "./ResourceStatusBar";
+
+const searchHelper = new SearchHelper();
+
+const EMPTY_COMPLIANCE: Record<Resource.ComplianceKey, number> = {
```
AronH99: Except for the ts issue which can probably be solved in another way i don't think we really need these default fallbacks, I tested it with empty environment and as i though the bars will still show a grayed out bar with 0 in it..
So I think I would delete these?

src/Slices/DashboardV2/UI/Components/Resources/ResourceStatusBar.tsx
```diff
@@ -0,0 +1,97 @@
...
+              </FlexItem>
+            </Flex>
+          ))}
+        </Flex>
+      )}
+    </Flex>
+  );
+};
+
+const Dot = styled.span`
```
AronH99: think this is a duplicate of an existing Dot component you already made for the other cards

src/Slices/DashboardV2/UI/resourcesHealth.ts
AronH99: Already confirmed or no?

LukasStordeur: Will need Bart for that, so next week. I think we can leave it as is for now, and tweak it later

AronH99 (approved):

PR #7182, opened 2026-08-05
dashboard v2 - e2e tests

LukasStordeur:
# Description

Add e2e test coverage for the new dashboard.
I ran all test suits locally to ensure they wouldn't break in the nightly builds.

closes:
- #7178

### DIFF
--- a/cypress/e2e/scenario-1.1-dashboard.cy.js
+++ b/cypress/e2e/scenario-1.1-dashboard.cy.js
@@ -1,0 +1,214 @@
+import environmentHelpers from "../support/environmentHelpers";
+
+const { selectEnvironment, getDefaultEnvName } = environmentHelpers;
+
+const isIso = Cypress.expose("edition") === "iso";
+
+describe("10 Dashboard", () => {
+  beforeEach(() => {
+    cy.visit("/console/");
+    selectEnvironment();
+  });
+
+  it("10.1 Selecting an environment lands on the Dashboard", () => {
+    cy.url().should("contain", "/console/dashboard");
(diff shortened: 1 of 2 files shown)


cypress/e2e/scenario-1.1-dashboard.cy.js
```diff
@@ -0,0 +1,214 @@
...
+    // iso's LSM catalog sync compiles automatically on environment setup, guaranteeing rows here;
+    // oss has no such automatic compile, so only assert the table itself renders without erroring.
+    cy.get("tbody tr", { timeout: 30000 }).should(
+      isIso ? "have.length.greaterThan" : "have.length.at.least",
+      0
+    );
+  });
+
+  it("10.4 Orchestrator detail card", () => {
+    cy.contains("h3", "Orchestrator").closest(".pf-v6-c-card").as("orchestratorCard");
```
AronH99, line 70: Can we try and remove all of the pf-v6 selectors?
=> future proof

LukasStordeur: There are many, and sometimes easier to target elements that are within the patternfly components.
For your future proof concern, it won't be a problem. Each upgrade from patternfly comes with a script that will replace old v6- references to what the new standard will be. So I think we can safely use those in the e2e tests?

AronH99: Ahn yes, I know you did mention that, I'd honestly prefer not to use it if possible but honestly if it won't be a big deal later on when migrating, it's fine by me..

AronH99 (approved):

PR #7189, opened 2026-08-07
Moved the Facts page filters (name, resource id) from the inline tool into a side-panel drawer

### LINKED ISSUE #6780: Migrate Facts page filters to side panel drawer
## Context

The Resources page filters have been migrated to a side panel drawer (using PatternFly `Drawer` + `DrawerPanelContent`). Other pages still use the old inline toolbar filter pattern and should be updated to match.

## Current state

`src/Slices/Facts/UI/TableControls.tsx` renders two inline `FreeTextFilter` components (Name and Resource Id) directly inside a PatternFly `Toolbar`. There is no dedicated `FilterWidget` component — the filter logic is embedded in the `TableControls`.

## Goal

Migrate the Facts page filters to the new side panel pattern:

- Extract the two `FreeTextFilter` components into a dedicated `FilterWidget` (or similar) component inside a `DrawerPanelContent`
- Add a `Drawer` + `DrawerContent` + `DrawerPanelContent` wrapping the table
- Add a filter toggle button to the toolbar (with active filter count badge)
- Track `isDrawerExpanded` and `activeFilterCount` state on the page component

## Reference implementation

See `src/Slices/Resource/UI/ResourcesPage/Page.tsx` and `src/Slices/Resource/UI/ResourcesPage/Components/ResourceTableControls.tsx` for the pattern to follow.

AronH99:
# Description

- Migrating the facts page filters to the side-pandel drawer

closes https://github.com/inmanta/web-console/issues/6780

https://github.com/user-attachments/assets/5b4c641f-6a6a-4ff8-9df5-5c4cc6aad65f

### DIFF
--- a/changelogs/unreleased/6780-facts-filter-drawer.yml
+++ b/changelogs/unreleased/6780-facts-filter-drawer.yml
@@ -1,0 +1,6 @@
+description: The Facts page filters (name, resource id) have been moved from the inline toolbar into a side-panel drawer, opened from a filter toggle button that shows the number of active filters, matching the Resources page.
+issue-nr: 6780
+change-type: patch
+destination-branches: [master]
+sections:
+  minor-improvement: "{{description}}"
--- a/src/Slices/Facts/UI/Page.tsx
+++ b/src/Slices/Facts/UI/Page.tsx
@@ -1,4 +1,5 @@
-import React from "react";
+import React, { useCallback, useMemo, useState } from "react";
+import { Drawer, DrawerContent, DrawerContentBody, Stack, StackItem } from "@patternfly/react-core";
 import { usePaginatedTable } from "@/Data";
 import { useGetFacts } from "@/Data/Queries";
 import { Filter, SortKey } from "@/Slices/Facts/Core/Types";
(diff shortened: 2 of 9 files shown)


changelogs/unreleased/6780-facts-filter-drawer.yml
```diff
@@ -0,0 +1,6 @@
+description: The Facts page filters (name, resource id) have been moved from the inline toolbar into a side-panel drawer, opened from a filter toggle button that shows the number of active filters, matching the Resources page.
+issue-nr: 6780
+change-type: patch
```
LukasStordeur: I checked, and we are indeed using minor for all previous filter panel changes.

src/Slices/Facts/UI/Page.tsx
```diff
@@ -52,18 +68,42 @@ export const Page: React.FC = () => {
...
-        ) : (
-          <FactsTable
-            aria-label="Facts-Success"
-            rows={tablePresenter.createRows(data.data)}
-            tablePresenter={tablePresenter}
-            sort={sort}
-            setSort={setSort}
-          />
-        )}
+        <Drawer
```
LukasStordeur, line 75: Use FilterDrawerPanelContent Component instead to stay consistent

AronH99: So what has happened in all refactors is the following:
1. All of the ConnectedFilterWidget (except for the resources one I see now) are using the FilterDrawerPanelContent as you suggested but the pages themselves are using this kind of format everywhere with no exceptions.
This is mainly because if you swap the DrawerContent for the FilterDrawerPanelContent the table has no place to live since DrawerContent has a slot specified for the ConnectedFilterWidget(s) (drawer) and the children are the tables then

=> What I suggest is that I make a followup ticket to extract all of this boilerplate for the pages seperately so we don't have 8 times the same logic for the pages themselves.. 
We can also add the small refactor for the resources widget as well to that ticket so it's at least consistent with the rest, what do you think?

